Issue #970, opened 2017-06-13
git submodule problem


zreigz:
There is some problem with repositories from src/
After 

```
git submodule sync --recursive
git submodule update --init  --recursive

```
some repositories are empty. For instance src/diego-release and I have the following error:

```
Please make sure you have the correct access rights
and the repository exists.
fatal: clone of 'git@github.com:suse/testbrain.git' into submodule path '/home/vagrant/scf/src/hcf-release/src/github.com/suse/testbrain' failed
Failed to clone 'src/hcf-release/src/github.com/suse/testbrain' a second time, aborting

```

viovanov:
@zreigz we need to make that submodule reference an http remote
can you make the change locally and verify it fixes your problem?
remote should be https://github.com/SUSE/testbrain.git

zreigz:
I've already fixed it in PR https://github.com/SUSE/scf/pull/971 and it seems to work

zreigz:
I can confirm it works now

viovanov:
thanks @zreigz

PR #971, opened 2017-06-13
Fix git submodule repositories

zreigz:
This fixes problem with git modules


zreigz:
Fix for https://github.com/SUSE/scf/issues/970

PR #974, opened 2017-06-13
Added make targets for running smoke and cats.

andreas-kupries:


make/cats
```diff
@@ -0,0 +1,47 @@
...
+    :
+else
+    echo 1>&2 Namespace ${NAMESPACE} is missing. SCF is not running.
+    exit 1
+fi
+
+stampy "${GIT_ROOT}/scf_metrics.csv" "$0" make-cats::setup start
+
+cf api api.cf-dev.io --skip-ssl-validation
+cf auth admin changeme
```
alexjh: The admin password is stored in `bin/settings/settings.env`, I think we could source it and use it from there.

make/cats
```diff
@@ -0,0 +1,47 @@
...
+cf enable-feature-flag diego_docker
+
+stampy "${GIT_ROOT}/scf_metrics.csv" "$0" make-cats::setup end
+
+stampy "${GIT_ROOT}/scf_metrics.csv" "$0" make-cats::create start
+
+# Delete left-over pod/definition from previous runs, then create/run
+kubectl delete --namespace="${NAMESPACE}" --filename="kube/bosh-task/acceptance-tests.yml"
```
alexjh: The script is failing on the delete for me, because I don't have any prior runs. Add a `2> /dev/null || /bin/true` ?

make/cats
andreas-kupries: Checking ...

make/cats
andreas-kupries: Yes, better do

make/cats
andreas-kupries: Done with commit 2599ecd1bdbd536c118d1181d8ed171909817ea6

make/cats
andreas-kupries: Done with commit 2599ecd1bdbd536c118d1181d8ed171909817ea6

jimmykarily:
Should we update the [README](https://github.com/suse/scf#how-do-i-run-smoke-and-acceptance-tests) ?

andreas-kupries:
Yes, IMHO.

andreas-kupries:
I will make a new PR for the README change.

PR #977, opened 2017-06-15
Centralized versions

andreas-kupries:
Trello: https://trello.com/c/U85NRLe4/149-2-collect-versions-all-in-versionssh
Rebased.


.gitignore
```diff
@@ -21,9 +21,10 @@ src/*-clone
 src/buildpacks/*-clone
 /*.tf
 /*.tf.json
-
+bin/settings/kube/ca.env
 # `make dist` output of various forms
 output/
 /kube/bosh/
 /kube/bosh-task/
+/kube/secrets/
```
jandubois: Should we just ignore `kube/` instead of the 3 lines above? I don't see that we want to check anything under that directory into git.

bin/dev/versions.sh
```diff
@@ -0,0 +1,55 @@
+#!/bin/bash
+set -o errexit -o nounset
+
+# Versions of the pieces pulled into SCF (outside of submodules).
+# This file is sourced everywhere one or more of the versions are
+# needed.
+
+# I. bin/dev/install_tool.sh
```
jandubois: I generally dislike numbered comment when the sequence isn't actually that important. It becomes a nuisance to renumber things when you insert something in the middle.

Also, roman numbers?

.gitignore
andreas-kupries: Definitely simpler. :+1:

bin/dev/versions.sh
andreas-kupries: Ok. Will change (drop numbers)

.gitignore
andreas-kupries: Done with commit 655c2f883456a4f62520d122886053c50bb5f324

bin/dev/versions.sh
andreas-kupries: Done with commit 655c2f883456a4f62520d122886053c50bb5f324

make/include/versioning
```diff
@@ -25,7 +25,8 @@ GIT_TAG=${GIT_TAG:-$(echo ${GIT_DESCRIBE} | awk -F - '{ print $1 }' )}
 GIT_COMMITS=${GIT_COMMITS:-$(echo ${GIT_DESCRIBE} | awk -F - '{ print $2 }' )}
 GIT_SHA=${GIT_SHA:-$(echo ${GIT_DESCRIBE} | awk -F - '{ print $3 }' )}
 
-CF_VERSION=$(cat "${GIT_ROOT}/CF_VERSION")
+. $(dirname $(dirname "$0"))/bin/dev/versions.sh
```
jandubois: Is `$0` really correct, given that this script is being sourced and not invoked? Shouldn't it be `$BASH_SOURCE`?

andreas-kupries: During my tests it found the correct file. So, looks to be ok.
I was actually uncertain and fully prepared to hunt down the syntax for the `$BASH_SOURCE` thing.
When it worked I shelved that.

jandubois: So I suggest to switch to

```
. $(dirname $(dirname $(dirname "$BASH_SOURCE")))/bin/dev/versions.sh
```

to become independent of the location of the caller and only rely on the location of this script itself.

andreas-kupries: Done with commit 4d873d3f04509e13ef6e64b21c8c915c9cfbd648

PR #978, opened 2017-06-15
Increased verbosity for p-d-s

andreas-kupries:
https://trello.com/c/8sPDuSg3/174-2-post-deployment-api-routing-api-setup-needs-to-output-something
Role `post-deployment`: `uaa-create-user` - Increase verbosity
- curl more verbose
- retry shows command invoked, explicitly notes success/failure, waiting times
- Extended status messages (added url), fixed typos, normalized to a single form.
Ditto same role job `cf-set-proxy`
Ditto role `api, script `fetch_uaa_...`

### DIFF
--- a/container-host-files/etc/hcf/config/scripts/fetch_uaa_verification_key.sh
+++ b/container-host-files/etc/hcf/config/scripts/fetch_uaa_verification_key.sh
@@ -29,13 +29,13 @@ function retry_forever () {
     shift 1
 
     while true ; do
-	printf "Trying: %s\n" "$*"
+        printf "Trying: %s\n" "$*"
         if "$@" ; then
-	    status ' SUCCESS'
+            status ' SUCCESS'
             break
         fi
-	trouble '  FAILED'
-	status "Waiting ${delay} ..."
+        trouble '  FAILED'
+        status "Waiting ${delay} ..."
         sleep "${delay}"
     done
 }
--- a/container-host-files/etc/hcf/config/scripts/fetch_uaa_verification_key.sh
+++ b/container-host-files/etc/hcf/config/scripts/fetch_uaa_verification_key.sh
@@ -7,19 +7,50 @@
 
 # Note that this is *sourced* into run.sh, so we can't exit the shell.
 
-# Wait for UAA
-while true ; do
-    if curl --fail \
-        $(if test "${SKIP_CERT_VERIFY_EXTERNAL}" = "true" ; then echo "--insecure" ; fi) \
-        "${HCF_UAA_INTERNAL_URL}/token_key"
-    then
-        break
-    fi
-    sleep 10
-done
+
+# Report progress to the user; use as printf
+status() {
+    local fmt="${1}"
+    shift
+    printf "\n%b${fmt}%b\n" "\033[0;32m" "$@" "\033[0m"
+}
+
+# Report problem to the user; use as printf
+trouble() {
+    local fmt="${1}"
+    shift
+    printf "\n%b${fmt}%b\n" "\033[0;31m" "$@" "\033[0m"
+}
+
+# helper function to retry a command several times, with a delay between trials
+# usage: retry <max-tries> <delay> <command>...
+function retry_forever () {
+    delay=${1}
+    shift 1
+
+    while true ; do
+	printf "Trying: %s\n" "$*"
+        if "$@" ; then
+	    status ' SUCCESS'
+            break
+        fi
+	trouble '  FAILED'
+	status "Waiting ${delay} ..."
+        sleep "${delay}"
+    done
+}
+
+SKIP=$(if test "${SKIP_CERT_VERIFY_EXTERNAL}" = "true" ; then echo "--insecure" ; fi)
+
+status "Waiting for UAA to be available at ${HCF_UAA_INTERNAL_URL}/token_key ..."
+retry_forever 10s curl --connect-timeout 5 --fail $SKIP "${HCF_UAA_INTERNAL_URL}/token_key"
+
+status "Extract JWT public signing key"
 export JWT_SIGNING_PUB="$(\
     { curl --fail $(if test "${SKIP_CERT_VERIFY_EXTERNAL}" = "true" ; then echo "--insecure" ; fi)\
         "${HCF_UAA_INTERNAL_URL}/token_key" \
         || exit 1 \
     ; } \
     | awk 'BEGIN { RS="," ; FS="\"" } /value/ { if ($2 == "value") print $4 } ')"
+
+status DONE
--- a/container-host-files/etc/hcf/config/scripts/fetch_uaa_verification_key.sh
+++ b/container-host-files/etc/hcf/config/scripts/fetch_uaa_verification_key.sh
@@ -42,7 +42,7 @@ function retry_forever () {
 
 SKIP=$(if test "${SKIP_CERT_VERIFY_EXTERNAL}" = "true" ; then echo "--insecure" ; fi)
 
-status "Waiting for UAA to be available at ${HCF_UAA_INTERNAL_URL}/token_key ..."
+status "Waiting for UAA to be available ..."
 retry_forever 10s curl --connect-timeout 5 --fail $SKIP "${HCF_UAA_INTERNAL_URL}/token_key"
 
 status "Extract JWT public signing key"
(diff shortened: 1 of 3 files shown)


container-host-files/etc/hcf/config/scripts/fetch_uaa_verification_key.sh
```diff
@@ -7,19 +7,50 @@
...
+# Report problem to the user; use as printf
+trouble() {
+    local fmt="${1}"
+    shift
+    printf "\n%b${fmt}%b\n" "\033[0;31m" "$@" "\033[0m"
+}
+
+# helper function to retry a command several times, with a delay between trials
+# usage: retry <max-tries> <delay> <command>...
+function retry_forever () {
```
mook-as: The comment seems wrong; there's no `max-tries` (this is, after all, forever).

container-host-files/etc/hcf/config/scripts/fetch_uaa_verification_key.sh
```diff
@@ -7,19 +7,50 @@
...
+    delay=${1}
+    shift 1
+
+    while true ; do
+	printf "Trying: %s\n" "$*"
+        if "$@" ; then
+	    status ' SUCCESS'
+            break
```
mook-as: Mix of tab & space all over in this function. (It makes the indenting on GitHub weird.)

container-host-files/etc/hcf/config/scripts/fetch_uaa_verification_key.sh
```diff
@@ -7,19 +7,50 @@
...
+        fi
+	trouble '  FAILED'
+	status "Waiting ${delay} ..."
```
mook-as: You do have format strings — `status "Waiting %s..." "${delay}"` works too. (But it doesn't look any clearer… why _do_ we have format strings?)

container-host-files/etc/hcf/config/scripts/fetch_uaa_verification_key.sh
```diff
@@ -7,19 +7,50 @@
...
+        sleep "${delay}"
+    done
+}
+
+SKIP=$(if test "${SKIP_CERT_VERIFY_EXTERNAL}" = "true" ; then echo "--insecure" ; fi)
+
+status "Waiting for UAA to be available at ${HCF_UAA_INTERNAL_URL}/token_key ..."
+retry_forever 10s curl --connect-timeout 5 --fail $SKIP "${HCF_UAA_INTERNAL_URL}/token_key"
```
mook-as: `retry_forever` now repeatedly prints out the command (which is the URL), so we probably don't need it in the status.

container-host-files/etc/hcf/config/scripts/fetch_uaa_verification_key.sh
```diff
@@ -7,19 +7,50 @@
...
+}
+
+# helper function to retry a command several times, with a delay between trials
+# usage: retry <max-tries> <delay> <command>...
+function retry_forever () {
+    delay=${1}
+    shift 1
+
+    while true ; do
+	printf "Trying: %s\n" "$*"
```
mook-as: Do we need to print the command list _every_ loop?

src/hcf-release/jobs/uaa-create-user/templates/run.erb
```diff
@@ -32,9 +43,13 @@ function retry () {
     shift 2
 
     while test ${i} -lt ${max} ; do
+	printf "Trying: %s\n" "$*"
         if "$@" ; then
+	    status ' SUCCESS'
             break
         fi
+	trouble '  FAILED'
+	status "Waiting ${delay} ..."
```
mook-as: I think we might be going the other way and being _too_ verbose here.  Printing the full command once, and use `--silent`, is probably enough?

container-host-files/etc/hcf/config/scripts/fetch_uaa_verification_key.sh
andreas-kupries: Yes. Forgot to update the comment after changing the function. Will fix.

andreas-kupries: Done with commit 41fd25bf9092658f6350c1201f2abb62805b3e4b

container-host-files/etc/hcf/config/scripts/fetch_uaa_verification_key.sh
andreas-kupries: I wanted the log to show something for every iteration, yes.
So that we can see the role is still doing something.
Printing the command was not only the easiest, but also the surest way. As the command itself my not produce output.
There also is the issue of the command scrolling of screen if we have enough iterations and thus making it more difficult to see the url. Especially if we start looking at the log late when role is not coming up. (The available delay may not be enough to scroll far enough back before the new output snaps it back to the bottom).

container-host-files/etc/hcf/config/scripts/fetch_uaa_verification_key.sh
andreas-kupries: Will fix. I suspect I have different emacs setting than whoever wrote or edited last. I will go for full spaces-only.

container-host-files/etc/hcf/config/scripts/fetch_uaa_verification_key.sh
andreas-kupries: I don't know. I do not remember writing this code. ... Possibly for the places where ERB templates come in, to separate them better from the actual text.

container-host-files/etc/hcf/config/scripts/fetch_uaa_verification_key.sh
andreas-kupries: True. The exact action to be taken here however depends on if we keep to printing the command in every iteration. Nothing will be done here until that part is sorted.

container-host-files/etc/hcf/config/scripts/fetch_uaa_verification_key.sh
andreas-kupries: Done with commit f2d41d2ad39aac93cdc2cc2da6502fc83ade52d6

src/hcf-release/jobs/uaa-create-user/templates/run.erb
andreas-kupries: The problem with `--silent` is that it also suppresses any error messages from the curl. And then command fails I really want to know that for evaluation of the problem.

And printing something every iteration is IMHO a good thing to show that the role is active, even when not live and ready.

mook-as: Ah. So `--silent` is _too_ silent.

The main thing I'm worried about re: printing every iteration is that it would make it hard to actually look at the messages before it succeeds (because the buffer might run out or something).  If that's not an issue, I guess it's fine?

andreas-kupries: In my testing it actually looked quite nice, and not too cluttered, as I had feared.
So, no, for me the output was not an issue.

Ok. I will remove the urls from the status messages, and leave that to the printed command.
I keep printing the command.

mook-as: 👍

container-host-files/etc/hcf/config/scripts/fetch_uaa_verification_key.sh
andreas-kupries: Handled with commit e1ae256d5bfa0adc10975a4ab3fb9a92f9a51e87

src/hcf-release/jobs/uaa-create-user/templates/run.erb
andreas-kupries: Done with commit e1ae256d5bfa0adc10975a4ab3fb9a92f9a51e87

src/hcf-release/jobs/hcf-set-proxy/templates/run.erb
```diff
@@ -49,9 +63,13 @@ function retry () {
     shift 2
 
     while test ${i} -lt ${max} ; do
+	printf "Trying: %s\n" "$*"
         if "$@" ; then
+	    status ' SUCCESS'
             break
         fi
+	trouble '  FAILED'
+	status "Waiting ${delay} ..."
```
mook-as: FWIW, you still have a mix of space/tab here, I think.

andreas-kupries: Will fix.

andreas-kupries: Done with commit 67774ca1cd77b267c39f6fab9494204a5d2260fb

PR #983, opened 2017-06-19
Add iptables rule to allow connecting to docker from outside vagrant box

jimmykarily:
https://docs.docker.com/engine/userguide/networking/default_network/container-communication/

Trying to connect to the running cf instance from outside the vagrant box fails because of the iptables rules in the box. E.g.

Running `cf api --skip-ssl-validation https://api.cf-dev.io` fails (api.cf-dev.io does resolve to 192.168.77.77). 

I think the above link explains the issue and the solution. We need to `ACCEPT` the FORWARD chain. I'm not sure if this is the right place to set this rule. Also, I only tried this but setting `run: "always"` and running `vagrant reload` to avoid running everything from scratch (described [here](https://www.vagrantup.com/docs/provisioning/basic_usage.html#run-once-or-always)).

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -27,7 +27,7 @@ Vagrant.configure(2) do |config|
 
   config.vm.provider "virtualbox" do |vb, override|
     # Need to shorten the URL for Windows' sake
-    override.vm.box = "https://cf-opensusefs2.s3.amazonaws.com/vagrant/scf-virtualbox-v2.0.1.box"
+    override.vm.box = "https://cf-opensusefs2.s3.amazonaws.com/vagrant/scf-virtualbox-v2.0.3.box"
 
     # Customize the amount of memory on the VM:
     vb.memory = vm_memory.to_s
@@ -43,7 +43,7 @@ Vagrant.configure(2) do |config|
 
 # Currently not built for vmware_fusion
 # config.vm.provider "vmware_fusion" do |vb, override|
-#   override.vm.box="https://cf-opensusefs2.s3.amazonaws.com/vagrant/scf-vmware-v2.0.1.box"
+#   override.vm.box="https://cf-opensusefs2.s3.amazonaws.com/vagrant/scf-vmware-v2.0.3.box"
 #
 #   # Customize the amount of memory on the VM:
 #   vb.memory = vm_memory.to_s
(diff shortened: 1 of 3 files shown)


mook-as:
This actually means we should drop part of #968 because it's solving it wrong, too (given upstream docker documentation). Yay!

jimmykarily:
After running `vagrant halt` and `vagrant up` the iptables seem to be back at `-P FORWARD DROP`. Something must be overwriting the rules on every boot. Our script runs only when provisioning (first time boot).

jimmykarily:
and after that it works just find with the following rules:

```
vagrant@vagrant-kube:~/scf> sudo iptables -S
-P INPUT ACCEPT
-P FORWARD DROP
-P OUTPUT ACCEPT
-N DOCKER
-N DOCKER-ISOLATION
-N KUBE-FIREWALL
-N KUBE-SERVICES
-A INPUT -j KUBE-FIREWALL
-A FORWARD -o docker0 -m comment --comment "Fix exposed services" -j ACCEPT
-A FORWARD -j DOCKER-ISOLATION
-A FORWARD -o docker0 -m conntrack --ctstate RELATED,ESTABLISHED -j ACCEPT
-A FORWARD -o docker0 -j DOCKER
-A FORWARD -i docker0 ! -o docker0 -j ACCEPT
-A FORWARD -i docker0 -o docker0 -j ACCEPT
-A OUTPUT -m comment --comment "kubernetes service portals" -j KUBE-SERVICES
-A OUTPUT -j KUBE-FIREWALL
-A DOCKER-ISOLATION -j RETURN
-A KUBE-FIREWALL -m comment --comment "kubernetes firewall for dropping marked packets" -m mark --mark 0x8000/0x8000 -j DROP
```

I suspect that initially I didn't have the commit from #968. After I provisioned the vm, I pulled the "develop" branch but didn't provision the box again so I though I had the fix although I didn't.

So it turns out that it is already fixed. It doesn't match exactly the docker documentation but I think this is not very important. Maybe we can change this line:

mook-as:
Okay, so that means _this_ PR as-is shouldn't be merged (since we need it in the systemd unit), but we should still fix the change to use the upstream-documented method.  I'll update the PR and you can review after 😀

mook-as:
Oh, for testing without grabbing a whole new box image (1GB download, once I finish building it), just modify the file in `/etc/systemd/system/kubelet.service.d/` and do a `vagrant reload`.

jimmykarily:
This is better, tried and it works fine. I fixed up your commit and pushed. LGTM.

PR #988, opened 2017-06-19
Mark cert for talking to UAA as secret. 

andreas-kupries:
Further remove duplicate `internal` marker.


andreas-kupries:
Oh. From the name (`xxx_CERT`) I thought it was something non-public.

mook-as:
It's… sort of public? It's the thing the UAA certificate was signed with, so it's externally relevant (at least, not part of SCF) anyway. But nobody other than UAA consumers is likely to care about it.  (If it had been `xxx_KEY` it would definitely be a secret.)

Eh, I guess I'll merge anyway, no harm, and it's the last cert.

PR #993, opened 2017-06-20
Match scripts and targets. Import same for UFR (matching to SCF).

andreas-kupries:
https://trello.com/c/eX0agPsP/189-clean-up-the-names-of-makefile-targets-and-scripts
See also https://github.com/SUSE/uaa-fissile-release/pull/24


andreas-kupries:
Rebased/resolved, with bump of submoduled UFR to rebased/resolved version

PR #997, opened 2017-06-20
Bump uaa-fissile-release for package-kube fix

alexjh:


mook-as:
This depends, essentially, on SUSE/uaa-fissile-release#21

PR #1007, opened 2017-06-26
Add instructions on submodules updating on the README

jimmykarily:


jimmykarily:
We tried to be a little more detailed but we have hard time verifying that the new image that is created actually runs the updated code. We can't find the git version tag anywhere (image tags etc). Is there a way to check that?

README.md
```diff
@@ -315,6 +359,41 @@ You can access any URL or endpoint that references this address from your host.
...
+First of all checkout the desired commit:
+
+```
+host> cd src/loggregator/ && git checkout v81
+```
+
+Then from inside the vagrant box regenarate the image for this release:
+
+```
+vagrant> cd scf && make loggregator-release compile images kube run
```
mook-as: It looks like you didn't mean to have `kube run` here, since we're doing it in the next step?

README.md
```diff
@@ -315,6 +359,41 @@ You can access any URL or endpoint that references this address from your host.
 1. Make a change to component `X`, in its respective release (`X-release`).
 1. Run `make X-release compile images run` to build your changes and run them.
 
+#### Bumping a version in a release (or just make a change)
+
+For this example, lets suppose we want to update a release to a later tag.
+First of all checkout the desired commit:
+
+```
+host> cd src/loggregator/ && git checkout v81
```
mook-as: You may also need `git submodule update --init --recursive` / `git submodule sync --recursive`.  Possibly over and over until it stabilizes.

README.md
```diff
@@ -157,6 +178,29 @@ Typically Vagrant box deployments encounter one of few problems:
...
+  `make run` to bring this up.
+
+* Pulling images during any of `vagrant up` or `make vagrant-prep` or `make docker-deps`
+  fails.
+
+  Make sure you have enabled ipv4 ip fowarding.
+
+  Temporarily:
+
+  ```echo "1" | sudo tee /proc/sys/net/ipv4/ip_forward```
```
mook-as: Odd, I thought docker did this automatically. Does it make sense to do this automatically (as a `docker.service` drop-in `ExitStartPre`)?

README.md
```diff
@@ -114,6 +114,27 @@ a working system.
...
+3. Pulling updates
+
+    When you want to pull the latest changes from the upstream you should:
+
+    ```
+    # Pull the changes (or checkout the commit you want):
+    git pull
+
+    # Update all submodules to match the checked out commit
+    git submodules update --recursive
```
mook-as: Sometimes `git submodule sync --recursive` is necessary as well. (And sometimes we need to alternate the two commands until they unblock each other).

jimmykarily: Didn't know about `git submodule sync` before but reading the docs now seems weird that we need it ([docs](https://git-scm.com/docs/git-submodule/1.6.5#git-submodule-sync)). The remote urls for our submodules should change that often (if ever). Are you sure this is needed? I would like to add a comment on why this might be the case.

README.md
jimmykarily: According to the [docs](https://docs.docker.com/engine/userguide/networking/default_network/container-communication/#communicating-to-the-outside-world) it should be the default. I don't know why it isn't in our case but I think we don't only need to enable this inside the vagrant box but also in the host system. If I recall correctly I didn't have access to the internet from inside the vagrant box when the host's value was not set to "1".

README.md
jimmykarily: You are right, `make run` also. Removing these 2.

README.md
mook-as: Ah, if it's required on the host then it could make sense. In that case we'd have to make that part clearer. `docker.service` should have fixed it, but it's possible that (since the host is long-running) it got reset by something else (e.g. `systemctl restart NetworkManager.service`).

README.md
mook-as: It's needed when we bump the BOSH release submodules, since sometimes they change (e.g. moving from cloudfoundry-incubator to cloudfoundry).  Not often, though.

README.md
jimmykarily: So, until we know better what happens inside the vagrant box, I will just state that it needs to be set both on the host and in the vagrant box. People will know at least what to check regardless of why it was not set in the first place.

README.md
jimmykarily: Added.

README.md
jimmykarily: If you run these commands then the submodule is back to the commit is was before. We intentionally leave this in a "dirty" state because this scenario is about changing stuff in submodules during development. There is the "If everything works" section below that explains how to commit this if it's working. Do you think that is enough?

mook-as: The submodule might itself have more submodules (in fact, it probably does unless it's loggregator, since they have separate `foo` and `foo-release` repos).  Basically, I'm saying that just checking out the top level submodule isn't going to get you the correct source code for `bosh create release`.

jimmykarily: Oh you are right, didn't know that there where "nested" submodules. Added text for this.

mook-as:
I haven't merged yet since it looks like you intend to squash your commits before merging. Let me know if you want me to push the merge button.

README.md
```diff
@@ -127,6 +127,45 @@ a working system.
...
+    ```
+      git submodule sync --recursive
+    ```
+
+    You might have to run the `git submodules update --recursive` again after the
+    last command.
+
+    If there are untracked changes from submodule directories you can safely remove them.
+
+    E.g. A command that will update all submodules and drop any changed or untracked files in them is:
```
jimmykarily: I added this example because sometimes when changing branches it is really hard to get rid of the untracked files and everything. I think the warning is also necessary here.

jimmykarily:
Squashed, rebased and added [one more change](https://github.com/SUSE/scf/pull/1007/files#diff-04c6e90faac2675aa89e2176d2eec7d8R154). If that is ok, feel free to merge.

PR #1009, opened 2017-06-29
Use public network and pull out scripts from Vagrantfiles

HartS:
This is part of the work required for https://trello.com/c/6V8qrrox/181-re-use-the-packer-configs-for-standalone-k8s

This puts the SCF provisioning steps which would be needed for non-vagrant CI deploys into scripts which vagrant can then use when provisioning machines.

As part of this work, it's expected that dev, build, and CI/test deployments of SCF will be on K8s hosts which have an IP which is not just accessible from the host. For libvirt+kvm, this means the user will have to create a host bridge interface if they don't already have one. For virtualbox, this should just work.


Vagrantfile
```diff
@@ -9,26 +9,31 @@ Vagrant.configure(2) do |config|
...
-  # config.vm.network "forwarded_port", guest: 8501, host: 8501
-
   vm_memory = ENV.fetch('VM_MEMORY', 10 * 1024).to_i
   vm_cpus = ENV.fetch('VM_CPUS', 4).to_i
 
+  os=`uname`.strip
+  if os == 'Darwin'
+    default_if=`route get default | grep interface`.split(" ").last
+  else
+    default_if=`/sbin/route | grep default`.split(" ").last
```
HartS: This point is still relevant after refactoring I did: See https://github.com/SUSE/scf/blob/d465164bea964412c34ddce4c7ef1b704bf20071/Vagrantfile#L15

My thoughts were that we're going to be shelling out either way. I get the point about sticking to ruby, but since I'll be using snippets from here for the non-vagrant CI deploys (mostly bash), I went with whatever was easiest (also the mixed-language approach is shorter and more readable to me). Let's talk more about it on Tuesday.

Vagrantfile
```diff
@@ -9,26 +9,31 @@ Vagrant.configure(2) do |config|
...
+  end
+  # Ugly hack to warn user about not using a host bridge with libvirt
+  user_warned_about_bridge = false
+
+  net_config = {
+    :use_dhcp_assigned_default_route => true
```
HartS: Thanks, I forgot ruby did this magic conversion under-the-hood for configuration hashes for all hash expressions/declarations

But this isn't relevant to the current state, since there was no network configuration common to all deployments in the end.

Vagrantfile
```diff
@@ -9,26 +9,31 @@ Vagrant.configure(2) do |config|
...
-  config.vm.network "private_network", ip: "192.168.77.77"
 
   config.vm.provider "virtualbox" do |vb, override|
     # Need to shorten the URL for Windows' sake
-    override.vm.box = "https://cf-opensusefs2.s3.amazonaws.com/vagrant/scf-virtualbox-v2.0.3.box"
-
+    override.vm.box = "https://cf-opensusefs2.s3.amazonaws.com/vagrant/scf-virtualbox-v2.0.5.box"
+    net_config[:nic_type] = "virtio"
+    net_config[:bridged] = default_if
+    override.vm.network "public_network", net_config
```
HartS: There's a lot of branching in the ways the net_config needs to be changed depending on Linux/OSX, bridged/NAT, DHCP/no-DHCP (only relevant for NAT), libvirt/vbox. I think the configuration hash needs to be constructed piecemeal, otherwise we'll have more conditional branches than necessary.

bin/common/usr_local_bin_to_path.sh
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env bash
+
+# Add /usr/local/bin to non-login path, since tools are installed there
+if ! grep ENV_SUPATH /etc/login.defs | grep -q /usr/local/bin; then
+  sed -i '/ENV_SUPATH/s/$/:\/usr\/local\/bin/' /etc/login.defs
```
mook-as: Is `/etc/environment.d` insufficient for this? Which non-login path are we talking about here?

bin/common/install_tools.sh
```diff
@@ -0,0 +1,47 @@
+#!/usr/bin/env bash
+set -o errexit -o nounset
+
+# Get version information and set destination dirs
+. "$(dirname "$0")/versions.sh"
+
+# Installs tools needed to build and run HCF
+SCF_BIN_DIR="${SCF_BIN_DIR:-/home/vagrant/bin}"
```
mook-as: Oh, this is a giant lie, and it's set in `versions.sh`? Then drop the lines here (and maybe add a comment if you want).

bin/common/versions.sh
```diff
@@ -57,3 +57,11 @@ then
...
     echo
 fi
+
+if id -u vagrant >& /dev/null; then
+  SCF_BIN_DIR="${SCF_BIN_DIR:-/usr/local/bin}"
+  SCF_TOOLS_DIR="${SCF_TOOLS_DIR:-/home/vagrant/tools}"
+else
+  SCF_BIN_DIR="${SCF_BIN_DIR:-output/bin}"
+  SCF_TOOLS_DIR="${SCF_TOOLS_DIR:-output/tools}"
+fi
```
mook-as: This makes no sense here, it's not a version number. Please don't put random required-for-functioning things here, it just confuses us down the road.

bin/common/write_k8s_ip_to_environment.sh
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env bash
+
+set -o errexit
+while ! ifconfig eth1 | grep -q addr; do
+  sleep 3
+done
+K8S_VM_IP=$(ifconfig eth1 | grep -oE 'addr:[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}' | sed 's/addr://g')
+echo K8S_VM_IP=$K8S_VM_IP >> /etc/environment
```
mook-as: This is entirely the wrong place for it. The IP address could change on reboot, at which point this makes no sense.

bin/common/write_k8s_ip_to_environment.sh
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env bash
+
+set -o errexit
+while ! ifconfig eth1 | grep -q addr; do
+  sleep 3
+done
+K8S_VM_IP=$(ifconfig eth1 | grep -oE 'addr:[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}' | sed 's/addr://g')
```
mook-as: Is this more readable?
```bash
ip -4 addr show eth1 | awk '/inet/ { print $2 }' | cut -d/ -f1
```

How sure are we that we want `eth1`?

Vagrantfile
```diff
@@ -5,13 +5,82 @@
...
+  `/sbin/route | grep default`.split(" ").last
+end
+
+# Returns the name of the default interface
+def default_if
+  @default_if_memo ||= os_is_osx? ? default_if_osx : default_if_linux
+end
+
+def interface_is_bridge?(interface)
+  `/usr/sbin/brctl show | cut -f1 | grep '^#{interface}$'`.strip.length > 0
```
HartS: This is only called from `bridged_net_linux?`. 'brctl' doesn't exist on OSX. Right now bridging on OSX is disabled.

I'll see if I can find a way to get the same information without shell

Vagrantfile
```diff
@@ -5,13 +5,82 @@
...
+
+# While VBox on Linux allows bridging without a host device, there's no easy way
+# to check the provider from the Vagrantfile (provider-specific config blocks
+# still run on the wrong providers) so assume libvirt on linux deployments
+#
+# It's also worth noting that the Vagrant libvirt provider (KVM) typically uses
+# 'direct' type interfaces for public_network interfaces. While these *do* work
+# without a host bridge interface, the host VM is unable to route to IPs the VM
+# receives this way. See https://libvirt.org/formatnetwork.html#examplesDirect
+def host_bridge_available?
```
HartS: So you're saying, avoid doing any validation of whether the user can do what they're trying to do (regarding bridged networking, etc.)?

Vagrantfile
```diff
@@ -5,13 +5,82 @@
...
+  elsif ! default_if
+    puts "No default interface detected"
+  elsif ! File.file? "/usr/sbin/brctl"
+    puts "'brctl' tool not found. Have you installed bridge-utils"
+  elsif ! interface_is_bridge?(default_if)
+    warning_env = {
+      "COMMAND"    => "vagrant up --provider=libvirt",
+      "DEFAULT_IF" => default_if
+    }
+    puts system(warning_env, "bin/common/warn_no_bridge.sh")
```
HartS: I thought this same warning might be useful for the CI deployment. After seeing how many different deployment scenarios there are, it might make more sense to just document the requirements.

Vagrantfile
```diff
@@ -5,13 +5,82 @@
...
+def bridged_net_osx?
+  false
+end
+
+def bridged_net?
+  @bridged_net_memo ||= os_is_osx? ? bridged_net_linux? : bridged_net_osx?
+end
+
+def net_dhcp?
+  ENV.fetch("VAGRANT_DHCP", false) ? true : bridged_net?
```
HartS: Thanks, I didn't know that was an option.

Vagrantfile
```diff
@@ -21,13 +90,34 @@ Vagrant.configure(2) do |config|
...
+    # bridged networking uses type: bridged (even though the virtual interface still
+    # gets its IP from dhcp. If not using dhcp, the VM will use the 192.168.77.77 IP
+    if ENV.fetch("VAGRANT_DHCP", false)
+      vb_net_config[:type] = "dhcp"
+    else
+      vb_net_config[:ip] = "192.168.77.77"
+    end
+  end
+  # Create a clone of this, otherwise it gets mutated in both providers' sections
+  libvirt_net_config = vb_net_config.clone
```
HartS: Let me know if you think of something. The only alternative I could think of would be to repeat the common configuration in all provider-specific blocks. I didn't like either solution

Vagrantfile
HartS: Maybe it's easiest to require the user to set the name of the linux bridge interface for KVM deployments. Like `export VAGRANT_KVM_BRIDGE=br0`

bin/common/create_kubedns.sh
```diff
@@ -0,0 +1,20 @@
+#!/usr/bin/env bash
+
+echo "Waiting for kubectl to respond..."
+while ! kubectl get pods --all-namespaces; do
```
HartS: Probably if it takes a long time. I typically see `==> default: No resources found.` once. I'll silence this

bin/common/create_kubedns.sh
```diff
@@ -0,0 +1,20 @@
...
+  sleep 2
+done
+echo "Waiting for kube-apiserver to be active..."
+while ! systemctl is-active kube-apiserver.service 2>/dev/null >/dev/null; do
+  sleep 10
+done
+# Due to timing, kube-system may not exist immediately
+while ! kubectl get ns kube-system >& /dev/null; do
```
HartS: Probably, but how are the two different here?

bin/common/create_kubedns.sh
```diff
@@ -0,0 +1,20 @@
+#!/usr/bin/env bash
+
+echo "Waiting for kubectl to respond..."
+while ! kubectl get pods --all-namespaces; do
+  sleep 2
+done
+echo "Waiting for kube-apiserver to be active..."
+while ! systemctl is-active kube-apiserver.service 2>/dev/null >/dev/null; do
```
HartS: I must have accidentally swapped the order here from https://github.com/SUSE/scf/blob/develop/packer/Vagrantfile.template#L75-L82 .  Nice catch!

bin/common/start_docker_service.sh
```diff
@@ -0,0 +1,4 @@
+#!/usr/bin/env bash
+
+sed -i 's/DOCKER_OPTS="\(.*\)"/DOCKER_OPTS="\1 --storage-driver=overlay2"/' /etc/sysconfig/docker
```
HartS: I probably took it out because the `--host $IP` flag was added, but I think you mentioned later that we might not need the docker engine to bind to an external interface anymore (I never modified it to do that).

bin/common/start_k8s_services.sh
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env bash
+
+set -o errexit
+mkdir -p /var/lib/kubelet
+mkdir -p /tmp/hostpath_pv
+mkdir -p /home/vagrant/.fissile
+systemctl daemon-reload
```
HartS: I don't think it does anymore.

bin/common/usr_local_bin_to_path.sh
HartS: If you take a look at https://github.com/SUSE/scf/blob/d96c25ea87d6e7a0838e0d9aa718566e50215df5/packer/Vagrantfile.template#L40-L43, I'm running [create_k8s_certs.sh](https://github.com/SUSE/scf/blob/d96c25ea87d6e7a0838e0d9aa718566e50215df5/bin/common/create_k8s_certs.sh) right after this. Certstrap (and everything else now) is getting installed to `/usr/local/bin`, but `/usr/local/bin` doesn't go in the path by default when run via a non-login shell (like when using the shell provisioners). The alternatives were putting these utilities in one of `/usr/bin:/bin:/usr/games` (the paths common to both vagrant and root non-login shells), all of which seemed less appropriate

bin/common/versions.sh
HartS: This was actually based on your feedback, about *not* putting things in `/usr/local/bin` for users who might run the `install_tools` scripts locally. Presumably no vagrant user would exist in such cases (so it makes sense to put things in the scf repo)

bin/common/write_k8s_ip_to_environment.sh
HartS: With both vagrant and CI deployments, there will be an eth1 which is the default route.

I can use `ip -4 addr show eth1 | awk '/inet/ { print $2 }' | cut -d/ -f1` to get the IP instead.

bin/common/write_k8s_ip_to_environment.sh
HartS: If the IP address changes, all of our kube stuff is going to break anyway. Where would you suggest putting it instead?

Vagrantfile
mook-as: Yep, since it'll be opt-in anyway.  We can try to have better error messages on failure or something, but all this dance to get `brctl` seems way too much.

Vagrantfile
mook-as: I _think_ a helper function to construct the base net config would be slightly better (because it never pretends to be the vbox net config). But not by much.

bin/common/create_kubedns.sh
mook-as: Oh, didn't realize `>&` worked too. As opposed to writing to a file named `&` (which it doesn't do).  Nevermind then!

bin/common/install_tools.sh
```diff
@@ -0,0 +1,47 @@
...
+SCF_TOOLS_DIR="${SCF_TOOLS_DIR:-/home/vagrant/tools}"
+cf_url="${cf_url:-https://cli.run.pivotal.io/stable?release=linux64-binary&version=${CFCLI_VERSION}&source=github-rel}"
+kubectl_url="${kubectl_url:-https://storage.googleapis.com/kubernetes-release/release/v${KUBECTL_VERSION}/bin/linux/amd64/kubectl}"
+k_url="${k_url:-https://github.com/aarondl/kctl/releases/download/v${K_VERSION}/kctl-linux-amd64}"
+kk_url="${kk_url:-https://gist.githubusercontent.com/jandubois/${KK_VERSION}/raw/}"
```
mook-as: Ignore this, PRed in #1017

bin/common/start_docker_service.sh
mook-as: Right, but there's no `--host $IP` here at least.  I guess you inherited bad code from me :)

bin/common/usr_local_bin_to_path.sh
mook-as: Right, I thought `/etc/environment.d/` is done via `pam_env`, which should be on login… I'm assuming by _non-login_ you mean _non-interactive_ here, though.  Not sure how else you'd get different privileges otherwise (since IIRC vagrant privileged shells are done via `sudo`)

bin/common/versions.sh
mook-as: Right, I'm just saying not in this _file_. We should still have the switching logic, just elsewhere :)

bin/common/write_k8s_ip_to_environment.sh
mook-as: I think I'd be happier with a comment about eth1 being valid in both places.  Only to document that we thought about it :)

bin/common/write_k8s_ip_to_environment.sh
mook-as: Hmm, actually, I don't see this being used anywhere but cert generation.  So, that script?
But we might want to run the cert generation as a `ExecStartPre` in that case.  Or rather as a thing both apiserver and kublet and whatnot depends on. Which would be sad, but ah well.

bin/common/usr_local_bin_to_path.sh
HartS: Based my understanding, the privileged provisioning commands are run in shells which are both non-login, and non-interactive. `/etc/profile` still gets sourced on non-interactive login shells, but when you run something with sudo, profile and rc scripts are not sourced because it's not a login shell. Otherwise `/etc/profile` would probably be a better place to modify the path.

bin/common/write_k8s_ip_to_environment.sh
HartS: If we run the cert generation in ExecStartPre, I'm pretty sure that will break already-existing deployments between restarts. Do you know if we actually *need* the IP to go into the certificate? At least right now, since apiserver binds to 127.0.0.1 insecurely (but surprisingly actually binds to 0.0.0.0), I'm not sure if we need the cert to contain the 'public' IP. I think the internal communication happens over one of the k8s internal domains.

mook-as:
I'm having trouble using this, and haven't figured out why yet. `vagrant up` gets stuck at various points (currently, "install tiller for helm"); `vagrant ssh` in while it's stuck just hangs too. I _can_ however login on tty1 as root, and that seems to work. Low CPU use (~5%), not very much memory in use (~2G of 10G), low disk I/O (~nothing), some network IO (basically nothing).

It looks like listing `~/vagrant` has issues. No btrfs-related warnings in dmesg (probably because there are no btrfs things mounted). It looks like NFS is having trouble? The host side has logs saying everything is authenticated and the NFS connection is allowed.  Doesn't look like anything is being logged in the VM side.

libvirt, @ 17035b5f0f4002214778bd9db7375f8c6aa69d38

bin/common/create_kubedns.sh
HartS: I *think* there could be a difference between the two? But I don't really understand what it is, I was hoping you did :P

HartS:
@mook-as You're deploying on kvm, right? Are you deploying on a remote KVM host? I think we've already talked about how the NFS mounts won't work remotely... I was able to disable those by commenting out the lines which set it up in the project Vagrantfile, and adding a shell provisioner to clone the SCF repo before the other shell provisioners in the .box Vagrantfile.

If deploying to a local KVM host with bridged networking, what does your host networking look like?

mook-as:
Yes, KVM, but locally. Well, I'm technically on a remote host, but that's over the SSH session; so as far as vagrant is concerned, it's local.  I'm just using default things, so I _don't_ think it's bridged networking.

Host:
```
1: lo    inet 127.0.0.1/8 scope host lo\       valid_lft forever preferred_lft forever
2: eth1    inet <lan>/14 brd <lan> scope global dynamic eth1\       valid_lft 565sec preferred_lft 565sec
65774: virbr0    inet 192.168.122.1/24 brd 192.168.122.255 scope global virbr0\       valid_lft forever preferred_lft forever
65776: virbr2    inet 192.168.100.1/24 brd 192.168.100.255 scope global virbr2\       valid_lft forever preferred_lft forever
65778: virbr1    inet 192.168.42.1/24 brd 192.168.42.255 scope global virbr1\       valid_lft forever preferred_lft forever
65780: virbr3    inet 192.168.121.1/24 brd 192.168.121.255 scope global virbr3\       valid_lft forever preferred_lft forever
65782: virbr5    inet 192.168.77.1/24 brd 192.168.77.255 scope global virbr5\       valid_lft forever preferred_lft forever
```

VM:
```
vagrant-kube:~ # ip -4 -o a
1: lo    inet 127.0.0.1/8 scope host lo\       valid_lft forever preferred_lft forever
2: eth0    inet 192.168.121.5/24 brd 192.168.121.255 scope global eth0\       valid_lft forever preferred_lft forever
3: eth1    inet 192.168.77.77/24 brd 192.168.77.255 scope global eth1\       valid_lft forever preferred_lft forever
4: docker0    inet 172.17.0.1/16 scope global docker0\       valid_lft forever preferred_lft forever
```

HartS:
I wonder if that's because of Virtio. I had to avoid it on vbox, but I don't remember if I tested non-bridged deployment on KVM.

Can you try commenting out `libvirt_net_config[:nic_model_type] = "virtio"`?

mook-as:
`virtio` is the default NIC type on my system. (And it works fine in old-vagrant, and I can ssh to the machine via the interface fine.)

bin/common/create_k8s_certs.sh
```diff
@@ -0,0 +1,14 @@
+#!/usr/bin/env bash
+
+K8S_VM_IP=${K8S_VM_IP:-192.168.77.77}
+certstrap --depot-path "/run/certstrap" init --common-name "CA.kube.vagrant" --passphrase "" --years 10
```
HartS: The path doesn't exist, but certstrap does a `mkdir -p` basically. I can add an explicit `mkdir -p` for readability though.

mook-as: Nah that's fine. As long as it works :)

bin/common/start_k8s_services.sh
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env bash
+
+set -o errexit
+mkdir -p /var/lib/kubelet
+mkdir -p /tmp/hostpath_pv
+mkdir -p /home/vagrant/.fissile
```
HartS: Does hostpath_pv get created automatically when the storageclass is created?

bin/common/write_k8s_ip_to_environment.sh
HartS: Tested cert generation without any special IPs and everything seems to work... so I should maybe just shove everything back into install_kubernetes and call it from the *packer* provisioner again

bin/common/start_k8s_services.sh
mook-as: No, it gets created automatically on… kubelet? start.  Well, one of the kube services anyway.

bin/common/write_k8s_ip_to_environment.sh
mook-as: If it works fine, awesome!

bin/common/start_k8s_services.sh
HartS: kubelet tries to chown it, but doesn't create it.. so if it isn't created beforehand, kubelet startup fails

bin/common/create_kubedns.sh
jimmykarily: https://stackoverflow.com/a/11255498

Vagrantfile
```diff
@@ -5,13 +5,82 @@
 # configures the configuration version (we support older styles for
 # backwards compatibility). Please don't change it unless you know what
 # you're doing.
+
+
+def os_is_osx?
+  `uname`.strip == 'Darwin'
```
HartS: No more shelling out from the Vagrantfile: https://github.com/SUSE/scf/blob/harts-vagrant-use-public-network-rebased/Vagrantfile

Vagrantfile
```diff
@@ -5,13 +5,82 @@
...
+end
+
+def default_if_osx
```
HartS: I removed all efforts to check the correct setup from the provisioning

Vagrantfile
```diff
@@ -5,13 +5,82 @@
...
+  `route get default | grep interface`.split(" ").last
+end
+
+def default_if_linux
+  `/sbin/route | grep default`.split(" ").last
+end
+
+# Returns the name of the default interface
+def default_if
+  @default_if_memo ||= os_is_osx? ? default_if_osx : default_if_linux
```
HartS: No longer relevant (removed all efforts to check the correct setup from provisioning)

Vagrantfile
HartS: No longer relevant (removed all efforts to check the correct setup from provisioning)

Vagrantfile
```diff
@@ -5,13 +5,82 @@
...
+end
+
+# In theory, vbox could manage a bridged deployment if the interface is not wireless
+# But for now, we'll assume vbox is primarily used for osx deployments, which means
+# a wired interface may not be available
+def bridged_net_osx?
+  false
+end
+
+def bridged_net?
```
HartS: No longer relevant (removed all efforts to check the correct setup from provisioning)

Vagrantfile
HartS: I'm using `ENV.include?` now for `VAGRANT_DHCP`: https://github.com/SUSE/scf/blob/harts-vagrant-use-public-network-rebased/Vagrantfile#L15. `VAGRANT_KVM_BRIDGE` and `VAGRANT_VBOX_BRIDGE` are now expected to have values (though I've still used `ENV.include?` to check if they're set)

Vagrantfile
HartS: I've added `base_net_config()` to set up common config for vbox and kvm. `use_dhcp_assigned_default_route` is now always set, and when not using bridged networking, set the IP to 192.168.77.77 when VAGRANT_DHCP isn't set, otherwise, set `config[:type] => true`: https://github.com/SUSE/scf/blob/harts-vagrant-use-public-network-rebased/Vagrantfile#L10-L25.

bin/common/create_kubedns.sh
HartS: Fixed the order of these: https://github.com/SUSE/scf/blob/harts-vagrant-use-public-network-rebased/packer/scripts/create-kubedns.sh#L5-L12

Vagrantfile
HartS: This is done, options are now:
- `VAGRANT_KVM_BRIDGE`
- `VAGRANT_VBOX_BRIDGE`
- `VAGRANT_DHCP` (when using virtual networking)

If the user supplies these incorrectly, they'll have to figure out what went wrong from Vagrant or the provider's error messages.

mook-as:
Closing, we're doing things in #1021 instead

PR #1016, opened 2017-07-05
Redo service discovery

mook-as:
Use the k8s API to find the relevant pods, and talk to them directly, instead of looking up hosts via DNS SRV.  This means we only talk to ready pods (except localhost, due to entries in `/etc/hosts`), but should help with places with weird / busted DNS setups.

Also renames `HCP_SERVICE_DOMAIN_SUFFIX` to `K8S_SERVICE_DOMAIN_SUFFIX`, and use full DNS for hosts instead of just the host name part (also for busted DNS).

Passed 🐱 locally.


container-host-files/etc/hcf/config/scripts/configure-HA-hosts.sh
```diff
@@ -10,60 +10,62 @@
...
         for (( i = 0 ; i < 60 ; i ++ )) ; do
-            names=($(dig "${component_name}.${HCP_SERVICE_DOMAIN_SUFFIX}" -t SRV | awk '/IN[\t ]+A/ { print $1 }'))
-            if test "${#names[@]}" -gt 0 ; then
+            hosts="$(k8s_api /pods | jq -crM '
+                .items |
+                map(
+                    select(.metadata.labels."skiff-role-name" == "'"${component_name}"'") |
+                    .metadata.name |
+                    select(.) |
+                    . + "'".${component_name}-pod.${K8S_SERVICE_DOMAIN_SUFFIX}"'"
```
viovanov: I think we should rename the `-pod` convention to `-set` here: https://github.com/SUSE/fissile/blob/master/kube/stateful_set.go#L47

viovanov:
We should try to use the same approach as this example: https://github.com/kubernetes/charts/blob/master/incubator/etcd/templates/etcd-statefulset.yaml#L34

Which will be made better by deployment with helm charts.

mook-as:
Most of the comments addressed; `-pod` ➡️ `-set` requires SUSE/fissile#258

PR #1021, opened 2017-07-10
Refactor packer building scripts, remove hardcoded IP, and enable bridged networking with Vagrant deploy

HartS:
This branch is a rework of https://github.com/SUSE/scf/pull/1009, with Mark's feedback addressed. It's also based on develop more recently to incorporate work that would have caused conflicts.


Vagrantfile
```diff
@@ -194,31 +175,31 @@ Vagrant.configure(2) do |config|
...
-    vb.vmx["sharedFolder1.enabled"] = "TRUE"
-    vb.vmx["sharedFolder1.readAccess"] = "TRUE"
-    vb.vmx["sharedFolder1.writeAccess"] = "TRUE"
-    vb.vmx["sharedFolder1.hostPath"] = bosh_cache
-    vb.vmx["sharedFolder1.guestName"] = "bosh"
-    vb.vmx["sharedFolder1.expiration"] = "never"
-    vb.vmx["sharedfolder1.followSymlinks"] = "TRUE"
-  end
-end
+# module VMwareHacks
```
mook-as: As before, if we don't support this then just delete it. It's in git history.

bin/common/install_tools.sh
```diff
@@ -0,0 +1,47 @@
...
+# Get version information and set destination dirs
+. "$(dirname "$0")/versions.sh"
+
+# Installs tools needed to build and run HCF
+SCF_BIN_DIR="${SCF_BIN_DIR:-/home/vagrant/bin}"
+SCF_TOOLS_DIR="${SCF_TOOLS_DIR:-/home/vagrant/tools}"
+cf_url="${cf_url:-https://cli.run.pivotal.io/stable?release=linux64-binary&version=${CFCLI_VERSION}&source=github-rel}"
+kubectl_url="${kubectl_url:-https://storage.googleapis.com/kubernetes-release/release/v${KUBECTL_VERSION}/bin/linux/amd64/kubectl}"
+k_url="${k_url:-https://github.com/aarondl/kctl/releases/download/v${K_VERSION}/kctl-linux-amd64}"
+kk_url="${kk_url:-https://gist.githubusercontent.com/jandubois/${KK_VERSION}/raw/}"
```
mook-as: This will probably be a merge conflict? The code has changed on develop.

HartS: The branch base includes the change I think you're referring to

Vagrantfile
```diff
@@ -165,26 +151,21 @@ Vagrant.configure(2) do |config|
...
-    # Install development tools
-    (
-      cd "${HOME}/scf"
-      ${HOME}/bin/direnv exec ${HOME}/scf/bin/dev/install_tools.sh
-    )
-
+    export HOME=/home/vagrant
+    export PATH=$PATH:/home/vagrant/bin
+    cd "${HOME}/scf"
+    bash ${HOME}/scf/bin/common/install_tools.sh
```
HartS: This was originally all part of the installation script that was run privileged from a packer provisioner. I'll see if it still needs to be privileged.. at a glance I don't see that it does

HartS:
The NFS issue is the same one I've seen. I can't figure out what could be causing it though.. as mentioned on Rocketchat, I've been seeing the same issue, but sporadically. After a vagrant reload, it'll then work if I run the same command myself.

bin/common/install_tools.sh
mook-as: Oh, so it does! In that case, you had a mistake in your rebase.
https://github.com/SUSE/scf/pull/1021/files#diff-41460eafa0a29710b9b34aa0d9ee141dL22

The version variable is going to the wrong place.

HartS: This seems to work for me: https://gist.githubusercontent.com/jandubois/40a5b3756cf4bcbed940e6156272c0af/raw

`KK_VERSION` is set in https://github.com/SUSE/scf/pull/1021/files#diff-e482966f3ca1c1f6ed808625edfe39c2R14.. 

Ah! You mean it should now be `576a42386770423ced46ab4ae9955bee59b0d4dd`?

mook-as: I mean that you're mistaking the gist ID for the gist version.  If you change `KK_VERSION`, we should have a different version of the same gist, not the latest version of a different gist.

HartS:
I think pretty much everything has been addressed now, except for, you know, the thing working. I've narrowed down the NFS issues to the box image itself. I'm building develop right now to see if the issues occurs (meaning it might be an issue of movement of packages which are getting installed in the packer build, or possibly some differences in the build platform)

bin/common/install_tools.sh
HartS: OK, I'm with you now. So the original value we had for `KK_VERSION` wasn't a tree hash after all, but the repo hash

Vagrantfile
HartS: This does need to be privileged, because in the vagrant box, the tools are going into `/usr/local/bin`. This is useful because if we want to use any of these tools from a non-interactive, non-login shell, they need to be in a directory which will be in PATH.

HartS:
The newer kernel from the Kernel:stable package seemed to be the cause of the NFS issues we were seeing. I removed those lines from the installation script used in packer building, and rebuilt the image (uploaded to s3://cf-opensusefs2/vagrant/scf-libvirt-v2.0.5-staging.box)

HartS:
I've pushed a commit which explicitly installs the k8s v1.6 packages when building the vagrant box. I've built another box from this image, tested vagrant deployment with it, and updated `scf-libvirt-v2.0.5-staging-staging.box` with it

bin/common/install_tools.sh
```diff
@@ -0,0 +1,47 @@
...
+echo "Making binaries executable ..."
+chmod a+x "${SCF_BIN_DIR}/cf"
+chmod a+x "${SCF_BIN_DIR}/k"
+chmod a+x "${SCF_BIN_DIR}/kk"
+chmod a+x "${SCF_BIN_DIR}/helm"
+
+# Note that we might not have a k8s available; do this only if we're in vagrant
+if systemctl is-active kube-apiserver.service ; then
+  echo "Installing tiller for helm ..."
+  helm init
```
HartS: I added `chown -R vagrant:users /home/vagrant/.helm`, which runs only if the user running the script is root. If run as a non-root user (on a dev machine instead of from vagrant up), I think it'll still do the right thing. I'll try to verify this inside of a VM with no vagrant user though.

bin/dev/install_tools.sh
```diff
@@ -1,90 +1,49 @@
...
-tar -xzf "${tools_dir}/cf.tgz" -C "${bin_dir}"
-
-echo "Fetching kubectl ${kubectl_url} ..."
-wget -q "${kubectl_url}" -O "${bin_dir}/kubectl"
-wget -q "${k_url}" -O "${bin_dir}/k"
-wget -q "${kk_url}" -O "${bin_dir}/kk"
-
-echo "Fetching helm from ${helm_url} ..."
-wget -q "${helm_url}" -O - | tar xz -C "${bin_dir}" --strip-components=1 linux-amd64/helm
+wget -q "$stampy_url"   -O - | tar xz -C "${SCF_BIN_DIR}" stampy
```
HartS: This is run privileged by vagrant up, so no perm issues (I think). It's fine if the binary is root-owned, as long as it's still executable by vagrant. For non-vagrant runs, SCF_BIN_DIR shouldn't be /usr/local/bin

HartS:
This work has merged in the latest work from develop (with conflict resolutions). Since another vagrant box has been published in the meantime, this box is now marked as v2.0.6. `scf-libvirt-v2.0.6-staging.box` has been pushed to s3

HartS:
[Move provisioning steps needed for CI from vagrant deployment into common scripts.](https://trello.com/c/2kvOOQwL/210-move-provisioning-steps-needed-for-ci-from-vagrant-deployment-into-common-scripts)

Vagrantfile
mook-as: If I'm reading things right, both `SCF_BIN_DIR` and `SCF_TOOLS_DIR` are unset at this point; this means they default to the locations in `/home/vagrant`.  And then we proceed to drop a bunch of files there as root.  That is Not Cool — files in the home directory shouldn't be owned by random other people (in this case, root).

```
vagrant@vagrant-kube:~> ls -laR bin tools
bin:
total 3188
drwxr-xr-x  2 vagrant users    4096 Jul 27 18:35 .
drwxr-xr-x 12 vagrant users    4096 Jul 27 18:54 ..
-rwxr-xr-x  1 vagrant users 3255774 May 25 08:14 direnv

tools:
total 6400
drwxr-xr-x  2 root    root     4096 Jul 27 18:35 .
drwxr-xr-x 12 vagrant users    4096 Jul 27 18:54 ..
-rw-r--r--  1 root    root  6543090 Sep 13  2016 cf.tgz

vagrant@vagrant-kube:~> ls -l /usr/local/bin/
total 161972
-rwxr-xr-x 1 root    root     4650644 Jul 26 23:17 certstrap
-rwxr-xr-x 1 root    root    23610336 Aug 10  2016 cf
-rwxrwxr-x 1 vagrant vagrant 63247746 May 17 19:33 helm
-rwxr-xr-x 1 root    root    69968469 Jul 27 18:38 helm-certgen
-rwxr-xr-x 1 root    root     2405672 Jun  9 14:31 k
-rwxr-xr-x 1 root    root        4224 Jul 27 18:35 kk
-rwxr-xr-x 1 root    root     1951738 May 23 17:58 stampy
```
So it looks like `bin/` ends up being owned by `vagrant` (good), but `tools` is owned by `root` (bad).  Also, the bin dir (except for direnv‽) is `/usr/local/bin`, somehow.  And not sure why `vagrant` owns `helm`…

bin/common/versions.sh
```diff
@@ -59,3 +59,11 @@ then
     echo ubuntu '       =' $UBUNTU_VERSION
     echo
 fi
+
+if id -u vagrant >& /dev/null; then
+  SCF_BIN_DIR="${SCF_BIN_DIR:-/usr/local/bin}"
+  SCF_TOOLS_DIR="${SCF_TOOLS_DIR:-/home/vagrant/tools}"
+else
+  SCF_BIN_DIR="${SCF_BIN_DIR:-output/bin}"
+  SCF_TOOLS_DIR="${SCF_TOOLS_DIR:-output/tools}"
```
mook-as: No. _The directories are not versions._  If you want to do this, do it in a file that's **not** named `versions.sh` please.

(I'd say to just set it in the Vagrantfile before running the install_tools scripts, and use the (already-existing) fallback in that script for the non-vagrant case.)

bin/common/install_tools.sh
```diff
@@ -0,0 +1,51 @@
...
+chmod a+x "${SCF_BIN_DIR}/kk"
+chmod a+x "${SCF_BIN_DIR}/helm"
+
+# Note that we might not have a k8s available; do this only if we're in vagrant
+if systemctl is-active kube-apiserver.service ; then
+  echo "Installing tiller for helm ..."
+  if [[ $(id -u) -eq 0 ]] && id -u vagrant &>/dev/null; then
+    sudo -u vagrant helm init
+  else
+    helm init
```
mook-as: We shouldn't ever get here, right? (Because that would be our vagrant box, but not our vagrant box)

So we should just abort instead?

I wonder if we should just do something like
```bash
if kubectl cluster-info 2>/dev/null ; then
  if [[ $(id -u) -eq 0 ]] && id -u vagrant &>/dev/null; then
    sudo --login --user=vagrant helm init
  else
    helm init
  fi
else
  echo "Skipping tiller installation for helm; no local kube found"
fi
```
That would make sense because it's less specific to our vagrant box. Maybe. Dunno, needs more thought.

bin/dev/install_tools.sh
mook-as: 👍 
I'd feel better if we asserted things somehow, but that's probably overkill

bin/dev/install_tools.sh
```diff
@@ -1,90 +1,49 @@
...
-docker run --rm -v "${bin_dir}":/out:rw -v "${helm_certgen_dir}:/go/src/github.com/SUSE/helm-certgen:ro" "golang:${GOLANG_VERSION}" /usr/bin/env GOBIN=/out go get github.com/SUSE/helm-certgen
-if [[ $(stat -c '%u' "${bin_dir}/helm-certgen") -eq 0 ]]; then
-  docker run --rm -v "${bin_dir}":/out:rw "golang:${GOLANG_VERSION}" /bin/chown "$(id -u):$(id -g)" /out/helm-certgen
+docker run --rm -v "${SCF_BIN_DIR}":/out:rw -v "${helm_certgen_dir}:/go/src/github.com/SUSE/helm-certgen:ro" "golang:${GOLANG_VERSION}" /usr/bin/env GOBIN=/out go get github.com/SUSE/helm-certgen
+if [[ $(stat -c '%u' "${SCF_BIN_DIR}/helm-certgen") -eq 0 ]]; then
+  docker run --rm -v "${SCF_BIN_DIR}":/out:rw "golang:${GOLANG_VERSION}" /bin/chown "$(id -u):$(id -g)" /out/helm-certgen
 fi
 mkdir -p "${HOME}/.helm/plugins" # Necessary if we didn't run `helm init`
 rm -rf "${HOME}/.helm/plugins/certgen"
 mv --no-target-directory "${helm_certgen_dir}/plugin" "${HOME}/.helm/plugins/certgen"
```
mook-as: We shouldn't be touching `${HOME}` as root (instead, install helm / its plugin as `vagrant`).

packer/scripts/create-k8s-certs.sh
```diff
@@ -0,0 +1,24 @@
+#!/usr/bin/env bash
```
mook-as: (For the file name) Vlad doesn't like `k8s` and prefers `kube` instead. Don't get it, but meh, might as well.

packer/scripts/install-certstrap.sh
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env bash
+
+set -o errexit -o xtrace
+
+# Certstrap is used for creating the k8s signing keys, which are necessary to avoid communication bugs on host restarts
+echo "Installing certstrap ..."
+# We run chown in docker to avoid requiring sudo
+systemctl start docker.service
```
mook-as: Need to clean up correctly (stop docker and delete the images &c) to avoid dropping turds in the built image.

packer/scripts/create-kubedns.sh
```diff
@@ -0,0 +1,22 @@
...
+# Due to timing, kube-system may not exist immediately
+while ! kubectl get ns kube-system >& /dev/null; do
+  sleep .1
+done
+perl -p -i -e '
+  s@clusterIP:.*@clusterIP: 10.254.0.254@ ;
+  s@170Mi@256Mi@ ;
+  s@70Mi@128Mi@ ;
+' /etc/kubernetes/addons/kubedns.yml
+kubectl create --namespace kube-system --filename /etc/kubernetes/addons/kubedns.yml
```
mook-as: Didn't realize we started up kube within packer‽ That seems… odd (because it requires certs, which means IP addresses… right?)

packer/scripts/install-kubernetes.sh
```diff
@@ -4,6 +4,8 @@
 
 set -o errexit -o xtrace
 
+swapoff -a
```
mook-as: Add comment please (something along the lines of "required by kubernetes" is good enough)

bin/common/install_tools.sh
HartS: Line 45 specifically? Sure, that can be run on some development k8s host which may not have a vagrant user. The vagrant deploy runs this privileged, so if we get here and we're not running as root, or there's no vagrant user, then we're not running the vagrant deploy. So in that case, just `helm init`ing as whatever user ran the script should be fine

packer/scripts/create-kubedns.sh
HartS: I'm not exactly sure how the IPs in the SubjectAltName are used. In any event, the apiserver cert is the only one that got this before (when 192.168.77.77 was assumed). This branch hasn't touched [this](https://github.com/SUSE/scf/blob/harts-vagrant-use-public-network-rebased/packer/scripts/create-k8s-certs.sh#L7) With single-node deployments (such as the vagrant deployment), the other kube services are talking to the apiserver via 127.0.0.1, so maybe this is why things work regardless of the IP used to access the host from outside... I can verify that things all work the same with a `10.` IP.

My understanding is that the vagrant image/deployment is currently not set up in such a way that authenticated client access is even possible. The kube api *is* exposed at https://$IP:6443, and I'm able to communicate with it via kubectl. It's worth noting that cf-dev.io isn't in the apiserver cert's domain list right now either, but from the client machine, if I set `kubernetes` to the right IP in my /etc/hosts, and use `https://kubernetes:6443`, I don't get any errors mentioning the certificate domain. Even still, kubectl prompts for user/password credentials, and I don't believe any user has been configured.

So right now, I think this is out of scope for this PR.

packer/scripts/install-certstrap.sh
HartS: Only the golang image should be downloaded at this point. I thought about cleaning up, but the scripts which run after `vagrant up` use this same image.. I found that `packer build`ing with the image cleaned up vs. without shaves off about 200 MB. However, this is at least 200 MB which would otherwise have to be downloaded on every vagrant up, so I believe the one-time overhead for each vagrantbox outweighs the overhead which cleanup introduces on every `vagrant up` otherwise.

packer/scripts/create-k8s-certs.sh
HartS: I'll do this last because it actually requires a rebuild of the box

bin/dev/install_tools.sh
HartS: Well.. when this is executed as root as part of `vagrant up`, `HOME` is still set to `/home/vagrant`. However, if the user is running this outside of the vagrant deployment, `HOME` should be the right thing for them.

Vagrantfile
HartS: I thought we ended up conceding (maybe a bit begrudgingly) that `/usr/local/bin` was a good place to put executables, because it's not tied to any user that might run them.

The helm owner was `vagrant:vagrant` when extracted by root, because 1000:1000 is apparently the permission in the helm tgz.. to avoid that, I passed the `--no-same-owner` flag to the tar extraction command.

The tools directory is gone now.. as far as I could tell, we weren't using it for anything.. I put the cf.tgz used for extracting the cf binary into /tmp instead

bin/dev/install_tools.sh
mook-as: Right; sorry for not being clear, but I was thinking that the helm-touching bits should be run as a not-root user (in vagrant) so that we don't have to think so hard about permissions issues. (For not-vagrant, we will still end up whatever user it happens to be, which should be fine.)

packer/scripts/install-certstrap.sh
mook-as: I guess this is okay, especially since `/var/lib/docker` isn't a mount anywhere… alright.

packer/scripts/create-kubedns.sh
mook-as: Yeah, I'm not worried at all about the SAN stuff. I was just surprised that we can actually start kubernetes during packer, since I was under the impression that it might write persistent data that would no longer be true post-`vagrant up` (IP address mostly).  But if it doesn't, I guess that's okay…

bin/dev/install_tools.sh
```diff
@@ -1,90 +1,54 @@
 #!/bin/bash
 set -o errexit -o nounset
 
-# Get version information
-. "$(dirname "$0")/versions.sh"
+# Get version information and set destination paths
+. "$(dirname "$0")/../common/versions.sh"
+
+if id -u vagrant >& /dev/null; then
+  SCF_BIN_DIR="${SCF_BIN_DIR:-/usr/local/bin}"
```
mook-as: Can't we just set this in the Vagrantfile where we call this? Then this can stay as a simple `SCF_BIN_DIR=${SCF_BIN_DIR:-output/bin}` and everything should be happy.

HartS: We can, but then all users of the .box will also have to do `export SCF_BIN_DIR=/usr/local/bin` before running this script ( and the one in `bin/common/..`) I'm fine with either one, but setting it here means I don't have to remember to do anything later (for the CI deployment)

packer/scripts/install-certstrap.sh
HartS: I'm glad you brought this up... I think I should put a comment in here anyway to explain why the docker image isn't getting cleaned up right now.. and to help us remember to clean up if our vagrant provisioning ever stops using the same image.

bin/common/install_tools.sh
HartS: I added a comment above this block

Vagrantfile
```diff
@@ -5,13 +5,32 @@
...
+
+def base_net_config
+  base_config = {
+    use_dhcp_assigned_default_route: true
+  }
+  if (ENV.keys & ["VAGRANT_KVM_BRIDGE", "VAGRANT_VBOX_BRIDGE"]).empty?
+    if ENV.include? "VAGRANT_DHCP"
+      # Use dhcp if VAGRANT_DHCP is set. This only applies to NAT networking, as
+      # bridged networking uses type: bridged (even though the virtual interface still
+      # gets its IP from dhcp. If not using dhcp, the VM will use the 192.168.77.77 IP
```
HartS: `)`ノ( ゜-゜ノ)

README.md
```diff
@@ -124,6 +124,42 @@ a working system.
    $ export FISSILE_STEMCELL=splatform/fissile-stemcell-opensuse:$FISSILE_STEMCELL_VERSION
    ```
 
+3. Environment variables to configure `vagrant up` (optional)
+    - `VAGRANT_VBOX_BRIDGE`: Set this to the name of an interface to enable bridged networking when
+      using the Virtualbox provider. Turning on bridged networking will allow your vagrant box to receive
+      an IP accessible anywhere on the network. While Virtualbox is able to bridge over an interface
+      without any special networking configuration (and may even do this on OSX), bridged networking may
+      not be supported when the provided interface is a wireless interface.
```
jimmykarily: Maybe link to the docs here? (maybe [this link](https://www.virtualbox.org/manual/ch06.html#network_bridged)?)

bin/dev/install_tools.sh
HartS: Yes, currently in https://github.com/SUSE/cf-ci/blob/harts-provision-kvm-k8s-host/k8s-kvm/provision_k8s_host_kvm.sh, though it sounds like this will be moving to a Triton-based deploy?

bin/common/install_tools.sh
```diff
@@ -0,0 +1,49 @@
...
+wget -q "${kk_url}" -O "${SCF_BIN_DIR}/kk"
+
+echo "Fetching helm from ${helm_url} ..."
+wget -q "${helm_url}" -O - | tar xz -C "${SCF_BIN_DIR}" --no-same-owner --strip-components=1 linux-amd64/helm
+
+echo "Making binaries executable ..."
+chmod a+x "${SCF_BIN_DIR}/cf"
+chmod a+x "${SCF_BIN_DIR}/k"
+chmod a+x "${SCF_BIN_DIR}/kk"
+chmod a+x "${SCF_BIN_DIR}/helm"
```
mook-as: Missing chmod for kubectl?

bin/dev/install_tools.sh
```diff
@@ -1,90 +1,51 @@
...
 helm_certgen_dir="$(mktemp -d)"
 trap "rm -rf '${helm_certgen_dir}'" EXIT
 git clone --branch "${HELM_CERTGEN_VERSION}" --depth 1 https://github.com/SUSE/helm-certgen.git "${helm_certgen_dir}"
-docker run --rm -v "${bin_dir}":/out:rw -v "${helm_certgen_dir}:/go/src/github.com/SUSE/helm-certgen:ro" "golang:${GOLANG_VERSION}" /usr/bin/env GOBIN=/out go get github.com/SUSE/helm-certgen
-if [[ $(stat -c '%u' "${bin_dir}/helm-certgen") -eq 0 ]]; then
-  docker run --rm -v "${bin_dir}":/out:rw "golang:${GOLANG_VERSION}" /bin/chown "$(id -u):$(id -g)" /out/helm-certgen
+docker run --rm -v "${SCF_BIN_DIR}":/out:rw -v "${helm_certgen_dir}:/go/src/github.com/SUSE/helm-certgen:ro" "golang:${GOLANG_VERSION}" /usr/bin/env GOBIN=/out go get github.com/SUSE/helm-certgen
+if [[ $(stat -c '%u' "${SCF_BIN_DIR}/helm-certgen") -eq 0 ]]; then
+  # The golang docker image is baked into the packer-built image. See packer/scripts/install-certstrap.sh if dependency on this image changes
+  docker run --rm -v "${SCF_BIN_DIR}":/out:rw "golang:${GOLANG_VERSION}" /bin/chown "$(id -u):$(id -g)" /out/helm-certgen
```
mook-as: Is it just me or is this chunk dealing with helm-certgen still being done as root?

Yep:
```
vagrant@vagrant-kube:~> ls -la ~/.helm/plugins ~/.helm/plugins/certgen
/home/vagrant/.helm/plugins:
total 12
drwxr-xr-x 3 vagrant users 4096 Aug  2 20:24 .
drwxr-xr-x 6 vagrant users 4096 Aug  2 20:22 ..
drwxr-xr-x 2 root    root  4096 Aug  2 20:23 certgen

/home/vagrant/.helm/plugins/certgen:
total 16
drwxr-xr-x 2 root    root  4096 Aug  2 20:23 .
drwxr-xr-x 3 vagrant users 4096 Aug  2 20:24 ..
-rwxr-xr-x 1 root    root  1003 Aug  2 20:23 certgen.sh
-rw-r--r-- 1 root    root   189 Aug  2 20:23 plugin.yaml
```

bin/common/install_tools.sh
HartS: Hmm.. looks like we're "missing"(?) the wget/curl for kubectl also. I think this was generally because kubectl comes in through the suse packages.. since this script isn't generally installing kubernetes, I'm not convinced it should be installing kubectl (but we should probably remove `kubectl_url` if not).

If we do want it to install kubectl, it can do so optionally only if `kubectl` isn't already in PATH somewhere (which will prevent it from installing kubectl over the package-installed kubectl for deployments from our packer-built image)

bin/dev/install_tools.sh
HartS: I must have just forgotten to add the fix to this, good catch

bin/common/install_tools.sh
mook-as: Still missing kubectl for the chmod ;)

bin/dev/install_tools.sh
```diff
@@ -1,90 +1,64 @@
...
-
 echo "Installing helm-certgen ..."
-helm_certgen_dir="$(mktemp -d)"
-trap "rm -rf '${helm_certgen_dir}'" EXIT
-git clone --branch "${HELM_CERTGEN_VERSION}" --depth 1 https://github.com/SUSE/helm-certgen.git "${helm_certgen_dir}"
-docker run --rm -v "${bin_dir}":/out:rw -v "${helm_certgen_dir}:/go/src/github.com/SUSE/helm-certgen:ro" "golang:${GOLANG_VERSION}" /usr/bin/env GOBIN=/out go get github.com/SUSE/helm-certgen
-if [[ $(stat -c '%u' "${bin_dir}/helm-certgen") -eq 0 ]]; then
-  docker run --rm -v "${bin_dir}":/out:rw "golang:${GOLANG_VERSION}" /bin/chown "$(id -u):$(id -g)" /out/helm-certgen
+
+if [[ $(id -u) -eq 0 ]] && id -u vagrant &> /dev/null; then
```
mook-as: Please use `test` over `[[` (more portable).

PR #1022, opened 2017-07-11
Poll for smoke+cats instead of hardcoded sleep

alexjh:
This timed out for me a few times. Would be nice to get a more accurate timing as well.

### DIFF
--- a/make/cats
+++ b/make/cats
@@ -38,7 +38,20 @@ kubectl create --namespace="${NAMESPACE}" --filename="kube/bosh-task/acceptance-
 
 stampy "${GIT_ROOT}/scf_metrics.csv" "$0" make-cats::create end
 
-sleep 5 ; # Slap dash: Wait for container to be up and running
+i=0
+
+set +o errexit
+while [ "$(k get pod cf:^acceptance-tests -o=json | jq -r .status.phase)" != "Running" ]
+do
+  i=$((i + 1))
+  if [ ${i} -gt 30 ]
+  then
+    echo "acceptance-tests container failed to reach Running state"
+    exit 1
+  fi
+  sleep 1
+done
+set -o errexit
 
 stampy "${GIT_ROOT}/scf_metrics.csv" "$0" make-cats::log end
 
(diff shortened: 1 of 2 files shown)


make/cats
```diff
@@ -38,7 +38,20 @@ kubectl create --namespace="${NAMESPACE}" --filename="kube/bosh-task/acceptance-
 
 stampy "${GIT_ROOT}/scf_metrics.csv" "$0" make-cats::create end
 
-sleep 5 ; # Slap dash: Wait for container to be up and running
+i=0
+
+set +o errexit
+while [ "$(k get pod cf:^acceptance-tests -o=json | jq -r .status.phase)" != "Running" ]
```
andreas-kupries: If we get 🎩 running again this may mismatch on `acceptance-test-brain`.

andreas-kupries: And I am not sure right now how to prevent that, given that the :cat:'s name is a prefix.

mook-as: `k get pod cf:^acceptance-tests-\\d`?

andreas-kupries: Can we assume that the first is a digit ?
This is a bosh-task, and I see no pod number, just the trailing unique hex code.

andreas-kupries: Maybe just make note of the possible trouble for now.
I mean, :tophat: is currently not working anymore either, and might not be resurrected, making it moot.

alexjh: Ah, we can get it by excluding the second `-`

```bash
vagrant@vagrant-kube:~/scf> k get pod cf:^acceptance-tests
fetching pod => 6
ambiguous query for pod matches: [cf:acceptance-tests-brain-fhp7k cf:acceptance-tests-t8mwn]
vagrant@vagrant-kube:~/scf> k get pod cf:^acceptance-tests-\\w+$
fetching pod => 6
kubectl get pod acceptance-tests-t8mwn --namespace cf
NAME                     READY     STATUS                         RESTARTS   AGE
acceptance-tests-t8mwn   0/1       secrets "secret-1" not found   0          2m
```

andreas-kupries: Oh, I see. The `-\w+` does not match the `-` found in `brain-....`. Ok, that looks good.

andreas-kupries:
Looks good, works for me.

PR #1024, opened 2017-07-12
Fix cert gen for helm

alexjh:


docker-images/cert-generator/cert-generator.sh
```diff
@@ -14,13 +14,14 @@ trap finish EXIT
...
 
-/generate-certs.sh -e ${env_dir} /tmp/uaa-certs.env > /dev/null
-/generate-dev-certs.sh -e ${env_dir} "${NAMESPACE}" /tmp/scf-certs.env > /dev/null
+/generate-certs.sh -e "${env_dir}" /tmp/uaa-certs.env > /dev/null
+/generate-dev-certs.sh -e "${env_dir}" "${NAMESPACE}" /tmp/scf-certs.env > /dev/null
 
-perl -pe 's@(.+?)=(.+)@$1: "$2"@' < /tmp/uaa-certs.env > /out/uaa-cert-values.yaml
-perl -pe 's@(.+?)=(.+)@$1: "$2"@' < /tmp/scf-certs.env > /out/scf-cert-values.yaml
+echo "env:" > /out/uaa-cert-values.yaml
+echo "env:" > /out/scf-cert-values.yaml
```
mook-as: This is because we want a `env:` [in fissile], right? Not sure why we do that, but okay…

[in fissile]: https://github.com/SUSE/fissile/blob/c51c493c6becf6511f2d7e28f0a6c47d98f8450e/kube/pod.go#L213

PR #1027, opened 2017-07-12
Ensure that output directory is an absolute path.

andreas-kupries:
**WIP** marker because this is untested. Not because I don;'t want to, given the triviality of the change, but because docker tells me:
```
Unable to find image 'splatform/cert-generator:latest' locally
docker: Error response from daemon: repository splatform/cert-generator not found: does not exist or no pull access.
See 'docker run --help'.
```
IOW I do not have the image locally, and the image is not found on the hub either.


alexjh:
You can build it locally with `make cert-generator`

andreas-kupries:
Ah, good. Trying that now.

andreas-kupries:
That worked to give me a local image, and I was able to test. Worked for me, now in CR.

PR #1029, opened 2017-07-13
Add a mysql broker role to scf

viovanov:
This requires a change to the stemcell (needs `rsync`).
The broker won't compile without it.
Until there's a proper stemcell for it, one can use `splatform/fissile-stemcell-opensuse:42.2-vladi-devel` as a stemcell to test this.


viovanov:
Opened https://github.com/SUSE/bosh-linux-stemcell-builder/pull/3 for adding rsync to the stemcell

viovanov:
The broker is now working, but we still need to automatically create a security group so applications can access their databases.

PR #1030, opened 2017-07-14
Cleaning up *_PROXY (already exposed, removed some bogus refs)

andreas-kupries:
https://trello.com/c/3suRQ4P7/234-expose-httpproxy-httpsproxy
Per the research the *_PROXY variables are already exposed as user parameters.

Only thing found were references in the vagrant-setup common.rb code which treated them as special in the role-manifest, and use of the lower-case variants in the settings.env for vagrant.

Actions for this:
* Remove the lower-case variants of proxy variables. Not in the role-manifest.
* Not special anymore either, removed from the vagrant-setup.

Note, as this removes stuff merging might be a risk.
Tested with the vagrant box that the setup as committed (no proxying) is still building. ✔️ 
Further tested that it is standing up. ✔️


bin/settings/settings.env
```diff
@@ -76,10 +76,6 @@ UAA_CLIENTS_TCP_EMITTER_SECRET=uaa_client_tcp_emitter_secret
...
 USE_DIEGO_PRIVILEGED_CONTAINERS=false
 USE_STAGER_PRIVILEGED_CONTAINERS=false
-
-# HTTP_PROXY, HTTPS_PROXY, NO_PROXY are set via the transformer where applicable
 HTTP_PROXY=
-http_proxy=
 HTTPS_PROXY=
-https_proxy=
-NO_PROXY=.hcf,127.0.0.1
+NO_PROXY=.scf,127.0.0.1
```
mook-as: I don't think we need `.scf` anymore; rather, we need `.local` (because the cluster is `….cluster.local`).

andreas-kupries: Will change.
Will write descriptions.

Would like to ignore `vagrant box behind proxy`

andreas-kupries: Done with commit 5046892b4902e0b25e69759eb8bbed38baa2772f

PR #1034, opened 2017-07-17
Add `stop-on-failure` tag to tests

alexjh:
A new feature was added to fissile to check this tag. If a bosh-task has
it, it will stop the pod when the test script fails. Without this tag,
kubernetes will restart the pod to try to get the script to finish
successfully.

https://github.com/SUSE/fissile/pull/262


alexjh:
Bumped the fissile version, although we don't have a pipeline working right now, so it's not uploaded.

alexjh:
@mook-as updated fissile now that it's been merged to master

PR #1039, opened 2017-07-19
Fix testbrain

jandubois:


make/brain
```diff
@@ -0,0 +1,62 @@
...
+else
+    echo 1>&2 Namespace ${NAMESPACE} is missing. SCF is not running.
+    exit 1
+fi
+
+stampy "${GIT_ROOT}/scf_metrics.csv" "$0" make-brain::setup start
+
+. "${GIT_ROOT}/bin/settings/settings.env"
+
+cf api api.cf-dev.io --skip-ssl-validation
```
mook-as: `api.${DOMAIN}`?

make/brain
```diff
@@ -0,0 +1,62 @@
...
+cf auth admin "${CLUSTER_ADMIN_PASSWORD}"
+cf enable-feature-flag diego_docker
+
+stampy "${GIT_ROOT}/scf_metrics.csv" "$0" make-brain::setup end
+
+stampy "${GIT_ROOT}/scf_metrics.csv" "$0" make-brain::create start
+
+# Delete left-over pod/definition from previous runs, then create/run
+kubectl delete --namespace="${NAMESPACE}" --filename="kube/bosh-task/acceptance-tests-brain.yml" \
+    2> /dev/null || /bin/true
```
mook-as: Don't we need to wait for the delete to complete?

jandubois: Yes, we should. Let me know if you want me to fix this, as you already approved the PR!

mook-as: It's a race! Either you fix it first, or I merge it first. I'd prefer you fix it, but will merge without.

jandubois: > It's a race! Either you fix it first, or I merge it first.

Looks like I won. 🥇

The `make/smoke`, `make/brain`, and `make/cats` scripts share 90% of their code and should be refactored. But given that we want to replace them with helm charts anyways, I don't want to spend any more time on this.

make/brain
```diff
@@ -0,0 +1,68 @@
...
+
+printf "Waiting for namespace %s to be deleted...\n" "${NAMESPACE}"
+while has_namespace ; do
+    sleep 1
```
mook-as: This is wrong now; the _namespace_ isn't going away, only the pod is. This wait will never finish.

make/tests
```diff
@@ -0,0 +1,79 @@
...
+  fi
+  sleep 1
+done
+
+stampy "${METRICS}" "$0" "make-tests::${POD_NAME}::log" end
+
+# First show the logs accumulated so far, then stream further logs in
+# a way which terminates when the pod (= testsuite) terminates.
+kubectl logs   --namespace "${NAMESPACE}" "${POD_NAME}"
+kubectl attach --namespace "${NAMESPACE}" "${POD_NAME}"
```
mook-as: I just realized that this made no sense _before_ your change either: this should just be `kubectl logs --follow …`

make/tests
```diff
@@ -0,0 +1,79 @@
...
+
+# Delete left-over pod/definition from previous runs, then create/run
+kubectl delete --namespace="${NAMESPACE}" --filename="kube/bosh-task/${POD_NAME}.yml" \
+    2> /dev/null || /bin/true
+
+echo "Waiting for pod ${POD_NAME} to be deleted..."
+while has_pod ; do
+    sleep 1
+done
+
```
mook-as: Would be nice to print something here before running `kubectl create` so we can tell when the waiting finished.  Except you won't have a chance to, because I'll merge this PR before you even read this comment 😁

make/tests
andreas-kupries: That is not quite true. The `attach` will  terminate when the test pod terminates.
The `logs --follow` did not.
That said, with the change of `kind` this may have been fixed.

PR #1041, opened 2017-07-19
Unpin node version in test app

jandubois:
That way we don't have to modify the tests whenever the assets bundled
with the nodejs-buildpack change.

### DIFF
--- a/src/hcf-release/src/acceptance-tests-brain/test-resources/node-env/package.json
+++ b/src/hcf-release/src/acceptance-tests-brain/test-resources/node-env/package.json
@@ -1,8 +1,5 @@
 {
   "name": "node-env",
   "version": "1.0.0",
-  "engines": {
-    "node": "5.x"
-  },
   "dependencies": {}
 }


src/hcf-release/src/acceptance-tests-brain/test-resources/node-env/package.json
```diff
@@ -2,7 +2,6 @@
   "name": "node-env",
   "version": "1.0.0",
   "engines": {
-    "node": "5.x"
   },
```
mook-as: We should either drop `engines` completely, or say `"engines": { "node": ">=5.0" }`, right?

jandubois: I don't really care either way; I've now removed `engines` completely and force-pushed.

Or do you want to remove `dependencies` as well?

```
  "dependencies": {}
```

mook-as:
We end up running on CF with node v4.8.1; is that okay? It's less than the listed minimum version.

jandubois:
> We end up running on CF with node v4.8.1; is that okay? It's less than the listed minimum version.

I think so. At least it passed all tests, which is good enough for me.

mook-as:
Sorry for the delay, had issues getting an image (turns out I actually forgot to re-build the kube yaml).

PR #1042, opened 2017-07-21
Make `make/kube` less hacky

mook-as:
We previously had `make/kube` make some assumptions, including `${FISSILE_OUTPUT_DIR}`.  That's totally confusing (why is that set to the helm directory‽) and could lead to trouble down the road because `--output-dir` is such a sensible argument name.

This fixes it to set `FISSILE_OUTPUT_DIR` to `scf.git/output/`, and have `make/kube` append the `kube`/`helm` suffix itself.

This does not currently attempt to solve the mess that is `FISSILE_DEFAULTS_FILE`.

This was needed to try to make the concourse pipeline a bit saner.

### DIFF
--- a/.envrc
+++ b/.envrc
@@ -95,5 +95,5 @@ export FISSILE_STEMCELL=${FISSILE_STEMCELL:-splatform/fissile-stemcell-opensuse:
 # The defaults are all for helm.
 # The make/kube script performs overrides for kube output.
 export FISSILE_USE_MEMORY_LIMITS=false
-export FISSILE_OUTPUT_DIR="${FISSILE_OUTPUT_DIR:-${PWD}/helm}"
+export FISSILE_OUTPUT_DIR="${FISSILE_OUTPUT_DIR:-${PWD}/output}"
 export FISSILE_DEFAULTS_FILE="bin/settings/certs.env,bin/settings/kube/ca.env"
--- a/.gitignore
+++ b/.gitignore
@@ -4,12 +4,8 @@ _work/
 target/
 password
 *.pem
-*.tfstate*
-*.tfvars
-.terraform/
-!.fissile/.gitkeep
-!.fissile/.bosh/.gitkeep
-.fissile/*
+!.gitkeep
+/.fissile/
 .vagrant
 .bash_history
 *.box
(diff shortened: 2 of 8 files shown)


jandubois:
If you change the output dir for `make helm`, then I would prefer to call it `output/cf/` because the directory name should match the name of the chart.

It is a small thing, but any errors from helm right now refer to `cf/templates/xxx.yml` and not `helm/templates/xxx.yml`, which might be confusing if you don't know about this quirk.

mook-as:
The output of `make helm-dist` will still call it `cf/…` though; `output/cf` is, on its own, confusing when exploring the directory.

jandubois:
Ok, maybe make it `output/helm/cf`, given that we want to create `output/helm/smoke` etc in the future?

viovanov:
This feels bad because `fissile` is not making it easy for us.
Since `FISSILE_OUTPUT_DIR` is the same for multiple commands, the commands should namespace their output. `fissile build helm` should create a `helm` dir, and the same for `fissile build kube`.
@mook-as we should be able to fix `FISSILE_DEFAULTS_FILE` by reorganizing the env files.
I don't think there's a reason to have more than 2 files anymore (certs and non certs), but I might be forgetting something.

mook-as:
Yeah, the fact that fissile is weird is annoying, but I think _not_ creating a toplevel subdirectory is the correct option here (since, as noted here, we want `…/helm/cf/` and fissile shouldn't know anything about the last part).

I'm punting the `FISSILE_DEFAULTS_FILE` stuff to a future PR, because this change is complicated enough.

viovanov:
+1 on env cleanup in another ticket
I don't think it's the correct option, mainly because what we want is just a legacy dir structure afaict. (`uaa` and `cf` being merged in some fashion).
The following seems like a reasonable output to me - the make scripts could then merge those into something else if we wanted to.
```
output/
  scf/
    helm/
    kube/
  uaa/
    helm/
    kube/
```

mook-as:
I don't think `uaa` should be in `output/` because that lives in `src/uaa-fissile-release/…`. (It would of course end up in the final distributed zip file, but this isn't that.)  For this one, I'm only concerned about the _scf_ kube/helm outputs.  Hence `output/kube/{bosh,bosh-task,secrets}/` and `output/helm/{Charts.yaml,…}` were decent enough to me.

The final dist output, though, should have something like your described output, except that `kube` and `helm` would likely end up as separate files or something, but that's for another PR (yet to be created, because it depends on this one).

mook-as:
This is _really_ stale now. We can redo this when we actually want it.

PR #1046, opened 2017-07-22
Generate separate kube and helm artifacts

mook-as:
This makes `make kube-dist` generate both kube and helm bits

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,9 +4,6 @@ _work/
 target/
 password
 *.pem
-*.tfstate*
-*.tfvars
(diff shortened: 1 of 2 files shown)


mook-as:
> * Is there a reason we keep `uaa/kube-test/exposed-ports.yml`? I thought this would be redundant with the latest service changes in fissile.

Oversight: it wasn't removed.
> * Why is `make kube-dist` deleting the `kube/` directory? It is cheap to recreate, but still, feels odd to me.

No idea.

PR #1054, opened 2017-07-25
Remove reference to `make helm`

alexjh:
helm charts are generated in the `images` step, called by `vagrant-prep`


alexjh:
> Didn't realize it made helm bits automatically. Less typing for me next time :D

It only started doing that last week :)

PR #1061, opened 2017-07-27
Increase size of data volume for garden

viovanov:
Also mount /var/vcap/data, so there's enough room for grootfs data, and it doesn't get stored on the container's volume.


mook-as:
So what's the difference between garden-data and garden-graph?

PR #1062, opened 2017-07-27
Remove cc-service-dashboards UAA client

jandubois:
because we don't deploy the dashboard, but the client may end up being
not unique:

```
Creating service broker mysql-dev as admin...
FAILED
Server error, status code: 502, error code: 270012, message: Service broker catalog is invalid:
Service mysql-dev
  Service dashboard client id must be unique

Service broker already exists - updating broker
```


jandubois:
Marked as WIP because I haven't tested it yet.

PR #1065, opened 2017-07-27
Update fissile version to always use .yaml for kube/helm config

jandubois:


jandubois:
https://github.com/SUSE/scf/pull/1071 already includes the changes from this PR, so it is now obsolete and can be closed.

The changes requested by the comment above are in https://github.com/SUSE/scf/pull/1075

PR #1067, opened 2017-07-28
Added test for tiller to the readiness checker.

andreas-kupries:
Ref: https://trello.com/c/cCgf3Rsr/271-1-ci-kube-ready-state-checksh-must-include-tiller-install-check

Tweaked the messages a bit.

Testing:
- On a (newly-made) vagrant box the check should succeed without any special work.
- For testing the other side `helm reset --force` uninstalls tiller. After that the check should fail.
- And redoing `helm init` should make it succeed again (after a while, tiller takes a bit to be fully running).

### DIFF
--- a/bin/dev/kube-ready-state-check.sh
+++ b/bin/dev/kube-ready-state-check.sh
@@ -87,7 +87,13 @@ fi
 # kube-dns shows 4/4 ready
 if having_category kube ; then
     kubectl get pods --namespace=kube-system --selector k8s-app=kube-dns | grep -Eq '([0-9])/\1 *Running'
-    status "kube-dns should shows 4/4 ready"
+    status "kube-dns should be running (show 4/4 ready)"
+fi
+
+# tiller-deploy shows 4/4 ready
+if having_category kube ; then
+    kubectl get pods --namespace=kube-system --selector name=tiller | grep -Eq '([0-9])/\1 *Running'
+    status "tiller should be running (1/1 ready)"
 fi
 
 # ntp is installed and running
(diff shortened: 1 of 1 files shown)


bin/dev/kube-ready-state-check.sh
```diff
@@ -87,7 +87,13 @@ fi
 # kube-dns shows 4/4 ready
 if having_category kube ; then
     kubectl get pods --namespace=kube-system --selector k8s-app=kube-dns | grep -Eq '([0-9])/\1 *Running'
-    status "kube-dns should shows 4/4 ready"
+    status "kube-dns should be running (show 4/4 ready)"
+fi
+
+# tiller-deploy shows 4/4 ready
+if having_category kube ; then
+    kubectl get pods --namespace=kube-system --selector name=tiller | grep -Eq '([0-9])/\1 *Running'
```
jandubois: You may want to add `2> /dev/null` to the `kubectl` command, otherwise you get some noise when the test fails:

```
No resources found.
```

andreas-kupries: Will do. I guess the kube-dns check coming before should get the same, I copy-pasta that one.

andreas-kupries: Done with commit 42c06fc3cf56e36b6c1fbe7c6e032f8e59afca1e

jandubois:
@mook-as We can't assume that `helm` is installed on the kube nodes. In general it shouldn't; it is a client-side tool, and only tiller should be on the server.

mook-as:
No, but it should exist on the `kube` (as opposed to `node` / `api`) category things, right? Same place `kubectl` exists? Otherwise the admin couldn't deploy anything…

andreas-kupries:
That sounds like a separate check (tiller on server - done, helm on client side?)
Should we have a check for `kubectl` ?

jandubois:
@mook-as Why would an admin deploy applications from inside the master node? That should be reserved for configuring kube itself.

PR #1069, opened 2017-07-28
Bump helm version to 2.5.1

HartS:
v2.4.2 has some issues that can result in helm communication with etcd
timing out during `make run`. Using helm v2.5.1 instead seems to avoid
this issue.


mook-as:
LGTM, but conflicts.

HartS:
This update already snuck in elsewhere; closing

PR #1070, opened 2017-07-28
Don't depend on systemctl for kube checks

alexjh:


bin/dev/kube-ready-state-check.sh
```diff
@@ -87,38 +91,40 @@ fi
...
+    kubelet=$(pgrep -ax hyperkube | grep " kubelet " )
     [[ $kubelet == *"--allow-privileged"* ]]
     status "Privileged must be enabled in 'kubelet'"
 fi
 
 # override tasks infinity in systemd configuration
 if having_category node ; then
-    systemctl cat containerd | grep -wq "TasksMax=infinity"
-    status "TasksMax must be set to infinity"
+    if has_command isystemctl ; then
```
jandubois: `isystemctl` should be `systemctl`?

alexjh: Yes, definitely a typo. Fixed in 24282002, I had to rebase and force push because of a conflict with the new tiller checks.

PR #1073, opened 2017-08-02
Replaced `cd/pwd` construct to get absolute path with `readlink -f`.

andreas-kupries:
https://trello.com/c/FTwkoGmF/297-unset-cdpath-in-certgen-script

The construct broke due to ev `CDPATH` causing the `cd` to generate additional output.
Instead of removing `CDPATH` the `readlink` gives us the same functionality without hassle.

Eight more places.
(Note: Currently unable to test myself, the local concourse blocks me from playing with vagrant).

### DIFF
--- a/bin/dev/vagrant-envrc
+++ b/bin/dev/vagrant-envrc
@@ -1,4 +1,10 @@
-ROOT=`readlink -f "$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )/../"`
+
+# prevent cd from printing the directory it changes to. This breaks
+# cd/pwd constructions (See **).
+unset CDPATH
+
+# (**)
+ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")/../" && pwd)"
 
 # Directory used when mounting volumes to containers
 export HCF_RUN_STORE="$HOME/.run/store"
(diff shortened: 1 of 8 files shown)


andreas-kupries:
:( Sigh

andreas-kupries:
Redone using `unset CDPATH`.
Removed two pre-existing instances of `readlink -f` with suitable rewrites.
Cleaned the history (rebase).

bin/dev/vagrant-envrc
```diff
@@ -1,4 +1,10 @@
-ROOT=`readlink -f "$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )/../"`
+
+# prevent cd from printing the directory it changes to. This breaks
+# cd/pwd constructions (See **).
+unset CDPATH
+
+# (**)
+ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")/../" && pwd)"
```
alexjh: $ROOT isn't needed any more, it was previously needed for `HCF_PACKAGE_COMPILATION_CACHE` in https://github.com/SUSE/scf/blob/46fbabad86c1a50c281f76462ef0f06e424a4536/bin/.runrc

runrc was renamed to vagrant-envrc

PR #1078, opened 2017-08-04
Cleanup vagrant envrc

alexjh:
Rather than waiting for Andreas' to come back, I've merged his branch, and added some cleanup here.


alexjh:
Good point, removed in a103250

PR #1079, opened 2017-08-04
Bump smoke & cats submodules to cf-release v265 level

jandubois:
When cf-release & friends were bumped to v265, the smoke and cat suites have not been updated.

For the new CATs to pass we need to first update the buildpacks to their corresponding v265 level as well.


alexjh:
I've integrated and tested a rebased version of this in my buildpacks branch, and included it as part of my PR.

PR #1080, opened 2017-08-04
Add Jenkinsfile for experimental Jenkins-based builds

mook-as:
Jenkins should be configured to use a multi-branch pipeline. Automatically triggering builds on commit is optional.

This uses the new Jenkins pipeline format (because it's actually advanced enough to do so now), and only goes up to publishing everything (no deploys).

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,0 +1,203 @@
+#!/usr/bin/env groovy
+// vim: set et sw=4 ts=4 :
+pipeline {
+    agent any
+    options {
+        disableConcurrentBuilds() // Otherwise clean would delete the images
+        skipDefaultCheckout() // We do our own checkout so it can be disabled
+        timestamps()
+        timeout(time: 5, unit: 'HOURS')
+    }
+    parameters {
+        booleanParam(
+            name: 'SKIP_CHECKOUT',
+            defaultValue: false,
+            description: 'Skip the checkout step for faster iteration',
+        )
+        booleanParam(
+            name: 'RESET',
+            defaultValue: false,
+            description: 'Remove all existing build artifacts',
+        )
+        booleanParam(
+            name: 'CLEAN',
+            defaultValue: true,
+            description: 'Remove build artifacts that should normally not be reused',
+        )
+        booleanParam(
+            name: 'PUBLISH',
+            defaultValue: true,
+            description: 'Enable publishing',
+        )
+        credentials(
+            name: 'S3_CREDENTIALS',
+            description: 'AWS access key / secret key used for publishing',
+            defaultValue: 'cred-s3-scf',
+        )
+        string(
+            name: 'S3_REGION',
+            description: 'AWS S3 region the target bucket is in',
+            defaultValue: 'us-east-1',
+        )
+        string(
+            name: 'S3_BUCKET',
+            description: 'AWS S3 bucket to publish to',
+            defaultValue: 'scf',
+        )
+        string(
+            name: 'S3_PREFIX',
+            description: 'AWS S3 prefix to publish to',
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -0,0 +1,203 @@
...
+                        usernameVariable: 'DOCKER_HUB_USERNAME',
+                        passwordVariable: 'DOCKER_HUB_PASSWORD',
+                    )]) {
+                        sh 'docker login -u "${DOCKER_HUB_USERNAME}" -p "${DOCKER_HUB_PASSWORD}" '
+                    }
+                    sh '''
+                        set -e +x
+                        source ${PWD}/.envrc
+                        set -x
+                        unset HCF_PACKAGE_COMPILATION_CACHE
```
aarondl: Continuously being suspicious.

Jenkinsfile
```diff
@@ -0,0 +1,203 @@
...
+        )
+        string(
+            name: 'S3_PREFIX',
+            description: 'AWS S3 prefix to publish to',
+            defaultValue: 'scf/config/',
+        )
+        credentials(
+            name: 'DOCKER_CREDENTIALS',
+            description: 'Docker credentials used for publishing',
+            defaultValue: 'cred-docker-scf',
```
mook-as: This is the id of the credential. It has a default value, otherwise when Jenkins makes a build (on a new commit / PR / whatever), it tries to start and falls over due to not having a value.  Yes, it's that dumb.

Jenkinsfile
```diff
@@ -0,0 +1,203 @@
...
+    }
+
+    stages {
+        stage('reset') {
+            when {
+                expression { return params.RESET }
+            }
+            steps {
+                checkout(
+                    $class: scm.$class,
```
mook-as: Sort of; the `scm` object is basically "what did I use to check out the branch that holds the Jenkinsfile", so I'm just copying all the fields (including the oddly-named `$class` field) over.  Those fields really only make sense for git.

Jenkinsfile
```diff
@@ -0,0 +1,203 @@
...
+            }
+        }
+        stage('tools') {
+            steps {
+                ansiColor('xterm') {
+                    sh '''
+                        set -e +x
+                        source ${PWD}/.envrc
+                        set -x
+                        unset HCF_PACKAGE_COMPILATION_CACHE
```
mook-as: Yes, [yes it is](https://github.com/SUSE/scf/blob/6edb63657a08ddbb1d6a355a90f96b0ddd4fc4f1/make/compile#L54).

Jenkinsfile
```diff
@@ -0,0 +1,203 @@
...
+                        make ${FISSILE_BINARY}
```
mook-as: Because that's what is listed in the Makefile as the target.  I don't actually remember what it is, but using the variable works :D  The target just runs `install_tools.sh`.  (The _right_ `install_tools.sh`, after Hart's changes for vagrant box v2.0.6)

Jenkinsfile
```diff
@@ -0,0 +1,203 @@
...
+        }
+        stage('build') {
+            steps {
+                ansiColor('xterm') {
+                    sh '''
+                        set -e +x
+                        source ${PWD}/.envrc
+                        set -x
+                        unset HCF_PACKAGE_COMPILATION_CACHE
+                        make vagrant-prep validate
```
mook-as: No, because `validate` requires `bosh create release`.  Of course, I'm pretty sure it's actually run as part of `make vagrant-prep` already…

Jenkinsfile
mook-as: I just _really_ don't want to copy the compile cache anywhere (I'm not in a VM!)

PR #1084, opened 2017-08-15
Packer: support building Jenkins slaves on Triton

mook-as:
Triton uses a different base image, so we need to add the vagrant user and install helm separately. We also need Java for Jenkins.  Since it requires a secondary disk for bulk storage, we need a bunch of systemd units to line things up correctly to ensure we mount the hostpath stuff on that disk.

### DIFF
--- a/packer/scripts/create-kube-certs.sh
+++ b/packer/scripts/create-kube-certs.sh
@@ -2,6 +2,8 @@
 
 set -o errexit -o xtrace
 
+export PATH="${PATH}:/usr/local/bin/"
+
 mkdir -p /run/certstrap
 certstrap --depot-path "/run/certstrap" init --common-name "CA.kube.vagrant" --passphrase "" --years 10
 certstrap --depot-path "/run/certstrap" request-cert --common-name "apiserver" --passphrase "" --ip 127.0.0.1,192.168.77.77,172.17.0.1,10.254.0.1 --domain kubernetes.default.svc,kubernetes.default,kubernetes,localhost
--- a/packer/vagrant-box.json
+++ b/packer/vagrant-box.json
@@ -1,6 +1,6 @@
 {
     "variables": {
-        "version": "2.0.6+{{isotime \"20060102-1504\"}}",
+        "version": "2.0.6",
         "vm_name": "scf-vagrant",
         "ssh_username": "vagrant",
         "ssh_password": "vagrant",
@@ -21,7 +21,7 @@
         {
             "type": "vmware-iso",
             "headless": true,
-            "vm_name": "{{user `vm_name`}}",
+            "vm_name": "{{user `vm_name`}}+{{isotime \"20060102-1504\"}}",
             "guest_os_type": "suse-64",
             "disk_size": 120480,
             "ssh_username": "{{user `ssh_username`}}",
@@ -47,7 +47,7 @@
         {
             "type": "virtualbox-iso",
             "headless": true,
-            "vm_name": "{{user `vm_name`}}",
+            "vm_name": "{{user `vm_name`}}+{{isotime \"20060102-1504\"}}",
             "guest_os_type": "OpenSUSE_64",
             "guest_additions_mode": "disable",
             "disk_size": 120480,
@@ -76,7 +76,7 @@
         {
             "type": "qemu",
             "headless": true,
-            "vm_name": "{{user `vm_name`}}",
+            "vm_name": "{{user `vm_name`}}+{{isotime \"20060102-1504\"}}",
             "disk_size": 120480,
             "disk_interface": "virtio",
             "disk_compression": true,
@@ -103,7 +103,7 @@
         {
             "type": "triton",
             "image_name": "{{ user `vm_name` }}",
-            "image_version": "{{ user `version` }}",
+            "image_version": "{{ user `version` }}+{{isotime \"20060102-1504\"}}",
             "source_machine_image": "{{ user `source_machine_image` }}",
             "source_machine_package": "{{ user `source_machine_package` }}",
             "ssh_private_key_file": "{{ user `ssh_private_key_file` }}",
--- a/packer/vagrant-box.json
+++ b/packer/vagrant-box.json
@@ -1,7 +1,7 @@
 {
     "variables": {
-        "version": "2.0.6",
-        "vm_name": "scf-vagrant-{{isotime \"20060102-1504\"}}",
+        "version": "2.0.6+{{isotime \"20060102-1504\"}}",
+        "vm_name": "scf-vagrant",
         "ssh_username": "vagrant",
         "ssh_password": "vagrant",
         "iso_url": "http://download.opensuse.org/distribution/leap/42.2/iso/openSUSE-Leap-42.2-NET-x86_64.iso",
@@ -9,6 +9,12 @@
         "iso_checksum_type": "sha256",
         "http_directory": "http",
         "ssh_wait_timeout": "300m",
+        "source_machine_package": "{{ env `SOURCE_MACHINE_PACKAGE` }}",
+        "source_machine_image": "{{ env `SOURCE_MACHINE_IMAGE` }}",
+        "ssh_private_key_file": "{{ env `SSH_PRIVATE_KEY_FILE` }}",
+        "triton_account": "{{ env `TRITON_ACCOUNT` }}",
+        "triton_key_id": "{{ env `TRITON_KEY_ID` }}",
+        "triton_url": "{{ env `TRITON_URL` }}",
         "shutdown_command": "echo vagrant | sudo -S shutdown -P now"
     },
     "builders": [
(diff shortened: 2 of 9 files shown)


packer/scripts/create-kube-certs.sh
```diff
@@ -2,6 +2,8 @@
 
 set -o errexit -o xtrace
 
+export PATH="${PATH}:/usr/local/bin/"
```
HartS: This should be unnecessary, since packer runs the provisioning shell script `scripts/usr-local-bin-to-path.sh` first

packer/vagrant-box.json
```diff
@@ -1,14 +1,20 @@
 {
     "variables": {
-        "version": "2.0.6",
-        "vm_name": "scf-vagrant-{{isotime \"20060102-1504\"}}",
+        "version": "2.0.6+{{isotime \"20060102-1504\"}}",
```
HartS: If we change the version format, wouldn't the Vagrantfile need to change also? Right now we can just reference the images which have already been built, sure, but when the box moves to a new version, we'd need to put the generated date component in the box URL manually. I'd prefer to add the date component directly into `image_version` in the triton builder configuration.

packer/scripts/create-kube-certs.sh
mook-as: Wrong packer; the Jenkins master doesn't get that, so `/usr/local/bin/` isn't in `/etc/login.defs`.

packer/vagrant-box.json
mook-as: Oh, I think you're right, because the output name is `"output": "scf-{{.Provider}}-v{{user `version`}}.box"`. Let me think about that for a bit.

mook-as: Fixed in 5548f499 (I think)

HartS:
^Ignore that note about referencing it from a different PR above, that was me intending to reference an entirely different PR.. Github doesn't remove that event even after the edit 😦

packer/scripts/create-kube-certs.sh
HartS: Why don't you run the above script so we can expect `/usr/local/bin` in PATH then?

mook-as: Mainly, because it's a global solution to a local problem, and making that change feels _wrong_.

HartS:
Most of [this section](https://github.com/SUSE/scf/blob/c81254fa483ff3fcd53a55ce8e931cae787d1e8b/packer/http/autoyast.xml#L193-L247) is redundant now, since packer is (or should be... we'll need to test) inserting these files into the right place. Since the Triton Jenkins slave build can't use the autoyast, a lot of that same work has been moved into the packer config file.. To further align the work for the different build targets, it would be preferable to move as much provisioning work as possible (or practical) from the autoyast into the packer config.

packer/scripts/create-kube-certs.sh
HartS: The Jenkins slave *should* get this, when built with the same packer config file as the vagrant image targets, as the provisioner which adds `/usr/local/bin` to PATH runs on all targets

packer/vagrant-box.json
```diff
@@ -151,7 +236,25 @@
...
+                "scripts/create-kubedns.sh"
+            ],
+            "start_retry_timeout": "7m"
+        },
+        {
+            "type": "shell",
+            "only": ["triton"],
+            "inline": [
+                "# Install helm",
+                "wget -O - https://kubernetes-helm.storage.googleapis.com/helm-v2.5.1-linux-amd64.tar.gz | tar xz -C /usr/local/bin --no-same-owner --strip-components=1 linux-amd64/helm",
```
HartS: Since we're already breaking parity with the other images built, I wonder if it would make sense to have two triton image targets... one for CI and one for development.. the CI one could use `bin/common/install_tools.sh` to install helm, k, and cf. The dev target could additionally provision the image with `bin/dev/install_tools.sh`

packer/vagrant-box.json
```diff
@@ -151,7 +236,25 @@
...
+                "sudo -i -u vagrant helm init",
+                "# Install additional packages",
+                "zypper --non-interactive install jre-openjdk-headless make zip"
```
HartS: What is the jre-openjdk-headless dep needed for? Also, `make` would be needed for dev deployments, but would CI deployments need it?

packer/vagrant-box.json
mook-as: Unfortunately, I don't think that's possible: the target name is the builder type.

packer/vagrant-box.json
mook-as: It's needed for Jenkins to run the slave bits. (After all, Jenkins is a giant Java app.)

Yes, CI needs `make` because it builds the images (`make vagrant-prep`, in fact).

packer/scripts/create-kube-certs.sh
mook-as: Hmm. I'll look into that some more. I am now suspecting the difference is 1) triton doesn't initially have a `vagrant` user, so I have to make packer login as root, combined with 2) my local SSH config default-enabling `ControlPath`

HartS: From my reading, the `vagrant` user is getting added before any of the other provisioners run, so this may just be due to `ControlPath`. If there's a reason we need this explicitly added that's eluding me, a comment around the export would be helpful.

packer/vagrant-box.json
HartS: My preference would be to create a `triton-dev.json` config file then, which takes the image produced from the regular CI triton build, and adds in the dev tools.

Alternatively, we could just throw all the tools into the image, or require people using the same image for development to install them themselves.

I dislike all of these options though :(

packer/vagrant-box.json
HartS: hmm... have you tested this? I wouldn't expect this to work without dev tools. I thought the idea was that for CI, the image we would use wouldn't need to `make vagrant-prep`, just take the pre-generated helm charts and deploy them.

packer/scripts/create-kube-certs.sh
mook-as: No, because there's only one spot to tell packer what user to use. Since packer creates the `vagrant` user, it has to be run as some other user (in fact, we only have `root` available).  So there's no `su` / relogin involved, hence the problems.

Hmm. I can probably get away with just spawning a new `type: shell`; trying that…

packer/vagrant-box.json
mook-as: Yeah, I don't really like any of the options either.  Hence taking the easy option for now until we get better ideas…

packer/vagrant-box.json
mook-as: Yes; I've got successful builds (heck, even 🚬 ) from it.  The Jenkins slaves here are doing double duty, both to generate the helm charts / images, as well as grabbing them (possibly on a different slave) to test them.  We don't really need dev tools, because we build everything in docker images (which themselves have the dev tools installed).

HartS: if the `vagrant-prep` is happening inside containers.. why do we need make?

mook-as: No, the `vagrant-prep` happens directly on the VM without any containers. The actual package building happens inside containers (spawned via `make` → `fissile` → `docker`), hence not needing dev tools.

packer/scripts/create-kube-certs.sh
HartS: I still don't understand how this has been working on the other builds, but since it looks like this is inexplicably needed for Triton, I'm not opposed to leaving the PATH change

Small nit: the trailing `/` is unconventional, but it doesn't cause problems

packer/vagrant-box.json
HartS: how are `make vagrant-prep` and `make run` working without direnv?

mook-as: [Sourcing .envrc directly](https://github.com/SUSE/scf/blob/9059d23cc0e59bcf5474181fdbe974acfd6d0ea9/Jenkinsfile#L140)

packer/scripts/create-kube-certs.sh
HartS: Ignore the above nit; as far as I can tell, there are no dangers of including a trailing `/` in `PATH`s... I know trailing `/` is typically preferred when referencing directories, so assumed there some weird compatibility reason for the convention of *not* including it in `PATH`.. as far as I can tell, there is none

packer/vagrant-box.json
HartS: We'll probably want to track the need for the triton dev image packer config, but this will work for CI

packer/scripts/create-kube-certs.sh
mook-as: Okay, looking at this more, I see that `/etc/sudoers` has:
> `Defaults secure_path="/usr/sbin:/usr/bin:/sbin:/bin"`
(Note lack of `/usr/local/bin`, even though `/etc/login.defs` is fixed)

Trying to figure out why…

mook-as: Oh, because later in the file:
```
## Uncomment to use a hard-coded PATH instead of the user's to find commands
# Defaults secure_path="/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin"
##

Yep, that's commented out.
```

mook-as:
Rebased per request

PR #1088, opened 2017-08-17
Add the Universal Service Broker to SCF

viovanov:


container-host-files/etc/hcf/config/opinions.yml
```diff
@@ -148,6 +148,15 @@ properties:
       tag: cc
     db_scheme: mysql
     port: 3306
+  cf-usb:
+    broker:
+      username: broker-admin
+      port: 24054
+    management:
+      dev_mode: true
```
mook-as: When would we want not-dev-mode? (Presumably we don't want to ship it to customers with dev mode turned on?)

container-host-files/etc/hcf/config/role-manifest.yml
```diff
@@ -1765,6 +1829,7 @@ configuration:
       value_type: certificate
     description: PEM-encoded CA certificate used to sign TLS certificate hosted by Diego BBS.
     required: true
+  - name: JWT_SIGNING_PUB
```
mook-as: We should at least have a description (though I think we want more than that)

container-host-files/etc/hcf/config/opinions.yml
viovanov: I don't know off the top of my head.

container-host-files/etc/hcf/config/role-manifest.yml
viovanov: What else other than the description are you looking for?
This is a value auto-retrieved by a script

mook-as: Ah; that wasn't clear (because it wasn't `internal`… or was that `type: environment`? I can never keep those two straight in my head)

PR #1093, opened 2017-08-18
Make make run run when IP is not 192.168.77.77

HartS:
These changes primarily do the following

- Clean up case logic around helm vs. kube in make/kube
- Use domain resolution to set external_ip (when deploying helm chart)
- Substitute DOMAIN and UAA_HOST in dotenv file & clean up after deploy

There's still an issue of fissile hardcoding 192.168.77.77 for `make helm` and `make kube`, but the helm deployment now overrides this with a more appropriate IP.

The work from this PR enables easy deployment of SCF to a vagrant box with an IP obtained via bridged networking (now possible by setting `VAGRANT_KVM_BRIDGE` or `VAGRANT_VBOX_BRIDGE` before `vagrant up`)  or an IP within the virtual network's IP range obtained by DHCP instead (by setting `VAGRANT_DHCP`). After a vagrant box is provisioned with one of these variables set, the user can create A and CNAME records for the desired domain, and then `export DOMAIN=${DOMAIN}` before running `make vagrant-prep` and `make run` in the VM


Makefile
```diff
@@ -25,8 +25,10 @@ ${FISSILE_BINARY}: bin/dev/install_tools.sh
 ########## VAGRANT VM TARGETS ##########
 
 certs: uaa-certs
+	[ -n ${DOMAIN} ] && sed -i "s/^DOMAIN=.*/DOMAIN=${DOMAIN}/g" ${GIT_ROOT}/bin/settings/network.env
```
andreas-kupries: ""-quote DOMAIN.
Add `-e` as script-indicator
Move `-i` to before path.

make/kube
```diff
@@ -6,24 +6,24 @@ GIT_ROOT=${GIT_ROOT:-$(git rev-parse --show-toplevel)}
...
     BUILD_TARGET=helm
-fi
-
-if [ "${BUILD_TARGET}" = "kube" ]; then
+else
+    BUILD_TARGET=kube
     # Overrides when generating kube config files instead of helm charts.
     FISSILE_OUTPUT_DIR="${PWD}/kube"
     ENV_FILES="$(echo bin/settings/*.env bin/settings/kube/*.env)"
+    DOMAIN=${DOMAIN:-$(grep '^DOMAIN=' bin/settings/network.env | sed 's/DOMAIN=//')}
```
andreas-kupries: `${GIT_ROOT}` missing.

make/kube
```diff
@@ -6,24 +6,24 @@ GIT_ROOT=${GIT_ROOT:-$(git rev-parse --show-toplevel)}
...
+    sed -e "s@^DOMAIN=.*@DOMAIN=$DOMAIN@" \
+        -e "s@^UAA_HOST=.*@UAA_HOST=uaa.$DOMAIN@" \
+        -i bin/settings/network.env
```
andreas-kupries: `${GIT_ROOT}` missing.

make/kube
```diff
@@ -6,24 +6,24 @@ GIT_ROOT=${GIT_ROOT:-$(git rev-parse --show-toplevel)}
...
     FISSILE_DEFAULTS_FILE="$(echo "${ENV_FILES}" | tr ' ' ,)"
 fi
 
 rm -rf "${FISSILE_OUTPUT_DIR}"
 
 fissile build "${BUILD_TARGET}"
+git checkout bin/settings/network.env
```
andreas-kupries: `${GIT_ROOT}` missing.

make/run
```diff
@@ -33,15 +33,18 @@ kubectl get storageclass persistent 2>/dev/null || {
 }
 
 source bin/settings/settings.env
-source bin/settings/network.env
+source <(sed -e '/^DOMAIN=/d' -e '/^UAA_HOST=/d' bin/settings/network.env)
+DOMAIN="${DOMAIN:-$(grep '^DOMAIN=' bin/settings/network.env | sed 's/DOMAIN=//')}"
```
andreas-kupries: `${GIT_ROOT}` missing in this and previous line.

make/run
```diff
@@ -33,15 +33,18 @@ kubectl get storageclass persistent 2>/dev/null || {
...
 
 helm install helm \
      --namespace ${NAMESPACE} \
      --set "env.CLUSTER_ADMIN_PASSWORD=$CLUSTER_ADMIN_PASSWORD" \
      --set "env.DOMAIN=${DOMAIN}" \
      --set "env.UAA_ADMIN_CLIENT_SECRET=${UAA_ADMIN_CLIENT_SECRET}" \
      --set "env.UAA_HOST=${UAA_HOST}" \
-     --set "env.UAA_PORT=${UAA_PORT}"
+     --set "env.UAA_PORT=${UAA_PORT}" \
+     --set "kube.external_ip=$(getent hosts ${DOMAIN} | awk '{ print $1; exit }')"
```
andreas-kupries: Same portability question on `getent` as in the UAA PR.

make/kube
jandubois: @andreas-kupries Where is it missing?

Makefile
```diff
@@ -25,8 +25,10 @@ ${FISSILE_BINARY}: bin/dev/install_tools.sh
...
 	${GIT_ROOT}/bin/generate-dev-certs.sh cf bin/settings/certs.env
 	${GIT_ROOT}/bin/settings/kube/ca.sh
+	git checkout ${GIT_ROOT}/bin/settings/network.env
```
jandubois: I object to throwing away the manual changes I have made to network.env

If you want to modify a file and reset it afterwards, please copy to a temp file instead.

make/kube
HartS: Ah.. since `cd $GIT_ROOT` is at the beginning of this script, I think relative references are OK.

Makefile
jandubois: This kind of change should not be in the `Makefile` but in the script instead.

jandubois:
> Substitute DOMAIN and UAA_HOST in dotenv file & clean up after deploy

I don't know what "dotenv" file here means; I assume you are talking about the `bin/settings/*` files.  I think this is the wrong approach.  You can provide an override mechanism via env variables, but any local changes in those files need to be kept around. *None* of the `make` targets should modify any of the files under version control.

make/kube
andreas-kupries: Agreed. Missed that. The `run` script does not have however. I would propose to go with how Jan is doing it, i.e. add `cd ${GIT_ROOT}` at the beginning and then remove all the other `GIT_ROOT` refs.

make/run
andreas-kupries: I would propose to go with how Jan is doing it, i.e. add cd ${GIT_ROOT} at the beginning and then remove all the other GIT_ROOT refs.

Makefile
andreas-kupries: Good point. +1 from me.

make/run
HartS: This is done now

HartS:
@jandubois Yes, the dotenv files I was referring to are the `bin/settings/*.env` files (in scf) and `env/*.env` (in uaa-fissile-release). These are parsed by fissile with godotenv, but effectively just sourced for helm deployments.

I've made changes to avoid permanently altering the `network.env` (in scf) and `defaults.env` (in uaa-fissile-release)

The way these work is effectively to *append* `DOMAIN=$DOMAIN` to the relevant files (and also `UAA_HOST=uaa.$DOMAIN` to `network.env` in scf), for making kube/helm with fissile, and then deleting the appended lines afterwards. For `make/run` the file is not modified at all (even temporarily).

make/run
HartS: I used `dig +short` in the end

Makefile
```diff
@@ -25,7 +25,15 @@ ${FISSILE_BINARY}: bin/dev/install_tools.sh
...
 
 certs: uaa-certs
-	${GIT_ROOT}/bin/generate-dev-certs.sh cf bin/settings/certs.env
+	# If DOMAIN is set, place DOMAIN and UAA_HOST at the end of network.env temporarily
+	# Line continuations are necessary to trap EXIT after running script generation
+	if test -n "${DOMAIN}"; then \
+		echo "DOMAIN=${DOMAIN}" >> bin/settings/network.env; \
+		echo "UAA_HOST=uaa.${DOMAIN}" >> bin/settings/network.env; \
+		trap "sed -i '$d' bin/settings/network.env && \
+			sed -i '$d' bin/settings/network.env" EXIT; \
```
andreas-kupries: This deletion does not seem to work, for me.
Also note the possible quoting hell (`''`-quoting inside of `""`-quoting.).

The outer `""`-quoting likely sees `$d` as variable to interpolate. Which is not defined. Thus the interpolation is empty. At that point the generated `sed` command looks like `sed -i '' bin/settings/network.env` and would do nothing, I believe. Which would match what I saw, with the appended lines not getting deleted.

`make cert -n` confirms.

I suspect the `$` will have to be `\`-escaped. No. That does not work either :(
It looks to be `make` itself doing the interpolation regardless of the quoting.
Adding a `\` generates `sed -i '\' bin/settings/network.env` which is not a usable sed command either.

andreas-kupries: Maybe put the code for the trap into a script, with no arguments?

make/kube
```diff
@@ -6,15 +6,18 @@ GIT_ROOT=${GIT_ROOT:-$(git rev-parse --show-toplevel)}
...
 source .envrc
 
-CREATE_HELM_CHART=false
-BUILD_TARGET=kube
+DOMAIN=${DOMAIN:-}
+if test -n "$DOMAIN"; then
+    # If DOMAIN is set, place DOMAIN and UAA_HOST at the end of network.env temporarily
+    echo "DOMAIN=${DOMAIN}" >> bin/settings/network.env
+    echo "UAA_HOST=uaa.${DOMAIN}" >> bin/settings/network.env
+    trap "sed -i '$d' bin/settings.network.env && sed -i '$d' bin/settings/network.env" EXIT
```
andreas-kupries: /home/vagrant/scf/src/make/kube: line 14: d: unbound variable
The outer `"`-quoting causes interpolation of `$d`, `d` does not exist.
Has to be `\`-escaped.
Gr. Still not working. No error anymore, but no deletion of lines either.
Because: `sed: can't read bin/settings.network.env: No such file or directory`
The sed seems to run in a different directory context.

andreas-kupries: Nevermind, the path is just a typo, using `.` where it should be `/`

andreas-kupries: Now it works. As in, with my fixes to quoting and to the path.

Makefile
HartS: I was able to find the right combination of escaping for both Make and shell to get this working... see updates! Thanks for finding this

mook-as: Alternatively, we could not modify the `.env` files, and just do `generate-dev-certs.sh -e bin/settings/ -e /tmp/some/dir bin/settings/certs.env`.

jandubois:
> Alternatively, we could not modify the .env files

Yes, please, as I stated before, I find any modification of files under source control unacceptable.

I understand that the scripts are supposed to undo any changes, but I don't want them to make any changes to begin with. :)

PR #1097, opened 2017-08-23
jenkinsfile: update archive name due to packager changes

mook-as:


Jenkinsfile
```diff
@@ -153,7 +153,7 @@ pipeline {
                         source ${PWD}/.envrc
                         set -x
                         unset HCF_PACKAGE_COMPILATION_CACHE
-                        rm -f scf-*-amd64-*.zip
+                        rm -f scf-*amd64*.zip
```
andreas-kupries: After ew-reading the zip PR, How about `...-*-amd64.zip` as pattern? A bit more specific.

mook-as: It _would_ be more specific, which is why I didn't choose it… I aimed for something that would match both the old name as well as the new one, in case we decide to change things around _again_.

andreas-kupries: Ok. Approving now.

PR #1100, opened 2017-08-30
kube: install cni-plugins

mook-as:
It was previously all in the `cni` package, which is required by docker. It's now split into a separate package, and is required to run kubelet.

### DIFF
--- a/packer/vagrant-box.json
+++ b/packer/vagrant-box.json
@@ -1,6 +1,6 @@
 {
     "variables": {
-        "version": "2.0.7",
+        "version": "2.0.8",
         "vm_name": "scf-vagrant",
         "ssh_username": "vagrant",
         "ssh_password": "vagrant",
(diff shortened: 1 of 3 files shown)


packer/vagrant-box.json
```diff
@@ -247,12 +247,23 @@
...
                 "sudo -i -u vagrant helm init",
+                "systemctl stop kube-apiserver kube-controller-manager kube-proxy kube-scheduler kubelet docker",
+                "rm -rf /data/docker/image/* /data/docker/containers/* /data/docker/overlay2",
                 "# Install additional packages",
                 "zypper --non-interactive install jre-openjdk-headless make zip"
             ]
         },
         {
             "type": "shell",
+            "only": ["vmware-iso", "virtualbox-iso", "qemu"],
```
mook-as: Why isn't this just shared between everyone?

HartS: Because for the Triton build it needs to delete docker stuff in /data, not /var/lib/docker. See above

mook-as: Oh right, we don't do the bind mount. 👍

PR #1102, opened 2017-08-31
Restore old state of persi support as a starting point of investigation.

andreas-kupries:


thardeck:
@andreas-kupries can this be closed? It seems to be stalled for a long time.

PR #1103, opened 2017-09-05
Update java-buildpack to v3.17

alexjh:


andreas-kupries:
Redone :cat: fails differently. Assuming that this is transient and due to me running a java app in parallel.
Accepting and merging.

PR #1104, opened 2017-09-05
Split output of Jenkins builds

alexjh:
Saves output into four different dirs:
* ./ (master)
* ./prs/
* ./branches/
* ./develop/

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -187,27 +187,31 @@ pipeline {
                         )]) {
                             script {
                                 def files = findFiles(glob: 'scf-*amd64*.zip')
-                                def prefix = "${params.S3_PREFIX}"
+                                def subdir = "${params.S3_PREFIX}"
+                                def prefix = ""
 
                                 // If CHANGE_ID env var exists, put the build in the `prs` subdir
                                 // If not, master goes in the root, develop goes in its own dir, and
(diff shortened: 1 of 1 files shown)


alexjh:
That makes sense. Right now the filenames are something like `scf-2.1.0-alpha-pre+cf265.44.ga2984687.linux-amd64.zip`, I'm assuming we'd want to put the additional text at the front so multiple builds of the same branch / PR are grouped together, e.g. `PR-1234-scf-2.1.0-alpha-pre+cf265.44.ga2984687.linux-amd64.zip`

alexjh:
@mook-as, updated to add prefix

Jenkinsfile
```diff
@@ -187,11 +187,31 @@ pipeline {
...
+                                } catch(Exception ex) {
+                                    if (env.BRANCH_NAME == 'develop') {
+                                        subdir = "${params.S3_PREFIX}develop/"
+                                    } else if (env.BRANCH_NAME != 'master') {
+                                        subdir = "${params.S3_PREFIX}branches/"
+                                        prefix = "${BRANCH_NAME}-"
+                                    }
+                                }
+
+                                prefix = java.net.URLEncoder.encode(prefix, "UTF-8")
```
mook-as: This may need to be whitelisted in some configurations. Ah well.

PR #1105, opened 2017-09-05
Enable HA

mook-as:
This makes things go green in a HA world. Not sure if everything's actually working yet, but it seems good enough for smoke tests (🐈 fails because the auth token times out).

### DIFF
--- a/container-host-files/etc/hcf/config/scripts/configure-HA-hosts.sh
+++ b/container-host-files/etc/hcf/config/scripts/configure-HA-hosts.sh
@@ -25,7 +25,7 @@ k8s_api() {
 find_cluster_ha_hosts() {
     local component_name this_component hosts
     component_name="${1}"
-    this_component="$(k8s_api api/v1 "/pods/${HOSTNAME}" | jq -crM '.metadata.labels."skiff-role-name"')"
+    this_component="$(cat /var/vcap/instance/name)"
 
     if test "${this_component}" != "${component_name}" ; then
         # Requesting a different component, use DNS name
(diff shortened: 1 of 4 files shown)


container-host-files/etc/hcf/config/scripts/configure-HA-hosts.sh
```diff
@@ -53,6 +56,14 @@ if test -z "${KUBE_SERVICE_DOMAIN_SUFFIX:-}" ; then
     # bootstrap node.
     export KUBE_SERVICE_DOMAIN_SUFFIX="$(awk '/^search/ { print $2 }' /etc/resolv.conf)"
 fi
+
+export KUBE_COMPONENT_INDEX="${HOSTNAME##*-}"
```
viovanov: iirc we had a deterministic way to transform the random suffix of the hostname to a number (to make sure we don't have collisions, which would be bad for something like the diego cell)

mook-as: Fixed in 5d8015fd7feed0a417d51c75d5643364655060c6

mook-as:
Actually, we might want to do it [in fissile](https://github.com/SUSE/fissile/blob/fffe31a8d0830266ba81a24de88d7d5da9f1ebba/scripts/dockerfiles/run.sh#L72) instead.

container-host-files/etc/hcf/config/scripts/configure-HA-hosts.sh
```diff
@@ -53,6 +56,9 @@ if test -z "${KUBE_SERVICE_DOMAIN_SUFFIX:-}" ; then
     # bootstrap node.
     export KUBE_SERVICE_DOMAIN_SUFFIX="$(awk '/^search/ { print $2 }' /etc/resolv.conf)"
 fi
+
+export KUBE_COMPONENT_INDEX="$(cat /var/vcap/instance/id)"
```
viovanov: @mook-as i think we should expose the ID as an env var [here](https://github.com/SUSE/fissile/blob/9ca01d6a6bd6ab9ddab78e112506ecdc0b592c8e/scripts/dockerfiles/run.sh#L85) and use it before merging.
Also, the pieces that are not `scf` specific in this file should move to `run.sh`

mook-as: Sure, I can do that.

PR #1106, opened 2017-09-07
Provide helm chart

alexjh:
Note that the uaa-fissile-release subdir should point to master before merging SCF.

### DIFF
--- a/make/kube
+++ b/make/kube
@@ -25,5 +25,9 @@ rm -rf "${FISSILE_OUTPUT_DIR}"
 
 fissile build "${BUILD_TARGET}"
 
+if [ "${BUILD_TARGET}" = "helm" ]; then
+    cp Chart.yaml ./helm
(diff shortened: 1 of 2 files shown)


mook-as:
Needs a bump for SUSE/uaa-fissile-release#45 merge

make/kube
```diff
@@ -25,5 +25,9 @@ rm -rf "${FISSILE_OUTPUT_DIR}"
 
 fissile build "${BUILD_TARGET}"
 
+if [ "${BUILD_TARGET}" = "helm" ]; then
+    cp Chart.yaml ./helm
```
mook-as: Actually, the copy target here should be `${FISSILE_OUTPUT_DIR}` (it's what you deleted) right?

alexjh: Yes, and uaa needs the fix as well.

PR #1112, opened 2017-09-12
jenkins: trigger tests on successful publish

mook-as:


Jenkinsfile
```diff
@@ -1,5 +1,33 @@
 #!/usr/bin/env groovy
 // vim: set et sw=4 ts=4 :
+
+String distSubDir() {
```
andreas-kupries: A bit of nit. Based on my understanding of the code this function returns the destination directory where the zip will be/is stored.
So maybe name this `distDestDir` ? Feels clearer to me.
Or maybe `distStoreDir` to account for the other use I saw, when the test stage pulls the zip from S3.

mook-as: Right, I called it that because it replaced the `subdir` variable that already existed. This is the `master/` part of `s3.amazonaws.com/cf-opensusefs2/scf/config/master/scf-xxxyyyzzz.zip`. (It's not used outside of S3.)

mook-as:
Change is being tested automatically — once the automated CI completes for this PR, click on the (hopefully) green checkmark to see that it has triggered a downstream test job.  That job will probably fail, but that's separate (`$DOMAIN` isn't being set in a bunch of jobs).

andreas-kupries:
I see, the `continuous-integration/jenkins/pr-merge` currently pending, correct ?

Jenkinsfile
andreas-kupries: I am ok with keeping as is, as said, just a nit. However please make the decision (keep/change) explicit here in the comment.

mook-as: Ah. I'm keeping it, then :)

andreas-kupries: Ok. Now waiting for the jenkins to complete.

PR #1113, opened 2017-09-15
Use java offline buildpack as default

alexjh:


.gitmodules
```diff
@@ -76,3 +76,6 @@
 [submodule "src/cf-usb"]
 	path = src/cf-usb
 	url = https://github.com/SUSE/cf-usb
+[submodule "src/buildpacks/java-offline-buildpack-release"]
```
jandubois: This commit adds the `java-offline-buildpack` submodule but doesn't remove the `java-buildpack`.

The Trello card https://trello.com/c/fgpyEfMj/216-add-back-java-offline-buildpack-and-remove-online-version says:

> @troytop has also agreed that the online version of the buildpack should be removed.

PR #1116, opened 2017-09-18
Add a pre-start script to enable docker in Diego

alexjh:


src/hcf-release/jobs/acceptance-tests/templates/pre-start.erb
```diff
@@ -0,0 +1,11 @@
+#!/bin/sh
+
+set -o errexit
+set -o nounset
+set -x
+
+PATH=$PATH:/var/vcap/packages/cli/bin
+
+cf api --skip-ssl-validation "https://api.$DOMAIN"
+cf auth admin "${CLUSTER_ADMIN_PASSWORD}"
```
mook-as: I don't think we want `-x` with that

alexjh: Fixed.

alexjh:
I added the enable in HATS as well, I think it's reasonable to have them set to the same configuration for the test run.

http://jenkins.howdoi.website/job/scf-test/70/consoleFull

Issue #1115, opened 2017-09-16
How can I upgrade the Kubernetes version to 1.7 in Vagrant?


zhangtbj:
I tried to follow this guide:
https://software.opensuse.org/download.html?project=Virtualization%3Acontainers&package=kubernetes

But it cannot work.

I tried to update manually, I can upgrade, but I don't know why your suse kubernetes 1.7 requires a low version docker docker_1_12_6, so it report runc error during create container.

So I am not sure if I follow the correct step.

Or do you have any guide?

I also try to use packer to create a new virtualbox iso, but also fail because the packages:
    kubernetes-addons-kubedns-1.5.3 \
    kubernetes-node-cni-1.5.3 \
don't exist.

Please help take a look, thanks a lot! :)

mook-as:
Hi there!

You'll need to
1. Update the [list of packages installed](https://github.com/SUSE/scf/blob/develop/packer/scripts/install-kubernetes.sh)
2. Rebuild the [vagrant box](https://github.com/SUSE/scf/tree/develop/packer)

It looks like there's a hard-coded dep of docker=1.12 in the [kubernetes spec file](https://build.opensuse.org/package/view_file/Virtualization:containers/kubernetes/kubernetes.spec?expand=1) (line 101, `%package kubelet` since I can't link to lines directly).  You may need to file a bug in [bugzilla](https://bugzilla.opensuse.org/enter_bug.cgi?classification=7340&product=openSUSE.org&component=3rd%20party%20software&assigned_to=containers-bugowner@suse.de&cc=fcastelli@suse.com&short_desc=Virtualization:containers/kubernetes:%20Bug) or make a request via OBS to get that changed.

It looks like the 1.5.3 packages no longer exist in the devel repos. That would mean a more invasive change to make those work. I do not know at this point what those changes might be.

zhangtbj:
Sorry I don't have permission for bugzilla.

Do you have any plan to upgrade to kube 1.7?

Thanks!

aarondl:
Yes we do. But the timeline is fuzzy. We're intending to start testing on top of 1.7 soon and start ironing out kinks.

mook-as:
It look sl ike we're on docker 1.12.6 because [Kubernetes says so](https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG.md#external-dependency-version-information).

Bugzilla signups are open, at least?

zhangtbj:
Thanks a lot all!

[...]

PR #1117, opened 2017-09-20
Add OSC to Jenkins workers

thardeck:
to be able to push files to OBS.


mook-as:
Just want to mention that rebuilding the slave is hitting issues because we're dumb enough to use packages from OBS without branching them first, and they have went away.  See also #1115

PR #1130, opened 2017-09-26
create-release: bosh v2 compat

mook-as:
Also bumps UAA for same

Depends on related PRs:
SUSE/cloudfoundry#57
SUSE/uaa-fissile-release#47
SUSE/fissile#283 (to handle releases with zero packages)

Should probably co-ordinate between all four (especially for the docker image change for `splatform/bosh-cli`).  This one will need an updated uaa-fissile-release (on the merge) and the fissile version string.

### DIFF
--- a/.envrc
+++ b/.envrc
@@ -2,16 +2,12 @@
 # This file is both used via direnv, as well as by directly sourcing it in a
 # shell (when being run from CI).
 
-# prevent cd from printing the directory it changes to. This breaks
-# cd/pwd constructions (See **).
-unset CDPATH
-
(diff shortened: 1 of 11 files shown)


mook-as:
Current status:
- It's actually (sometimes) reusing the bosh cache now, that's good.
- It sometimes falls over not finding blobs, though:
>`Cannot find blob named 'auctioneer/24b1cee040b2e008880251c6c94d927ad3ede5bd' with SHA1 '59301026584098bce499b4296c0821bcd7654426'`
- Using `bosh create-release --sha2` isn't going to be useful for us, because fissile doesn't understand those yet.

PR #1134, opened 2017-09-29
Commit source tars

dragonchaser:
This PR adds functionality for publishing the source codes used to build scf on obs

### DIFF
--- a/make/tar-sources
+++ b/make/tar-sources
@@ -2,11 +2,14 @@
 
 set -o errexit -o nounset
 
+DIRECTORY_BLACKLIST="*-buildpack cflinuxfs2 opensuse42 busybox golang*"
+
 GIT_ROOT=${GIT_ROOT:-$(git rev-parse --show-toplevel)}
 
-HOMEDIR=~ ## TODO: verify if this works on jenkins
+HOMEDIR=~ 
 BASEDIR="$HOMEDIR/.fissile/compilation"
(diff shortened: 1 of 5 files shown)


make/tar-sources
```diff
@@ -0,0 +1,99 @@
...
+
+pushd $BASEDIR > /dev/null
+DIRS=`ls`
+for DIR in $DIRS 
+do
+  DO_TAR=true
+  SOURCE_PATH="$DIR/sources/var/vcap/source"
+  SOURCE_NAME=`ls "$SOURCE_PATH"`
+  SOURCE_SIZE=`du "$SOURCE_PATH"|tail -n1`
+  for ITEM in $DIRECTORY_BLACKLIST
```
aduffeck: Maybe the blacklisting could be done simpler using `| egrep -v` or something?

mook-as:
Yay the build finished!

But it didn't seem to commit any sources. Am I doing something wrong?
> `ls: cannot access '/data/jenkins/scf/source-output/*.tar.gz': No such file or directory`

It does appear that it attempted to create the sources:
> `Creating tar archive for nsync (stage: capi-release source size: 8.0M	0442ecc49988724a68ddce1611603608b905a9e1/sources/var/vcap/source)`

mook-as:
I'm going to merge this anyway, while you look into the problem :)

PR #1140, opened 2017-10-03
Add topologyKey to affinity rules

alexjh:
Kubernetes 1.7+ requires this key, it was optional in earlier versions.


mook-as:
Kurbernetes successfully started deploying on [Jenkins](http://jenkins.howdoi.website/blue/organizations/jenkins/scf-test/detail/scf-test/143/pipeline). Don't need to actually wait for it to complete deploying (because it would have rejected the configs if they were broken).

PR #1146, opened 2017-10-06
Split publish step into publish-docker and publish-s3

dragonchaser:
This allows for publishing the SLE stemcell to the staging registry but
not to S3. 
**NOTE:** The publish stage in the pipeline has been replaced by two stages (publish-docker and publish-s3). When merging this the jenkins triggers have to be adapted to this new pipeline layout.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -55,9 +55,14 @@ pipeline {
             description: 'Remove build artifacts that should normally not be reused',
         )
         booleanParam(
-            name: 'PUBLISH',
+            name: 'PUBLISH_DOCKER',
             defaultValue: true,
-            description: 'Enable publishing',
+            description: 'Enable publishing to docker',
+        )
+        booleanParam(
+            name: 'PUBLISH_S3',
+            defaultValue: true,
+            description: 'Enable publishing to amazon s3',
         )
         booleanParam(
             name: 'TEST',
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -171,11 +188,31 @@ pipeline {
...
+                }
                 sh '''
                     set -e +x
                     source ${PWD}/.envrc
                     set -x
                     unset HCF_PACKAGE_COMPILATION_CACHE
+
+                    # Make sure to use the .envrc variables if the override variables are empty
+                    if [ -z "${FISSILE_STEMCELL}" ]; then
+                        unset FISSILE_STEMCELL
```
mook-as: But you're doing all this _after_ sourcing `.envrc` ~6 lines above.

Also note that `.envrc` mostly uses `${FOO:-default}` anyway, which uses the default value if there's no override.  Meaning you don't actually need this block at all.

aduffeck: Nice catch. 

And we thought `${FOO:-default}` would only fall back to the default if `FOO` isn't set at all, but not if it's empty. That's good to know.

aduffeck:
@mook-as One unresolved issue is that the pull request status only reflects the last of the two builds. We couldn't figure out how this works exactly. Do you think it's possible to have a variable for the check name somewhere, for example, so that we'll get two results for the two builds?

mook-as:
Unfortunately, as far as I know the status is all _automatic_, which means we don't have any knobs to tweak. I'm not actually all that familiar with the [plugin](https://plugins.jenkins.io/blueocean-github-pipeline), though.

PR #1148, opened 2017-10-10
Prep for fissile supporting `--tag-extra`. 

andreas-kupries:
Use current commit hash as additional key.

 🚧 (Needs official fissile build to be added to `bin/common/versions.sh`)

See https://github.com/SUSE/fissile/pull/287 for the fissile change.

PR #1156, opened 2017-10-16
Use latest git commit as part of the fissile image tag

jandubois:
This replaces #1148, which only set `FISSILE_TAG_EXTRA` via `.envrc`, so would not notice any checkout changes while inside the directory.


bin/fissile
```diff
@@ -0,0 +1,3 @@
+#!/bin/sh
+export FISSILE_TAG_EXTRA=$(git rev-parse HEAD)
```
mook-as: Doesn't this need to be in the correct directory?

bin/dev/install_tools.sh
```diff
@@ -21,7 +21,7 @@ mkdir -p "${SCF_BIN_DIR}"
 export SCF_BIN_DIR="$(unset CDPATH ; cd "${SCF_BIN_DIR}" && pwd)"
 
 echo "Fetching fissile $fissile_url ..."
-wget -q "$fissile_url"   -O - | tar xz --to-stdout fissile > "${FISSILE_BINARY}"
+wget -q "$fissile_url"   -O - | tar xz --to-stdout fissile > "${FISSILE_BINARY}.binary"
```
mook-as: Only if you're going to change this anyway: I think I've seen `foo.real` more often than `foo.binary`.

PR #1157, opened 2017-10-17
Add tests to Jenkins main job

aarondl:


Jenkinsfile
```diff
@@ -221,6 +266,104 @@ pipeline {
...
+                '''
+            }
+        }
+
+        stage('deploy') {
+			when {
+				expression { return params.TEST_SMOKE || params.TEST_BRAIN || params.TEST_CATS }
+			}
+            steps {
+                timeout(time: 2, unit: 'HOURS') {
```
mook-as: There's still a 10h global timeout for this pipeline; I assume that's still in effect? Would a wipe build + full (including 🐱 ) test run + source tarring blow the budget?

Jenkinsfile
```diff
@@ -221,6 +266,104 @@ pipeline {
...
+                    sh """
+                        kubectl create -f - <<< '{"kind":"StorageClass","apiVersion":"storage.k8s.io/v1","metadata":{"name":"'"${JOB_BASE_NAME}-${BUILD_NUMBER}-"'hostpath"},"provisioner":"kubernetes.io/host-path"}'
+
+                        # Unzip the bundle
+                        rm -rf unzipped
+                        mkdir unzipped
+                        unzip -e \$(ls | grep -e 'scf-*linux-amd64*.zip') -d unzipped
```
mook-as: Do we need `ls` + `grep`? Isn't this just `unzip -e scf-*linux-amd64*.zip -d unzipped`?

Jenkinsfile
aarondl: Fixed.

Jenkinsfile
aarondl: Fixed.

Jenkinsfile
```diff
@@ -175,6 +224,11 @@ pipeline {
             }
             steps {
                 sh '''
+                    kubectl get namespace | awk '/-scf|-uaa/ {print $1}' | xargs --no-run-if-empty kubectl delete ns
+                    while kubectl get namespace | grep 'scf|uaa'; do
```
mook-as: Your `awk` and `grep` expressions don't match; it's possible that you'll end up never exiting this loop because you never deleted a matching namespace.

Jenkinsfile
```diff
@@ -325,24 +466,30 @@ pass = ${OBS_CREDENTIALS_PASSWORD}
...
-                    )
-                }
-            }
+    }
+
+    post {
+        always {
+            sh """#!/bin/bash
+            set -o xtrace
+            if kubectl get storageclass ${jobBaseName()}-${BUILD_NUMBER}-hostpath ; then
```
mook-as: I think this is from a previous buggy change. The storage class name no longer depends on the job name / build number.

aarondl: Changed to just hostpath/hostpath

Jenkinsfile
aarondl: Added dashes to the grep

PR #1163, opened 2017-10-25
Add persi support to SCF

viovanov:
This includes NFS support and some ceph pieces.
The ceph pieces are mostly commented out, because the ceph release is not working.


Makefile
```diff
@@ -101,6 +101,12 @@ mysql-release:
 usb-release:
 	make/bosh-release src/cf-usb/cf-usb-release
 
+ceph-release:
+	BOSH_RELEASE_NAME=cephfs make/bosh-release src/cephfs-bosh-release
```
mook-as: This can just be `make/bosh-release src/cephfs-bosh-release cephfs`

.gitmodules
```diff
@@ -82,3 +82,9 @@
 [submodule "src/cf-sle12-release"]
 	path = src/cf-sle12-release
 	url = https://github.com/SUSE/cf-sle12-release.git
+[submodule "src/cephfs-bosh-release"]
+	path = src/cephfs-bosh-release
```
mook-as: We _could_ just make the submodule be called `src/cephfs-release` even though it doesn't match the repo name…

README.md
```diff
@@ -673,3 +674,85 @@ pod-status --watch
...
+
+### Allow access to the NFS server
+
+- Security group JSON file (nfs-sg.json)
+```json
+[
+    {
+        "destination": "192.168.77.77",
+        "protocol": "tcp",
+        "ports": "2049,111,662,875,32803,892"
```
mook-as: Might want to sort that port list for sanity

README.md
```diff
@@ -673,3 +674,85 @@ pod-status --watch
...
+    {
+        "destination": "192.168.77.77",
+        "protocol": "udp",
+        "ports": "111,662,875,2049,32769,892"
+    }
+]
+```
+
+```bash
+# Create the security group - JSON below
```
mook-as: JSON above.

container-host-files/etc/hcf/config/role-manifest.yml
```diff
@@ -278,49 +278,6 @@ roles:
       properties.cf_mysql.mysql.galera_healthcheck.endpoint_password: ((MYSQL_GALERA_HEALTHCHECK_ENDPOINT_PASSWORD))
       properties.cf_mysql.proxy.api_password: ((MYSQL_PROXY_ADMIN_PASSWORD))
       properties.cf_mysql.proxy.proxy_ips: '["mysql-proxy.((KUBE_SERVICE_DOMAIN_SUFFIX))"]'
-- name: mysql-broker
```
mook-as: Wait what? That doesn't seem right. Why are we deleting it (even though it's terrible, it's a thing).

bin/settings/settings.env
```diff
@@ -70,7 +65,7 @@ UAADB_PASSWORD=uaaadmin_password
 UAA_ADMIN_CLIENT_SECRET=admin_secret
 UAA_CC_CLIENT_SECRET=cc_client_secret
 UAA_CLIENTS_CC_ROUTING_SECRET=cc_routing_secret
-UAA_CLIENTS_CC_SERVICE_SECRET=cc_service_secret
+#UAA_CLIENTS_CC_SERVICE_SECRET=cc_service_secret
```
mook-as: Why is this being removed?

container-host-files/etc/hcf/config/role-manifest.yml
```diff
@@ -1334,6 +1290,90 @@ roles:
                 values:
                 - diego-access
             topologyKey: "beta.kubernetes.io/os"
+# - name: ceph-broker
```
mook-as: ‽‽‽

viovanov:
Don't merge until we have a good stemcell

bin/settings/settings.env
viovanov: Because the mysql broker went away

container-host-files/etc/hcf/config/role-manifest.yml
viovanov: This should have been removed a couple of sprints ago, it was only meant as a demo talk point while USB wasn't available.

container-host-files/etc/hcf/config/role-manifest.yml
viovanov: I didn't want to get rid of that piece, because we'll need to make it work anyway.
But I guess I'll stash it away in a different branch

viovanov:
With the last fissile stemcell bump, this is working again (tested manually)

container-host-files/etc/hcf/config/role-manifest.yml
mook-as: Then we should do that in a different PR, not this one. This would make it difficult to figure out later :)

bin/settings/settings.env
```diff
@@ -85,3 +80,5 @@ USE_STAGER_PRIVILEGED_CONTAINERS=false
 HTTP_PROXY=
 HTTPS_PROXY=
 NO_PROXY=.local,127.0.0.1
+PERSI_NFS_BROKER_PASSWORD=nfs-password
+PERSI_NFS_MYSQL_PASSWORD=mysql-nfs-password
```
mook-as: EOL

bin/settings/settings.env
mook-as: Then this still needs to live in the _mysql broker went away_ PR :)

container-host-files/etc/hcf/config/scripts/patches/fix_persi_nfsdriver.sh
```diff
@@ -0,0 +1,51 @@
+set -e
+
+PATCH_DIR="/var/vcap/jobs-src/nfsv3driver/templates"
+SENTINEL="${PATCH_DIR}/${0##*/}.sentinel"
+
+if [ -f "${SENTINEL}" ]; then
+  exit 0
+fi
+
+read -r -d '' ctl_patch <<'PATCH' || true
```
mook-as: I don't understand why we do this weird `read` business, instead of just
```bash
patch --force -d "${PATCH_DIR}" <<'PATCH'
@@ patch file goes here @@
PATCH
```

container-host-files/etc/hcf/config/scripts/patches/fix_persi_nfsdriver.sh
```diff
@@ -0,0 +1,51 @@
+set -e
```
mook-as: The file name could be more explanatory, and it could use a comment on what it does. Something like `persi_nfsdriver_user_group.sh`, and
```bash
# SUSE compatibility: this forces the `cvcap` user to have its own group, and users `usermod` instead of `adduser`
```

src/hcf-release/jobs/configure-scf/spec
```diff
@@ -48,3 +48,10 @@ properties:
   cf_mysql.mysql.port:
     description: 'Port used by the service broker and bound applications to reach the service'
     default: 3306
+  nfsbroker.url:
+    description: 'Internal URL of the Persi NFS broker'
```
mook-as: FWIW, you don't actually need to quote values when they're unambiguously strings.

README.md
```diff
@@ -683,3 +684,84 @@ export FISSILE_DOCKER_AUTH=basic
...
+
+### Running a test NFS server
+
+```bash
+# Enable NFS modules
+sudo modprobe nfs
+sudo modprobe nfsd
+
+docker run -d --name nfs \
+    -v "/home/toaster/tools/nfs_share:/exports/foo" \
```
mook-as: What's a toaster?

README.md
```diff
@@ -683,3 +684,84 @@ export FISSILE_DOCKER_AUTH=basic
...
+    -p 875:875/udp \
+    -p 875:875/tcp \
+    -p 2049:2049/udp \
+    -p 2049:2049/tcp \
+    -p 32769:32769/udp \
+    -p 32803:32803/tcp \
+    -p 892:892/udp \
+    -p 892:892/tcp \
+    --privileged \
+    viovanov/nfs-server /exports/foo
```
mook-as: We need sources for that :D

mook-as:
On a happy note, this does appear to work for me 🎉 So it's just a matter of fixing the review stuff, not figuring out anything wrong

container-host-files/etc/hcf/config/scripts/patches/fix_persi_nfsdriver.sh
viovanov: consistency - the other scripts are the same;
iirc we tried this before and something didn't work - we could try again maybe? (add a card to refactor all the scripts)

README.md
viovanov: it's a cylon centurion

README.md
viovanov: I don't think we do - this is just a how-to on setting up a quick NFS server - we're not delivering any of it. But here it is: https://github.com/viovanov/nfs-server

container-host-files/etc/hcf/config/scripts/patches/fix_persi_nfsdriver.sh
viovanov: I think it's better to have one file for the job, and we can comment in the file like you mentioned.

PR #1166, opened 2017-10-26
Jenkinsfile: Allow branch name override as a param

mook-as:
This is to work with develop/master builds where we don't pull from the normal PR-based multi-branch pipeline.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -162,12 +162,19 @@ pipeline {
 
     stages {
         stage('trigger_sles_build') {
-          when {
+            when {
                 expression { return params.TRIGGER_SLES_BUILD }
-          }
-          steps {
-            build job: 'scf-sles-trigger', wait: false, parameters: [string(name: 'JOB_NAME', value: env.JOB_NAME)]
-          }
+            }
+            steps {
+                build(
+                    job: 'scf-sles-trigger',
+                    wait: false,
+                    parameters: [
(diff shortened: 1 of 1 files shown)


mook-as:
Marking as WIP until Jenkins builds the PR branch (to make sure I didn't break anything).

Jenkinsfile
```diff
@@ -162,12 +162,19 @@ pipeline {
...
-            build job: 'scf-sles-trigger', wait: false, parameters: [string(name: 'JOB_NAME', value: env.JOB_NAME)]
-          }
+            }
+            steps {
+                build(
+                    job: 'scf-sles-trigger',
+                    wait: false,
+                    parameters: [
+                        string(name: 'JOB_NAME', value: env.JOB_NAME),
+                        string(name: 'BRANCH_NAME', value: params.BRANCH_NAME),
```
aduffeck: I think we have to pass `env.BRANCH_NAME` if `params.BRANCH_NAME` is not set in order to cover all cases, no?

mook-as: You're right; fixed in 1de8875bfcae66579063360a7be5a72dd913829f

Though now that I think about it, basing things on `scm.GIT_BRANCH` would have been easier, but needs more testing  (because I am not sure that actually works).

mook-as:
This looks really stale; I'll recreate the PR the next time it looks relevant.

PR #1167, opened 2017-10-26
Move the SLE trigger to the end of the build

aduffeck:
We have some build failures in Jenkins since a few days where docker images
can't be cleaned up because they are still referenced by running containers.
I think this might be related to the concurrent builds that were re-enabled
recently.

This commit moves the SLES trigger to the very end of the build, so although
we have concurrent builds in Jenkins the openSUSE and the SLES builds will be
practically sequential which hopefully fixes these issues.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -153,14 +153,6 @@ pipeline {
     }
 
     stages {
-        stage('trigger_sles_build') {
-          when {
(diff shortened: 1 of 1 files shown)


aduffeck:
This is supposedly already fixed by another pr.

PR #1168, opened 2017-10-26
Add cached libvirt

aarondl:


packer/scripts/install-helm-images.rb
```diff
@@ -0,0 +1,70 @@
...
+  images = []
+
+  templates.each do |filename|
+      File.read(filename).each_line do |line|
+          # Check for a line that starts with image:
+          match = line.match(/^\s*-?\s*image:\s*"?(.*?)"?$/)
+          next if match.nil?
+
+          # Parse and replace any helm templates
+          images.push match[1].gsub(/{{\s*\.Values.([^}\s]+)\s*}}/) { lookup_nested(values, $1) }
```
mook-as: `/{{\s*\.Values.([^}\s]+)\s*}}/` should be the same as `/{{\s*\.Values.(.+?)\s*}}/`

packer/scripts/install-helm-images.rb
```diff
@@ -0,0 +1,70 @@
...
+      end
+  end
+
+  return images
+end
+
+def find_chart_dirs(path)
+  return Dir.glob(File.join(path, '**/Chart.yaml'))
```
mook-as: No need for `return`, the result of the last expression is implicitly returned.
(Also for the previous function)

packer/scripts/install-helm-images.rb
```diff
@@ -0,0 +1,70 @@
...
+end
+
+if ARGV.length == 0
```
mook-as:
```ruby
if ARGV.empty?
```

packer/scripts/install-helm-images.rb
```diff
@@ -0,0 +1,70 @@
...
+  puts "Usage: #{File.basename($0)} <directory>"
+  puts
+  puts "Must supply a directory to be recursively searched for helm charts."
+  exit 1
+end
+
+chart_dirs = find_chart_dirs(ARGV[0])
+charts_images = chart_dirs.map { |dir| scrape_images(File.dirname(dir)) }
```
mook-as:
```ruby
images = chart_dirs.flat_map { |dir| scrape_images(File.dirname(dir)) }
…
images.each do |image|
```
Actually, what's the `File.dirname` doing there anyway?
Hmm, we probably want to tack `.uniq` on the end of that `flat_map` too.

packer/scripts/install-helm-images.rb
```diff
@@ -0,0 +1,70 @@
...
+
+charts_images.each do |chart_images|
+  chart_images.each do |image|
+    puts "Caching #{image}"
+    Open3.popen2e("docker pull #{image}") do |stdin, stdouterr, thread|
```
mook-as: I _think_ we can get away with `Open3.pipeline(['docker', 'pull', image])` (which would only return after the command finishes, so you don't need to worry about it).

packer/scripts/install-helm-images.rb
```diff
@@ -0,0 +1,70 @@
...
+      stdin.close
+      IO.copy_stream(stdouterr, STDOUT) # Copy until EOF which means its probably done
+      thread.join                       # Wait until the process has exited and then close our pipe
+      stdouterr.close
+    end
```
mook-as: We probably still want to handle errors if `docker pull` errors out though.  At least, `exit 1` if any of the commands failed.

packer/scripts/stop-kube-and-docker.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/sh
+
+set -o errexit
+set -o verbose
+
+systemctl stop kube-apiserver kube-controller-manager kube-proxy kube-scheduler kubelet docker
```
mook-as: EOL before EOF

packer/scripts/install-cf.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/sh
+
+set -o errexit
+set -o verbose
+
+# Get predefined version of cf-cli
+wget -O - https://s3-us-west-1.amazonaws.com/cf-cli-releases/releases/v6.32.0/cf-cli_6.32.0_linux_x86-64.tgz \
```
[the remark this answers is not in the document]
aarondl: Fixed.

packer/scripts/install-cap-archives.sh
```diff
@@ -0,0 +1,16 @@
...
+set -o verbose
+
+# Fetch predetermined version of scf
+SCF_URL=https://github.com/SUSE/scf/releases/download/2.2.0-beta/scf-2.2.0-beta-pre.cf265.1.gd50775b4.linux-amd64.zip
+SCF_ARCHIVE=$(basename $SCF_URL)
+wget $SCF_URL
+unzip $SCF_ARCHIVE
+
+# Fetch predetermined version of the ui
+CONSOLE_URL=https://github.com/SUSE/stratos-ui/releases/download/0.9.5/console-helm-chart-0.9.5.tgz
```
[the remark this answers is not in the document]
aarondl: It would be nice, but that's for a time when we have the luxury of time. Need to get this merged and be done with it.

packer/scripts/install-helm-images.rb
```diff
@@ -0,0 +1,70 @@
+#!/usr/bin/env ruby
+
+require 'open3'
+require 'yaml'
+
+def lookup_nested(hash, key)
```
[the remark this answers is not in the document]
aarondl: Deleted file. No longer complicated.

packer/scripts/stop-kube-and-docker.sh
aarondl: Fixed.

packer/http/user-limits.conf
```diff
@@ -3,3 +3,5 @@
 
 vagrant hard nproc unlimited
 vagrant soft nproc unlimited
+scf hard nproc unlimited
+scf soft nproc unlimited
```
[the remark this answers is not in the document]
aarondl: I think this is harmless enough. Going to leave it and not template it for simplicity. If the user doesn't exist the OS simply doesn't care.

packer/scripts/install-cap-archives.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/sh
+
+set -o errexit
+set -o verbose
+
+# Fetch predetermined version of scf
+SCF_URL=https://github.com/SUSE/scf/releases/download/2.2.0-beta/scf-2.2.0-beta-pre.cf265.1.gd50775b4.linux-amd64.zip
+SCF_ARCHIVE=$(basename $SCF_URL)
```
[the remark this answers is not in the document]
aarondl: Couldn't download shellcheck, did the next best thing and added every symbol that could possibly offer protection.

PR #1169, opened 2017-10-27
Bump to cf-deployment v0.26.0

mook-as:
This is generally a bit ahead of where cf-release v272 is. Also switched to upstream smoke tests release.


make/uaa-images
```diff
@@ -15,4 +15,4 @@ GIT_ROOT=${GIT_ROOT:-$(git rev-parse --show-toplevel)}
 unset FISSILE_REPOSITORY
 
 make -C "${GIT_ROOT}/src/uaa-fissile-release" \
-    build
+    images
```
jandubois: This change does not work unless there is a corresponding change to let `make releases` also build the UAA releases. Currently the UAA releases are build by `make images` in the SCF release. ☹️

PR #1177, opened 2017-11-06
`make kube` / `make helm` / `make dist` should output into output/

mook-as:
This moves all the generated files into `output/` instead of being at the top level. This makes it easier to check in files that we want to include verbatim inside the helm chart dist bundles later.

### DIFF
--- a/.envrc
+++ b/.envrc
@@ -106,5 +106,5 @@ export FISSILE_STEMCELL=${FISSILE_STEMCELL:-splatform/fissile-stemcell-opensuse:
 # The defaults are all for helm.
 # The make/kube script performs overrides for kube output.
 export FISSILE_USE_MEMORY_LIMITS=false
-export FISSILE_OUTPUT_DIR="${FISSILE_OUTPUT_DIR:-${PWD}/helm}"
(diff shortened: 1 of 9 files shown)


mook-as:
But the point of `make/clean` _is_ to delete things from the build. It was more a bug that it previously didn't clean up the kube/helm artifacts (which is why they accumulate on the Jenkins slaves).

andreas-kupries:
Ok. So removal of the new stuff put into `output` is ok.
Still, the removal of `output/bin`, which is not a build result, should still be addressed. That subdir  should not be removed, because that breaks the build system (`fissile` will be gone).

(And this PR seems to be a good place to do it, given that it already works on `output`.

mook-as:
There's no reason it would break the build, since [`make compile`] etc. would just download fissile again.

[`make compile`]: https://github.com/SUSE/scf/blob/40e706a2e2e737da3bd375e012b61ce43c952327/Makefile#L209

andreas-kupries:
Ok. ... Next argument ... At
https://github.com/SUSE/scf/blob/develop/make/clean#L10
the `make/clean` calls on a script which tries to use the just-deleted fissile.
No automatic download at this point.

andreas-kupries:
Moved the change to `make/clean` into a separate PR, see https://github.com/SUSE/scf/pull/1209

PR #1181, opened 2017-11-08
Bump to cf-release 278

jandubois:
All releases have been bumped to the version recommended in the 278 release notes, except for cflinuxfs2-release, which has been bumped to the latest version available right now.

The tcp_emitter job of the routing release no longer exists; it has been merged with the regular route_emitter job of the diego-release. This job is now co-located on the diego_cell so that local_mode can be enabled.

The tcp_emitter will hang indefinitely (like the tcp_router) if UAA isn't fully working by the time it starts, so the new wait-for-uaa job provides a pre-start script to implement this.


zhangtbj:
Hi @jandubois ,

Glad to hear that you are upgrading CF version. :)

Actually, we are upgrading our SCF to CF 270 too. And we find some issues (Most of are UAA issue). I provide here, hope they are helpful.

**1, UAA Default Authorities for new zone is changed:**
https://github.com/cloudfoundry/uaa-release/releases/tag/v43

So the new cf zone which SCF created will miss cloud_controller.write and cloud_controller.read permission, like this:
```
    "userConfig": {
      "defaultGroups": [
        "openid",
        "password.write",
        "uaa.user",
        "approvals.me",
        "profile",
        "roles",
        "user_attributes",
        "uaa.offline_token"
      ]
    }
```
So SCF has to update post-deployment-setup uaa-create-user job run.erb to do like this:
```
uaac curl -X POST /identity-zones -H 'Content-Type: application/json' -d '{
    "id": "cf",
    "subdomain": "cf",
    "name": "cf",
    "config": {
        "userConfig": {
            "defaultGroups": [
                "openid",
                "cloud_controller.read",
                "cloud_controller.write",
                "password.write",
                "uaa.user",
                "approvals.me",
                "profile",
                "roles",
                "user_attributes",
                "uaa.offline_token"
            ]
        }
    }
}'
```
Then new user can have cc permission. Or new user will get this error after cf login:
```
API endpoint: https://api.sleeve-dev.cc (API version: 2.92.0)
User: bmxsvt1@us.ibm.com
No org or space targeted, use 'cf target -o ORG -s SPACE'
FAILED
Error finding available orgs
Invalid auth token: Invalid Auth Token
```

2, openid should be the default group, that member admin cannot be added to it, will report an error about member cannot be added to default group (Sorry I don't have that log right now) in the end of post-deployment uaa-create-user rub.rb: `uaac member add '<%= authority %>' <%= username %>`

3, Should use regex to check if cf zone exists?
In my test:
if uaac curl <%= curl_ssl %> "/identity-zones/${UAA_ZONE_ID}" | grep --quiet '404 Not Found' ; then
I cannot get correct result from above command because new UAA return **404** only.

But SCF hard code '404 Not Found'.

There is no other issues, and we can startup CF 270 on our env correctly.

Hope these information are helpful, and I can also provide a PR fix if need.

Thanks!

jandubois:
Hi @zhangtbj,

> Actually, we are upgrading our SCF to CF 270 too.

I'm slightly confused, the `develop` branch of SCF was already at 272, so what does it mean that you are updating to 270? Were you still running from an older branch?

But you raise a good point about UAA; it seems like we have been neglecting updating UAA as well; `src/uaa-fissile-release/src/uaa-release` is still at v29, which looks quite outdated.

We will look into this soon, but maybe not as part of this PR.

zhangtbj:
@jandubois 

Oh yes, I mean UAA. I think UAA is older than CF release:
It should be uaa-release version v29.

But current UAA version is > v50+

jandubois:
Rebased branch on top of `develop` to see if that fixes the Jenkins build problems.

This also adds the cf-usb test using the mysql sidecar, so provides more testing for the tcp routing changes.

mook-as:
Purely because I forgot to do it when I bumped to 272: There's a [CF version](https://github.com/SUSE/scf/blob/ef4b2c54c2203d757571a5e974ce58ab94ce738c/bin/common/versions.sh#L35) string.

jandubois:
> Purely because I forgot to do it when I bumped to 272: There's a CF version string.

Oh no, I didn't update that one either... Thankfully this PR is not merged yet! 😄

zhangtbj:
Let me know if you need any other info. I think you will also see those errors when you are upgrading. :)

jandubois:
> Let me know if you need any other info.

Thanks! We now have a card for bumping UAA on our planning board, and I added a link to your comment to it. We'll do this as a separate task from this PR.

viovanov:
I've had CATs pass on an earlier version of this, but I want to run again.
Will merge after I see it work

viovanov:
Passed smoke tests

aarondl:
@zhangtbj We really need to bump our UAA. It'd be great to have a PR that applied your fixes, do you mind submitting it?

zhangtbj:
HI @aarondl , Sure, I have submitted a PR to fix the 3 issues:
https://github.com/SUSE/scf/pull/1198

Pls review, for the third change, I think we also need to confirm if it is good for SCF, or we can refine the change.

Pls let me know if you have any commit. Thanks!

PR #1183, opened 2017-11-09
Bump stemcell version for configgin 0.14.0

mook-as:
This changes behaviour such that we will run bootstrap on any pod where it is the only pod with a given image version.  This is to support upgrades and similar scenarios where the image versions will change, but not necessarily starting from the lowest-numbered pod.

### DIFF
--- a/bin/common/versions.sh
+++ b/bin/common/versions.sh
@@ -24,7 +24,7 @@ export STAMPY_MAJOR=$(echo "$STAMPY_VERSION" | sed -e 's/\.g.*//' -e 's/\.[^.]*$
 
 # Used in: .envrc
 
-export FISSILE_STEMCELL_VERSION=${FISSILE_STEMCELL_VERSION:-42.2-15.g38c573e-29.44}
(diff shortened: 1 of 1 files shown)


bin/common/versions.sh
```diff
@@ -24,7 +24,7 @@ export STAMPY_MAJOR=$(echo "$STAMPY_VERSION" | sed -e 's/\.g.*//' -e 's/\.[^.]*$
 
 # Used in: .envrc
 
-export FISSILE_STEMCELL_VERSION=${FISSILE_STEMCELL_VERSION:-42.2-15.g38c573e-29.44}
+export FISSILE_STEMCELL_VERSION=${FISSILE_STEMCELL_VERSION:-42.2-15.g38c573e-29.48}
```
andreas-kupries: Well, change looks good, and stemcell is properly pulled into vagrant.
The branch fails for `make diego-release`, and only that release.
I do not see that as blocker for merge.
I suspect an independent problem. (And maybe even with my checkout, despite doing `submodule sync && submodule update --init`).

PR #1184, opened 2017-11-13
Create separate submit request for each package

jonsger:
Hi,

this change creates for each package it's own submit request. It automatically supersedes the old SR when needed :) There is no need to checkout the project, the script can be executed anywhere.

For unknown reasons the approach with `osc submitrequest --seperate-requests $PROJECT:reviewed` doesn't work.

Jonathan


thardeck:
Works now with `--message`

PR #1189, opened 2017-11-15
Increase send_timeout for blobstore uploads from 2 to 6 min

jandubois:
Currently our buildpacks contains assets for 3 different stacks, and on some slower test deployments they fail to upload within the default 2 min allowed by HTTPClient.


mook-as:
Added as https://trello.com/c/Kgc9xftl/ for the backlog. Good to merge now.

PR #1194, opened 2017-11-16
Add an "-opensuse" suffix to the chart name and dir when on opensuse …

jimmykarily:
…stemcell

### DIFF
--- a/make/kube
+++ b/make/kube
@@ -40,10 +40,14 @@ rm -rf "${FISSILE_OUTPUT_DIR}"
 fissile build "${BUILD_TARGET}"
 
 if [ "${BUILD_TARGET}" = "helm" ]; then
+    if [[ $FISSILE_STEMCELL =~ ^.*fissile-stemcell-opensuse.*$ ]] ; then
+      chart_name_suffix="-opensuse"
+    fi
+
     cat > "${FISSILE_OUTPUT_DIR}/Chart.yaml" << EOF
 apiVersion: v1
 description: A Helm chart for SUSE Cloud Foundry
-name: scf
+name: cf${chart_name_suffix:-}
 version: ${APP_VERSION}
 EOF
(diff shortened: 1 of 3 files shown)


make/kube
```diff
@@ -40,10 +40,14 @@ rm -rf "${FISSILE_OUTPUT_DIR}"
...
 if [ "${BUILD_TARGET}" = "helm" ]; then
+    if [[ $FISSILE_STEMCELL =~ ^.*fissile-stemcell-opensuse.*$ ]] ; then
+      chart_name_suffix="-opensuse"
+    fi
+
     cat > "${FISSILE_OUTPUT_DIR}/Chart.yaml" << EOF
 apiVersion: v1
 description: A Helm chart for SUSE Cloud Foundry
-name: scf
+name: cf${chart_name_suffix}
```
aduffeck: Does this work for the SLE case? I'd think that the `nounset` option makes it fail if it's not set in line 44.

make/kube-dist
```diff
@@ -14,10 +14,16 @@ for variant in kube helm ; do
...
-    mkdir -p "${tmp_dir}"/{cf,uaa}
+    # Add "-opensuse" suffix to the chart directory if we are on opensuse stemcell.
+    # The directory should match the chart name.
+    if [[ $FISSILE_STEMCELL =~ ^.*fissile-stemcell-opensuse.*$ ]] ; then
+      suffix="-opensuse"
+    fi
 
-    cp -r "${GIT_ROOT}/${variant}/"* "${tmp_dir}/cf"
-    cp -r "${GIT_ROOT}/src/uaa-fissile-release/${variant}/"* "${tmp_dir}/uaa"
+    mkdir -p "${tmp_dir}"/{cf,uaa}$suffix
```
aduffeck: Same here, is this `nounset` proof?

make/kube
jimmykarily: Fixed both

PR #1196, opened 2017-11-16
rolemanifest fixes

jandubois:


container-host-files/etc/hcf/config/role-manifest.yml
```diff
@@ -2666,7 +2666,7 @@ configuration:
     properties.diego.rep.log_level: '"((GO_LOG_LEVEL))((#LOG_LEVEL))((/LOG_LEVEL))"'
     properties.diego.rep.server_cert: '"((REP_SERVER_CERT))"'
     properties.diego.rep.server_key: '"((REP_SERVER_KEY))"'
-    properties.diego.rep.trusted_certs: '((#TRUSTED_CERTS))"((TRUSTED_CERTS))"((/TRUSTED_CERTS))((#TRUSTED_CERTS))null((/TRUSTED_CERTS))'
+    properties.diego.rep.trusted_certs: '((#TRUSTED_CERTS))"((TRUSTED_CERTS))"((/TRUSTED_CERTS))'
```
mook-as: Your commit message has a couple typos; it was `it used ((^TRUSTED_CERTS)) instead of ((^TRUSTED_CERTS)` but it should be `it used ((#TRUSTED_CERTS)) instead of ((^TRUSTED_CERTS))`. (Wrong symbol in the first example, and missing close parenthesis in the second one.)

PR #1198, opened 2017-11-18
Fix new UAA release 3 migration issues

zhangtbj:
Fix 3 issues for new UAA release:
1,  use `404` instead of `404 not found` to check if identity-zone exists, or will report error because UAA message change.

2, Add cloud_controller.read and cloud_controller.write groups for new identity-zone, because new UAA release remove them that new user cannot use CC (Report Invaild Token Error):
https://github.com/cloudfoundry/uaa-release/releases/tag/v43

3, We need to discuss the third change, because I change it in run.erb to use `set +e` and `set -e` to ignore the error, because `openid` is the default group, when you add admin to group openid, the process of post-deployment-setup will report and fail in the end:
```
api/0:/# uaac member add 'openid' admin

error response:
{
  "error_description": "Trying to add member to default group",
  "error": "member_already_exists",
  "message": "Trying to add member to default group"
}
```

Pls refer:
https://github.com/cloudfoundry/uaa/blob/master/docs/Sysadmin-Guide.rst
```
some default groups are always available and do not need to be explicitly populated: openid, password.write, cloud_controller.read, cloud_controller.write, tokens.read, tokens.write
```

But I suggest remove `openid` from `CLUSTER_ADMIN_AUTHORITIES` property directly if no any other component use it.

Pls let me know if you have any suggestion. Thanks!

### DIFF
--- a/src/hcf-release/jobs/uaa-create-user/templates/run.erb
+++ b/src/hcf-release/jobs/uaa-create-user/templates/run.erb
@@ -77,12 +77,28 @@ status "Checking for UAA zone ${UAA_ZONE_ID}..."
 uaac target <%= uaac_ssl %> '<%= p("scf.uaa.root-zone.url") %>'
 uaac token client get default_zone_admin --secret '<%= p("uaa.admin.client_secret") %>'
 
-if uaac curl <%= curl_ssl %> "/identity-zones/${UAA_ZONE_ID}" | grep --quiet '404 Not Found' ; then
+if uaac curl <%= curl_ssl %> "/identity-zones/${UAA_ZONE_ID}" | grep --quiet '404' ; then
     status "Creating UAA zone ${UAA_ZONE_ID}"
     uaac curl <%= curl_ssl %> -X POST /identity-zones -H 'Content-Type: application/json' -d '{
         "id": "'"${UAA_ZONE_ID}"'",
         "subdomain": "'"${UAA_ZONE_ID}"'",
-        "name": "'"${UAA_ZONE_ID}"'"
+        "name": "'"${UAA_ZONE_ID}"'",
+        "config": {
+          "userConfig": {
+            "defaultGroups": [
+                "openid",
+                "cloud_controller.read",
+                "cloud_controller.write",
+                "password.write",
+                "uaa.user",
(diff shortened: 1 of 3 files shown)


src/hcf-release/jobs/uaa-create-user/templates/run.erb
```diff
@@ -77,12 +77,28 @@ status "Checking for UAA zone ${UAA_ZONE_ID}..."
...
+if uaac curl <%= curl_ssl %> "/identity-zones/${UAA_ZONE_ID}" | grep --quiet '404' ; then
     status "Creating UAA zone ${UAA_ZONE_ID}"
     uaac curl <%= curl_ssl %> -X POST /identity-zones -H 'Content-Type: application/json' -d '{
         "id": "'"${UAA_ZONE_ID}"'",
         "subdomain": "'"${UAA_ZONE_ID}"'",
-        "name": "'"${UAA_ZONE_ID}"'"
+        "name": "'"${UAA_ZONE_ID}"'",
+        "config": {
+          "userConfig": {
+            "defaultGroups": [
```
mook-as: Hmm, this should probably be coming from the opinions or something (next to where the list of authorities lives).

src/hcf-release/jobs/uaa-create-user/templates/run.erb
```diff
@@ -161,7 +177,14 @@ if uaac group get '<%= authority %>' | grep -q '  displayname: <%= username %>'
...
 elif ! uaac group get '<%= authority %>' | grep -q "    value: ${user_id}" ; then
     status "+ %s" <%= username %>
+    set +e
     uaac member add '<%= authority %>' <%= username %>
+    if [ $? -eq 0 ]; then
+    	status "Add %s to %s successfully" <%= username %> '<%= authority %>'
+    else
+        status "Add %s to %s failed but ignore the error first" <%= username %> '<%= authority %>'
+    fi
+    set -e
```
mook-as: Given that the list of default groups are under our control (see above), would it be possible to check it against that list and skip?

zhangtbj:
Hi @mook-as I also verified in my env. It works:

These fixes work. Thanks too! :)

PR #1199, opened 2017-11-21
Task: Add syslog-drain-release to replace the lost syslog-drain-binder

andreas-kupries:
https://trello.com/c/dCsdt8hO/457-add-cf-syslog-drain-release-to-scf
Inclusion needed now (Bump of CF versions came with loggregator release which lost the old job.)
Under active work now.
Rebased to head of dev.
Currently not standing up, some issue with the new cert params and the secret-generator.


container-host-files/etc/hcf/config/role-manifest.yml
```diff
@@ -2102,6 +2199,57 @@ configuration:
     default: support@example.com
     description: Support contact information for the cluster
     required: true
+  - name: SYSLOG_ADAPT_CERT
+    secret: true
+    generator:
+      id: syslog_adapt_cert
+      type: Certificate
+      value_type: certificate
```
jandubois: You should have a `subject_names` list of "Subject Alternate Names" for the cert and/or a `role_name` that will be used to generate the various internal names with `KUBE_SERVICE_DOMAIN_SUFFIX` etc.

container-host-files/etc/hcf/config/role-manifest.yml
```diff
@@ -2102,6 +2199,57 @@ configuration:
...
+    description: PEM-encoded certificate
+    required: true
+  - name: SYSLOG_ADAPT_KEY
+    secret: true
+    generator:
+      id: syslog_adapt_cert
+      type: Certificate
+      value_type: private_key
+      key_length: 4096
```
jandubois: The `key_length` isn't used by the secrets generator and can/should be omitted.

andreas-kupries: Will do.

container-host-files/etc/hcf/config/role-manifest.yml
andreas-kupries: Hm. Will check existing certs and keys for examples. Sort of like the `cn` fields which got added to the `opinions.yml` for this, I guess ?

container-host-files/etc/hcf/config/role-manifest.yml
andreas-kupries: See commit 84c47e4c94470197738340ded7aadd7453b71f56

container-host-files/etc/hcf/config/role-manifest.yml
andreas-kupries: See commit 84c47e4c94470197738340ded7aadd7453b71f56.
Note, my names might be wrong.
As nothing changes. I.e. roles report same issue as before, with validation.

PR #1200, opened 2017-11-21
Fix configure-HA-hosts k8s_api error issue 

zhangtbj:
SCF should check error in this script. in my environment, I don't have kube serviceaccount, so if I cannot get `k8s_api apis/apps/v1beta1` for replicas, the error will be skipped. then some of statefulset pods may work incorrectly, such as consul and nats. Without set -e, the pod will be started without any error exit, but user don't know why CF function doesn't work fine.

### DIFF
--- a/container-host-files/etc/hcf/config/scripts/configure-HA-hosts.sh
+++ b/container-host-files/etc/hcf/config/scripts/configure-HA-hosts.sh
@@ -39,6 +39,15 @@ find_cluster_ha_hosts() {
         statefulset_name="$(k8s_api api/v1 "/pods/${HOSTNAME}" | jq -crM '.metadata.annotations."kubernetes.io/created-by"' | jq -crM .reference.name)"
         replicas=$(k8s_api apis/apps/v1beta1 "/statefulsets/${statefulset_name}" | jq -crM .spec.replicas)
 
+        if [ "${statefulset_name}" == "" ]; then
+            echo "Cannot get statefulset name from kubernetes API, exit"
+            exit 1
+        fi
+        if [ "${replicas}" == "" ]; then
+            echo "Cannot get replicas from kubernetes API, exit"
+            exit 1
+        fi
+        
         # Return a list of all replicas
         local hosts=""
         for ((i = 0 ; i < "${replicas}" ; i ++)) ; do
@@ -48,10 +57,14 @@ find_cluster_ha_hosts() {
     fi
 }
 
-export KUBE_CONSUL_CLUSTER_IPS="$(find_cluster_ha_hosts consul)"
-export KUBE_NATS_CLUSTER_IPS="$(find_cluster_ha_hosts nats)"
-export KUBE_ETCD_CLUSTER_IPS="$(find_cluster_ha_hosts etcd)"
-export KUBE_MYSQL_CLUSTER_IPS="$(find_cluster_ha_hosts mysql)"
+KUBE_CONSUL_CLUSTER_IPS="$(find_cluster_ha_hosts consul)"
+export KUBE_CONSUL_CLUSTER_IPS
+KUBE_NATS_CLUSTER_IPS="$(find_cluster_ha_hosts nats)"
+export KUBE_NATS_CLUSTER_IPS
+KUBE_ETCD_CLUSTER_IPS="$(find_cluster_ha_hosts etcd)"
+export KUBE_ETCD_CLUSTER_IPS
+KUBE_MYSQL_CLUSTER_IPS="$(find_cluster_ha_hosts mysql)"
+export KUBE_MYSQL_CLUSTER_IPS
 
 unset k8s_api
 unset find_cluster_ha_hosts


zhangtbj:
Hi @mook-as ,

Hmm.... This is an interesting thing. I go through the error path again. And I found add set -e for this script is not a good way. I think because SCF cannot get error from curl. `echo $?` return 0 although it is wrong. Let me show the log:

For example, for nats, nats-0 works fine because it goes into:
```
    elif test "${KUBE_COMPONENT_INDEX}" == "0" ; then
        # This is index 0; don't look for other replicas, this needs to bootstrap
        echo "[${component_name}-0.${component_name}-set.${KUBE_SERVICE_DOMAIN_SUFFIX}]"
```

but for nats-1, it should go into:
```
   else
        # Find the number of replicas we have
        local statefulset_name replicas i
        statefulset_name="$(k8s_api api/v1 "/pods/${HOSTNAME}" | jq -crM '.metadata.annotations."kubernetes.io/created-by"' | jq -crM .reference.name)"
        replicas=$(k8s_api apis/apps/v1beta1 "/statefulsets/${statefulset_name}" | jq -crM .spec.replicas)
.....
```

But I don't have kube api access permission right now(I didn't create serviceaccount for that), then this k8s_api method should return error, result should not eq 0, then exit the script.

But from log you can see:
+++++ cat /var/run/secrets/kubernetes.io/serviceaccount/token
+++++ cat /var/run/secrets/kubernetes.io/serviceaccount/namespace
++++ curl --silent --cacert /var/run/secrets/kubernetes.io/serviceaccount/ca.crt -H 'Authorization: bearer eyJhbGciOiJSUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJrdWJlcm5ldGVzL3NlcnZpY2VhY2NvdW50Iiwia3ViZXJuZXRlcy5pby9zZXJ2aWNlYWNjb3VudC9uYW1lc3BhY2UiOiJjZiIsImt1YmVybmV0ZXMuaW8vc2VydmljZWFjY291bnQvc2VjcmV0Lm5hbWUiOiJkZWZhdWx0LXRva2VuLTcxcGRzIiwia3ViZXJuZXRlcy5pby9zZXJ2aWNlYWNjb3VudC9zZXJ2aWNlLWFjY291bnQubmFtZSI6ImRlZmF1bHQiLCJrdWJlcm5ldGVzLmlvL3NlcnZpY2VhY2NvdW50L3NlcnZpY2UtYWNjb3VudC51aWQiOiJhMTRlZTVlNy1jOWNmLTExZTctYTlkYy0xMjExYjM4MTNhNjIiLCJzdWIiOiJzeXN0ZW06c2VydmljZWFjY291bnQ6Y2Y6ZGVmYXVsdCJ9.kj2NqtOcZpUQ8uZZwTIoTdJ-_O5Rqw9VfUHoSwcu5AqkDA2E_XCedgi9ha7HoGg0wCR1n-PpEF3q5DHT0obPx2Cmj82Y-DfiVczuv8fqczJtVSeRY45TDpc-syOWFf7ctNIgoaaVzsN_J797kQFTIiScQkprYPaIdWn_8HwW3Y29saA5Bg5i22-rR1YSNayN_sRIt15kfyn2xISlntshQiBnJdzOUejnQKLw7ZDn09RWnt84mkwAQuKWCmHzzQwesJ2oEee2EKwXnnr1YJRLLRxOhxXZJ3c4iHWMuOEDAwInHhGlL6Sjc0UtfJ15ckvJ8g2bYLZOIKMp6YuhGBvBOO26MJ1CakaBkNrc_o2asf-LbQ6dn_Da9yu-N33cJaiB4RMkijDItBczrDsNsJQnqRjpykxOfRTPpnSt-5q4i7zQ9sw-MR_StRmIcyZUtmxqgwRGCff5kjjaG13eLEGk5PSsLZXMXtF8d2N8V2d-ggsVVLhu1jaIndH5x0rkysbm7AoElyTzKgA48L-V3CH_DFFoCMv6S8_0sbv_V-i208lMIKMj0oYX6c73kNJSLfWEu_DnDSmz2hEEGxwPId5GWEop-MfJLE_UsgAdcDPjTFB3mQTQFtfnHl4QtxERgm3tI_nyUioXKMudcNRXFBQWAsv8yhfQpJCULKaPkbgXsl4' https://172.21.0.1:443/api/v1/namespaces/cf/pods/nats-1
**parse error: Invalid numeric literal at line 1, column 5**
+++ this_component=
+++ test '' '!=' nats
+++ echo '["nats.cf.svc.cluster.local"]'
++ export 'KUBE_NATS_CLUSTER_IPS=["nats.cf.svc.cluster.local"]'
++ KUBE_NATS_CLUSTER_IPS='["nats.cf.svc.cluster.local"]'

From the log, you can see there is an error `parse error: Invalid numeric literal at line 1, column 5` but is not caught correctly, then my nats ips is totally wrong. it should be KUBE_NATS_CLUSTER_IPS='[nats-0.nats-set.cf.svc.cluster.local,nats-1.nats-set.cf.svc.cluster.local]', but it is not, then my nats for router or other components will work fail.

So I think SCF should catch this curl error in this script then ask run.sh to stop/exit as expected.

Hope this is helpful! :)

mook-as:
The parse error comes from `jq` (which makes sense: curl error messages are unlikely to be properly formatted JSON).  The problem here, though, is that the _second_ `jq` doesn't error out (there isn't a duplicate message), and since we don't have `pipefail` on, that counts as a success. Somehow.
> ```bash
> $ echo 💩 | jq -crM '.metadata.annotations."kubernetes.io/created-by"' | jq -crM .reference.name ; echo $?
> parse error: Invalid numeric literal at line 2, column 0
> 0
> ```

I think an adequate fix in this case is to just test if `${statefulset_name}` is empty after the call, and `exit 1` from there (with an appropriate message for the operator).  Same for `${replicas}`, I guess.

zhangtbj:
Hi @mook-as , I have modified my PR. I also think check statefulset_name and replicas if is empty is a good way. but just check is not enough. because SCF use export like `export KUBE_NATS_CLUSTER_IPS="$(find_cluster_ha_hosts nats)"`, the exit command will be caught by export and method. So have to add `trap bash' EXIT` to force exit main script if find_cluster_ha_hosts method gets exit.

Here is an example test script you can test:
```
root@cf-lite-dev:~/backup/tmp# cat test.sh
set -x
set -e

show_error() {
  # trap 'bash' EXIT
  echo "this is an error"
  exit 1
}
echo "before source"
export result="$(show_error)"
echo "after source"
```

Without trap command, the script won't exit even get `exit 1`.

I have add my change in this PR, pls let me know if you have any good suggestion. Thanks! :)

mook-as:
Unfortunately, you just end up spawning a new shell instead when you `trap bash EXIT`:
```bash
$ ./test.sh </dev/null
++ set -e
++ echo 'before source'
before source
+++ show_error
+++ trap bash EXIT
+++ echo 'this is an error'
+++ exit 1
+++ bash
++ export 'result=this is an error'
++ result='this is an error'
++ echo 'after source'
after source
```

I think the correct thing to do (according to [shellcheck]) is to split it into two lines:
```bash
result="$(show_error)"
export result
```

This seems to give the correct output:
```bash
$ ./test.sh </dev/null
++ set -e
++ echo 'before source'
before source
+++ show_error
+++ trap bash EXIT
+++ echo 'this is an error'
+++ exit 1
+++ bash
++ result='this is an error'
```

[shellcheck]: https://github.com/koalaman/shellcheck/

zhangtbj:
Hi @mook-as ,

Yes, I agree to split export into two lines. I have verified and submitted again.

PR #1201, opened 2017-11-21
Create helm charts pr script

jimmykarily:


bin/dev/create_helm_charts_pr.sh
```diff
@@ -0,0 +1,67 @@
...
+unzip $bundle
+popd
+
+cd kubernetes-charts-suse-com
+# Remove old charts
+rm -rf stable/cf
+rm -rf stable/uaa
+# Place the new ones
+cp -r ../bundle/helm/cf stable/
+cp -r ../bundle/helm/uaa stable/
```
jimmykarily: This will only work for sle charts. The opensuse ones will have an `-opensuse` suffix. Should we ever want to put the opensuse charts on the repo, we whould handle that here.

bin/dev/create_helm_charts_pr.sh
```diff
@@ -0,0 +1,71 @@
...
+#trap "rm -rf ${TMP}" EXIT
+
+pushd $TMP
+
+# Get the "hub" cli
+# https://hub.github.com/hub.1.html
+wget https://github.com/github/hub/releases/download/v2.3.0-pre10/hub-linux-amd64-2.3.0-pre10.tgz
+tar xvzf hub-linux-*.tgz
+cp hub-linux-*/bin/hub hub
+HUB=$(readlink -f hub)
```
mook-as: Wait, what? The symlink in the tar file comes broken (because it points into a subdirectory that didn't exist in there)?

```bash
$ tar tvf hub-linux-amd64-2.3.0-pre10.tgz --wildcards '*/bin/hub'
-rwxr-xr-x travis/travis 9396782 2017-06-26 11:35 hub-linux-amd64-2.3.0-pre10/bin/hub
```

It looks like a standard file to me. Might want to rephrase this bit as
```bash
curl -L https://github.com/github/hub/releases/download/v2.3.0-pre10/hub-linux-amd64-2.3.0-pre10.tgz | tar xvz --wildcards --strip-components=2 '*/bin/hub'
HUB="${PWD}/hub"
```

bin/dev/create_helm_charts_pr.sh
```diff
@@ -0,0 +1,71 @@
...
+chmod +x $HUB
+
+# escape user and password
+USER=$(ruby -e "require 'cgi'; puts CGI.escape(ENV['GITHUB_USER'])")
+PASS=$(ruby -e "require 'cgi'; puts CGI.escape(ENV['GITHUB_PASSWORD'])")
+# Clone the kubernetes-charts-suse-com github repo
+$HUB clone https://${USER}:${PASS}@github.com/SUSE/kubernetes-charts-suse-com.git
```
mook-as: It looks like `hub` should be handling auth via the environment variables already; do we need to put them in the URL too?

(Using ruby means we'll need to make sure we _have_ ruby on the slaves… which we actually do, but still.)

bin/dev/create_helm_charts_pr.sh
```diff
@@ -0,0 +1,71 @@
...
+
+mkdir bundle
+pushd bundle
+wget "$CAP_BUNDLE"
+bundle=$(ls *.zip)
```
mook-as: This is way easier if we just make sure to give an explicit name for this.
```bash
mkdir bundle
wget -O bundle.zip "${CAP_BUNDLE}"
unzip -d bundle/ bundle.zip
```

Or do you actually use the bundle file name for later use? In that case you might want to strip off the `.zip` part…

bin/dev/create_helm_charts_pr.sh
```diff
@@ -0,0 +1,71 @@
...
+rm -rf stable/uaa${suffix:-}
+# Place the new ones
+cp -r ../bundle/helm/cf${suffix:-} stable/
+cp -r ../bundle/helm/uaa${suffix:-} stable/
+
+$HUB config user.email "cf-ci-bot@suse.de"
+$HUB config user.name "${GITHUB_USER}"
+$HUB checkout -b $bundle
+$HUB -c core.fileMode=false add .
+$HUB commit -m "Submitting "$bundlev
```
mook-as: I don't think `$bundlev` is bound? Is this a typo?

Jenkinsfile
```diff
@@ -506,6 +506,15 @@ pass = ${OBS_CREDENTIALS_PASSWORD}
                                 )
                             }
                         }
+                        sh """
+                          set -e +x
+                          FILE=\$(ls scf-*linux-amd64*.zip)
+                          # Escape twice or the url will be unescaped when passed to the Jenkins form. It will then not work in the script.
+                          ENCODED_FILE_NAME=\$(ruby -e "require 'cgi'; puts CGI.escape('\${FILE}')")
+                          ENCODED_CAP_BUNDLE_URI=\$(ruby -e "require 'cgi'; puts CGI.escape('https://s3.amazonaws.com/${params.S3_BUCKET}/${params.S3_PREFIX}${distSubDir()}${distPrefix()}\${ENCODED_FILE_NAME}')")
+                          ENCODED_BUILD_URL=\$(ruby -e "require 'cgi'; puts CGI.escape('${BUILD_URL}')")
```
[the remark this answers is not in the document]
jimmykarily: This is intentionally manual because we don't want to open PRs against [the helm repo backend](https://github.com/suse/kubernetes-charts-suse-com) for every build. The release manager [is the one that decides](https://github.com/SUSE/cloudfoundry/wiki/CAP-Release-Process#sle) what charts should be released. I'm not sure if it makes sense to document this (business) logic here.

Regarding escaping, can we mix Groovy and sh? I mean, if we prepare the escaped string in Groovy how can we expose them to the sh script? We tried to do that but couldn't make it work.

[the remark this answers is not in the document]
mook-as: Given that it's in a `"""` block, yes; `${}` is actually more like `eval()` in groovy, I think.  Of course, since this block is so simple we could also port this to groovy instead (we grab the list of files where we do the S3 upload, so there's code from there to copy off of)

bin/dev/create_helm_charts_pr.sh
```diff
@@ -0,0 +1,71 @@
+#!/bin/bash
+
+# This script is used to open a PR against this repository:
+# https://github.com/suse/kubernetes-charts-suse-com
+# It will fetch the bundle specified by CAP_BUNDLE and will update the helm
+# charts in the above repo.
+
+set -e
+
+if [ -z $GITHUB_USER  ]; then
```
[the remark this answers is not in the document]
jimmykarily: Fixed.

bin/dev/create_helm_charts_pr.sh
```diff
@@ -0,0 +1,71 @@
...
+  exit 1
+fi
+
+if [[ -z "${CAP_BUNDLE}" ]]; then
+  echo "CAP_BUNDLE not set"
+  exit 1
+fi
+
+TMP=$(mktemp -d -p $PWD)
+#trap "rm -rf ${TMP}" EXIT
```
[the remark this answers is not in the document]
jimmykarily: Sure, we have this commented out for debugging. Enabling...

bin/dev/create_helm_charts_pr.sh
```diff
@@ -0,0 +1,71 @@
...
+  echo "GITHUB_PASSWORD environment variable not set"
+  exit 1
+fi
+
+if [[ -z "${CAP_BUNDLE}" ]]; then
+  echo "CAP_BUNDLE not set"
+  exit 1
+fi
+
+TMP=$(mktemp -d -p $PWD)
```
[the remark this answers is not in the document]
jimmykarily: Changed to `TMP_WORKDIR`

bin/dev/create_helm_charts_pr.sh
jimmykarily: `readlink -f` was one way to get the absolute path to the file (since it is not in the $PATH to call it directly). You solution is more elegant. Changing.

bin/dev/create_helm_charts_pr.sh
jimmykarily: We need the `$bundle` later indeed. Will use `basename` for that. Changing.

bin/dev/create_helm_charts_pr.sh
jimmykarily: typo it is, fixing.

bin/dev/create_helm_charts_pr.sh
jimmykarily: `hub` only uses the `GITHUB_USER` and `GITHUB_PASSWORD` environment variables to authenticate against the github API. The prompt you get when using `hub clone` with `https` comes from `git` itself. `git` doesn't use env vars to authenticate. There are other ways to do that, but in the end [it seems](https://git-scm.com/docs/git-credential-store#_storage_format) that they might still need escaping because all they store is this url above.

mook-as: Actually, when is this script run? Via Jenkins or via somebody typing it on the command line?  Because we could totally just use the Jenkins-stored creds for that…

jimmykarily: It will run on Jenkins. These are the Jenkins creds coming as GITHUB_USER and GITHUB_PASSWORD env vars. Maybe we could have Jenkins pulling in the repo (can we?) but we would still need the creds in order to push later in this script.

mook-as: Yeah, we can have Jenkins pull in the repo (that's how we check out the Jenkinsfile for the main scf pipelines).

I guess the other option is to use `GIT_ASKPASS` (but that needs to be a path to a program that echos the password, not the password itself), and it's not necessarily any better.

jimmykarily: We completely removed this escaping logic by setting the `GIT_SSH_COMMAND` in the wrapper script in Jenkins. `git clone` now works with ssh.

PR #1204, opened 2017-11-22
jenkins: refer to pipeline param as $param.*

mook-as:
This fixes deploying SCF on the test kube (probably)

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -336,7 +336,7 @@ pipeline {
             steps {
                 sh """
                     set -e +x
-                    source ${PWD}/.envrc
+                    source \${PWD}/.envrc
                     set -x
 
                     kubectl delete storageclass hostpath || /bin/true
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -354,7 +354,7 @@ pipeline {
                     ./unzipped/cert-generator.sh -d "${domain()}" -n ${jobBaseName()}-${BUILD_NUMBER}-scf -o unzipped/certs
 
                     suffix=""
-                    if echo "$FISSILE_STEMCELL" | grep -qv "fissile-stemcell-sle"; then
+                    if echo "${params.FISSILE_STEMCELL}" | grep -qv "fissile-stemcell-sle"; then
```
jimmykarily: I think the reason we `source` the .envrc above is to make `FISSILE_STEMCELL` available in `sh`. Now we get the stemcell in Jenkins space so is there still a reason to source .envrc?

PR #1212, opened 2017-11-24
pod-status: don't re-column the output

mook-as:
We previously used awk to find the right field to check for readiness. Unfortunately, this meant that the columns were all changed, and we had to run `column -t` on it; this doesn't work well if any of the columns had whitespace within it (e.g. when the pod status had a longer explanation on why it wasn't working).

This switches to using a perl one-liner that doesn't affect the column layout, so that issue does not occur.


alexjh:
LGTM

PR #1221, opened 2017-12-01
Delete index files if cache files are missing

alexjh:
Deleting the entire dev_builds dir causes the bosh-cache files to
be orphaned.

This includes a fix to the UAA bin/create-release.sh script.

~~~*NOTE* This needs https://github.com/SUSE/uaa-fissile-release/pull/63 to be merged to master first, so the submodule pointer can be updated.~~~

### DIFF
--- a/bin/create-release.sh
+++ b/bin/create-release.sh
@@ -15,6 +15,27 @@ fi
 release_path="$1"
 release_name="$2"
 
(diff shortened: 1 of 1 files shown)


alexjh:
@mook-as I updated the submodule pointer for UAA to master

PR #1227, opened 2017-12-06
Bump cf-usb submodule for UAA token fetching fix

alexjh:


andreas-kupries:
OK. This is just the SCF PR for a previously reviewed cf-usb PR (author: @mook-as, reviewed @alexjh).
LGTM and no need for another round of testing.

PR #1230, opened 2017-12-08
Remove ETCD and Consul from SCF

viovanov:
Since the bump to loggregator 101.3 and the availability of diego-locket, these to pets are no longer required.
Both consul and etcd were clusterable components, difficult to manage.
There's also one less agent per role (consul_agent) and our patch count is reduced.


viovanov:
Passed smoke tests locally

PR #1231, opened 2017-12-08
Cleanup cert gen without consul and etcd

jandubois:


jandubois:
This PR was an experimental combination of #1224 and #1230. Both have been merged individually now.

PR #1235, opened 2017-12-12
Vagrant kube1.8

aarondl:


alexjh:
The Vagrant box build and passed tests (other than the syslog drain). Can you add the version bump so I can merge?

PR #1257, opened 2017-12-19
j8BtwZ51 Make SUSE_STACK configurable

jandubois:
An SCF build will container either the opensuse42 stack or the sle12 stack, but not both. The SUSE stack will be configured as the default stack.

Techically the other stack is still included in the base image package directory, but it will be completely inaccessible from the user level.


container-host-files/etc/scf/config/role-manifest.yml
```diff
@@ -2428,6 +2437,7 @@ configuration:
     properties.capi.cc_uploader.mutual_tls.ca_cert: '"((INTERNAL_CA_CERT))"'
     properties.capi.cc_uploader.mutual_tls.server_cert: '"((CC_UPLOADER_CRT))"'
     properties.capi.cc_uploader.mutual_tls.server_key: '"((CC_UPLOADER_KEY))"'
+    properties.capi.nsync.lifecycle_bundles: '{"buildpack/((SUSE_STACK))":"buildpack_app_lifecycle/buildpack_app_lifecycle.tgz","buildpack/cflinuxfs2":"buildpack_app_lifecycle/buildpack_app_lifecycle.tgz","docker":"docker_app_lifecycle/docker_app_lifecycle.tgz"}'
```
mook-as: Huh, I'm surprised this works. The old opinions were a list of strings (which contains colons in them), not a hash.

That gets used for [bulker](https://github.com/cloudfoundry/capi-release/blob/50e96e2eeedaedb865099a87fee30508c0a86265/jobs/nsync/templates/nsync_bulker_config.json.erb#L35) and [listener](https://github.com/cloudfoundry/capi-release/blob/50e96e2eeedaedb865099a87fee30508c0a86265/jobs/nsync/templates/nsync_listener_config.json.erb#L29), which goes to [bulker](https://github.com/cloudfoundry/nsync/blob/6709e23af984d298b3217317b5b6078fb6017857/config/config.go#L62) and [listener](https://github.com/cloudfoundry/nsync/blob/6709e23af984d298b3217317b5b6078fb6017857/config/config.go#L79), which are both arrays of strings.

alexjh:
The change LGTM.

It was built with: `SUSE_STACK=sle12 make vagrant-prep`

```
vagrant@vagrant-kube:~/scf> cf stacks
Getting stacks in org org / space space as admin...
OK

name         description
cflinuxfs2   Cloud Foundry Linux-based filesystem
sle12        SLE-based filesystem
```

Pushing an app uses the correct stack:

```
requested state: started
instances: 1/1
usage: 32M x 1 instances
urls: go-env.cf-dev.io
last uploaded: Fri Jan 5 22:46:35 UTC 2018
stack: sle12
buildpack: go

     state     since                    cpu    memory        disk          details
#0   running   2018-01-05 05:47:42 PM   0.0%   6.1M of 32M   7.4M of 10M
```

@aarondl can you approve this PR and merge?

PR #1258, opened 2017-12-20
Add brain test to ensure that only uncached buildpacks are shipped

aduffeck:
In order to do so a special inspector buildpack is added which inspects all
other buildpacks at staging time for cached dependencies.
An empty app is then pushed which won't be accepted by any of the regular
buildpacks. Only the inspector buildpack will accept the app if the "uncached"
check passes.


src/scf-helper-release/src/acceptance-tests-brain/test-resources/buildpack_inspector_buildpack/bin/detect
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
+
+# Redirect STDOUT to STDERR, only STDERR is relayed by the cf cli
+exec 1>&2
+
+offenders=`find /tmp/buildpacks -type d -name dependencies`
```
jimmykarily: I wonder if this is checking the right thing. I added the offline java buildpack to my CAP instance and run the brain tests. None failed. Looking in the directory structure of the unzipped buildpack, I don't see any `dependencies` directory. I do see one named `resources/cache/` though which seem to be the one holding the cached dependencies.

PR #1269, opened 2018-01-08
Use a single flag in Jenkins for stack and stemcell

alexjh:
Generating a build with a different base for the stack and stemcell
doesn't make sense.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -182,7 +182,7 @@ pipeline {
             description: 'Docker organization to publish to',
         )
         booleanParam(
-            name: 'USE_SLE',
+            name: 'USE_SLES_BASE',
             defaultValue: false,
             description: 'Generates a build with the SLE stemcell and stack',
         )
@@ -201,8 +201,7 @@ pipeline {
     environment {
         FISSILE_DOCKER_REGISTRY = "${params.FISSILE_DOCKER_REGISTRY}"
         FISSILE_DOCKER_ORGANIZATION = "${params.FISSILE_DOCKER_ORGANIZATION}"
-        USE_SLES_STEMCELL = "${params.USE_SLE}"
-        SUSE_STACK = "${params.USE_SLE ? "sle12" : "opensuse42"}"
+        USE_SLES_BASE = "${params.USE_SLES_BASE}"
     }
 
     stages {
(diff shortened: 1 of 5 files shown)


Jenkinsfile
```diff
@@ -201,7 +201,8 @@ pipeline {
     environment {
         FISSILE_DOCKER_REGISTRY = "${params.FISSILE_DOCKER_REGISTRY}"
         FISSILE_DOCKER_ORGANIZATION = "${params.FISSILE_DOCKER_ORGANIZATION}"
-        USE_SLES_STEMCELL = "${params.USE_SLES_STEMCELL}"
+        USE_SLES_STEMCELL = "${params.USE_SLE}"
+        SUSE_STACK = "${params.USE_SLE ? "sle12" : "opensuse42"}"
```
mook-as: Shouldn't this come from `bin/settings/{sle12,opensuse42}.env`?

alexjh: I don't understand the question... this env var is what's selecting between those two files:
https://github.com/SUSE/scf/blob/a07fb19e143696ff0978a730c9fb79b91e48b2cd/make/kube#L10

Do you mean that we'd source those files directly? Some of the values have spaces in them, so they don't work.
https://github.com/SUSE/scf/blob/a07fb19e143696ff0978a730c9fb79b91e48b2cd/bin/settings/sle12.env#L3

mook-as: Ooof. So we end up having to set the same env var (to the same value) in two places. That doesn't sound fun, but I guess is what we have 😢

alexjh: I've added the quotes, but if they show up in the stack description I'll revert it because it will look strange.

Jenkinsfile
```diff
@@ -182,9 +182,9 @@ pipeline {
             description: 'Docker organization to publish to',
         )
         booleanParam(
-            name: 'USE_SLES_STEMCELL',
+            name: 'USE_SLES_BASE',
```
cornelius: We should be consistent here and call it `USE_SLE_BASE`. SLE is the SUSE Linux Enterprise code stream from which various products are built. SLES is the server product. We are close to the server, but it's not really the SLES product we use so SLE is more accurate. We also use `sle` in other places such as the user-visible name of the stack.

README.md
```diff
@@ -119,17 +119,30 @@ a working system.
     # see the Troubleshooting guide.
     k logs -f cf:^api-[0-9]
     ```
-3. Changing the default STEMCELL
+3. Changing the default STEMCELL and STACK
+
+   The default stemcell and stack are set to OpenSUSE. The versions are defined
+   in `bin/common/versions.sh`.
+
+   To build with the SLES stemcell and stack, the environment variable
```
cornelius: "SLE stemcell and stack"

README.md
```diff
@@ -119,17 +119,30 @@ a working system.
     # see the Troubleshooting guide.
     k logs -f cf:^api-[0-9]
     ```
-3. Changing the default STEMCELL
+3. Changing the default STEMCELL and STACK
+
+   The default stemcell and stack are set to OpenSUSE. The versions are defined
```
cornelius: One more spelling nitpick: It's "openSUSE".

PR #1271, opened 2018-01-10
Increase cf-usb test timeout

alexjh:


src/scf-helper-release/jobs/acceptance-tests-brain/templates/run.erb
```diff
@@ -34,4 +34,4 @@ if [ -d /tests ]; then
 fi
 
 set -x
-testbrain run ${PARAM}
+testbrain run --timeout 1200 ${PARAM}
```
andreas-kupries: Hm. The timeout in `cf_usb_test.sh` was changed from 60 to 300 seconds, i.e. 1 minute to 5.
Here the outer timeout is bumped to 20 minutes.
Would 10 be enough ?

alexjh: Ah, yes, it should be. The Jenkins build took just over 5 minutes to run. Fixed in e38477b0.

andreas-kupries: Thanks. Now waiting on Jenkins.

PR #1272, opened 2018-01-11
Fissile: Update to use stemcell in compile directory

mook-as:
Also update the compile cache script to support it. This is needed to ensure that we will rebuild packages when the stemcell changes (e.g. when we change distros)

We don't attempt to support migrating the old caches over because it is unknown if those were compiled correctly.

### DIFF
--- a/make/compile
+++ b/make/compile
@@ -5,6 +5,7 @@ set -o errexit -o nounset
 GIT_ROOT=${GIT_ROOT:-$(git rev-parse --show-toplevel)}
 
 SCF_PACKAGE_COMPILATION_CACHE=${SCF_PACKAGE_COMPILATION_CACHE:-''}
+STEMCELL_HASH="$(echo -n "${FISSILE_STEMCELL}" | sha1sum | awk '{ print $1 }')"
 
 # Return a list of desired packages with their versions, as in package/00000000
 _get_package_version_list() {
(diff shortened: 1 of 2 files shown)


make/compile
```diff
@@ -28,19 +29,19 @@ _cache() {
...
-        archive="${SCF_PACKAGE_COMPILATION_CACHE}/${package_hash}.tar.xz"
+        archive="${SCF_PACKAGE_COMPILATION_CACHE}/${STEMCELL_HASH}/${package_hash}.tar.xz"
         test -e "${archive}" && {
             # Don't recreate an archive that already exists
             echo "Found:   ${archive}"
             continue
         }
-        test -d "${FISSILE_WORK_DIR}/compilation/${package_hash}/compiled" || {
+        test -d "${FISSILE_WORK_DIR}/compilation/${STEMCELL_HASH}/${package_hash}/compiled" || {
             echo "Missing: ${package_hash}/compiled"
```
andreas-kupries: Should we add the stemcell-hash to the output as well ?
When I looked at the log I was confused at first, because I thought it was looking at and reporting some old directory.

mook-as: We could; I was thinking that might be too much text (which doesn't change anyway), pushing the thing too wide for display in things like Jenkins.

andreas-kupries: Ok.

PR #1274, opened 2018-01-12
Various syslog fixes from bsc#1075602

mook-as:
Please see [bsc#1075602](https://bugzilla.suse.com/show_bug.cgi?id=1075602) for details; this basically unbreaks syslog forwarding

### DIFF
--- a/container-host-files/etc/scf/config/scripts/forward_logfiles.sh
+++ b/container-host-files/etc/scf/config/scripts/forward_logfiles.sh
@@ -121,7 +121,7 @@ function searchTargetDir {
         done
         args=( "${args[@]}" '(' -name '*.log' -a -type f -a -print0 ')' )
         while IFS= read -r -d '' file ; do
-                TARGET_BASENAME="$(basename "$1" .log)"
+                TARGET_BASENAME="$(basename "${file}" .log)"
                 TARGET_NAME="${RSYSLOG_CONF_DIR}/${RSYSLOG_CONF_PREFIX}-${TARGET_BASENAME}.conf"
                 if checkConfigExists ; then
                         echo "${TARGET_NAME} exists"
--- a/container-host-files/etc/scf/config/scripts/forward_logfiles.sh
+++ b/container-host-files/etc/scf/config/scripts/forward_logfiles.sh
@@ -1,60 +1,61 @@
-#used for initial config
+#!/usr/bin/env bash
+
+# This script sets up rsyslog to forward logs in /var/vcap/sys/log to a remote
+# host
+
+# used for initial config
 MAIN_CONFIG="90-vcap.conf"
 RSYSLOG_CONF_DIR="/etc/rsyslog.d"
 
-#used for adding individual logs
-BACKUP_WATCH_DIR=/var/vcap/sys/log #in case no ENV variable is set for RSYSLOG_FORWARDER_WATCH_DIR
-RSYSLOG_CONF_PREFIX=91-vcap
-RSYSLOG_CONF_DIR=/etc/rsyslog.d
+# used for adding individual logs
+BACKUP_WATCH_DIR="/var/vcap/sys/log" #in case no ENV variable is set for RSYSLOG_FORWARDER_WATCH_DIR
+RSYSLOG_CONF_PREFIX="91-vcap"
 IGNORE_DIR="gocode"
 TARGET_NAME=
 TARGET_BASENAME=
 
-SCRIPT_FILE=/usr/sbin/forward_logfiles.sh
-PB_OUT=/var/log/pb.out
+SCRIPT_FILE="/usr/sbin/forward_logfiles.sh"
+PB_OUT="/var/log/pb.out"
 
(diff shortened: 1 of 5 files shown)


container-host-files/etc/scf/config/scripts/forward_logfiles.sh
```diff
@@ -1,175 +1,171 @@
...
-        echo "*/1 * * * * bash $SCRIPT_FILE >> /dev/null 2>&1" >> tempcrontab
+        sed 's@^\s*@@' >>"${SCRIPT_FILE}" <<-EOF
+                export RSYSLOG_FORWARDER_WATCH_DIR=${RSYSLOG_FORWARDER_WATCH_DIR}
+                export SCF_LOG_HOST=${SCF_LOG_HOST}
+                export SCF_LOG_PORT=${SCF_LOG_PORT}
+                export SCF_LOG_PREFIX=${SCF_LOG_PREFIX}
+                export SCF_LOG_PROTOCOL=${SCF_LOG_PROTOCOL}
+	EOF
+        cat "$0" >> "${SCRIPT_FILE}"
+        echo "*/1 * * * * /usr/bin/env bash ${SCRIPT_FILE} >> /dev/null 2>&1" >> tempcrontab
```
mook-as: The whole "stick ourselves in cron to find new files" thing is kinda ugly. Wish we could rewrite this to use inotify instead.

andreas-kupries: Mini simplification opp: `*/1` means "every minute in steps of 1 minute".
That is the same as `*`, i.e. "every minute".

container-host-files/etc/scf/config/scripts/forward_logfiles.sh
```diff
@@ -1,175 +1,171 @@
...
         # We need to strip leading whitespace introduced by the heredoc (because
         # it doesn't strip leading spaces, just tabs)
-        cat <<-EOF | sed 's@^\s*@@' >${TARGET_NAME}
-            \$InputFileName ${1}
-            \$InputFileTag vcap-${TARGET_BASENAME}
-            \$InputFileStateFile ${TARGET_BASENAME}_state
-            \$InputFileFacility local7
-            \$InputRunFileMonitor
+        sed 's@^\s*@@' >"${TARGET_NAME}" <<-EOF
+                input(type="imfile"
```
andreas-kupries: Is `input(type` proper syntax ?

mook-as: It does seem to work, yes. (The closing parenthesis is a few lines down)

container-host-files/etc/scf/config/scripts/forward_logfiles.sh
```diff
@@ -1,175 +1,171 @@
...
-                                        echo $TARGET_NAME exists
-                                else
-                                        echo "Creating $TARGET_NAME"
-                                        createTargetConf $file
-                                        filesAdded=0
-                                fi
-                        fi
+                        echo "Creating ${TARGET_NAME}"
+                        createTargetConf "${file}"
+                        filesAdded=0
```
andreas-kupries: Is `filesAdded` like `exit 0/1` where `0 = ok/true`, and `1 = fail/false`, i.e. reverse sense ?

container-host-files/etc/scf/config/scripts/forward_logfiles.sh
andreas-kupries: Ah 💡. Completely missed that. Ok then, good enough.
(Although I wonder if indenting the remainder would help to make the association).

container-host-files/etc/scf/config/scripts/forward_logfiles.sh
mook-as: Yes, it is. It's insane, but it was insane before my changes.

andreas-kupries: Yep.

container-host-files/etc/scf/config/scripts/forward_logfiles.sh
```diff
@@ -1,175 +1,171 @@
...
-                                fi
-                        fi
+        local dir file filesAdded=1
+        local args=( "$1" )
+        for dir in ${IGNORE_DIR} ; do
+                args=( "${args[@]}" '(' -name "${dir}" -a -prune ')' -o )
+        done
+        args=( "${args[@]}" '(' -name '*.log' -a -type f -a -print0 ')' )
+        while IFS= read -r -d '' file ; do
+                TARGET_BASENAME="$(basename "$1" .log)"
```
andreas-kupries: The `$1` should apparently be a `$file`. As is the generated file has the same name, regardless of originating logfile, causing only one job to be logged, instead of all.

andreas-kupries: Yes, that change fixes the issue with the missing config files.

PR #1286, opened 2018-01-16
Limit cc-uploader to 1 instance

viovanov:


viovanov:
This wasn't the problem - it was actually mysql and diego-access that were being scaled improperly

PR #1292, opened 2018-01-17
RVRbC8mE Avoid race condition while adding admin buildpacks

jandubois:
There are 2 local worker processes adding buildpacks and Sequel::Plugins::List is not concurrency-safe, so add a lock for updating the `position` value when the new record is created.

### DIFF
--- a/container-host-files/etc/scf/config/role-manifest.yml
+++ b/container-host-files/etc/scf/config/role-manifest.yml
@@ -635,6 +635,7 @@ roles:
   - scripts/nginx_log_level.sh
   scripts:
   - scripts/patches/fix_blobstore_send_timeout.sh
+  - scripts/patches/fix_duplicate_buildpack_positions.sh
   - scripts/patches/fix_nodejs_buildpack.sh
   - scripts/forward_logfiles.sh
   - scripts/patches/use_routing_api_private_endpoint.sh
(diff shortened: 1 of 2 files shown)


mook-as:
Hmm, I _think_ this would let us pass code climate… haven't tested it yet though:
```ruby
        def perform
          logger = Steno.logger('cc.background')
          logger.info "Installing buildpack #{name}"

          created = false
          buildpacks_lock = Locking[name: 'buildpacks']
          buildpacks_lock.db.transaction do
            buildpacks_lock.lock!
            buildpack = Buildpack.find_or_create(name: name) { created = true }
            if buildpack.locked
              logger.info "Buildpack #{name} locked, not updated"
              return
            end
          end

          begin
            buildpack_uploader.upload_buildpack(buildpack, file, File.basename(file))
          rescue => e
            buildpack.destroy if created
            raise e
          end

          buildpack.update(opts)
          logger.info "Buildpack #{name} installed or updated"
        rescue => e
          logger.error("Buildpack #{name} failed to install or update. Error: #{e.inspect}")
          raise e
        end
```

jandubois:
> Hmm, I think this would let us pass code climate… haven't tested it yet though:

I don't want to spend any effort on the nonsense unless the CAPI teams comes back and insists on it. I heard from Eric that they tend to disregard code complexity warnings about their own commits. Either way, a bug fix should not be mixed with refactoring IMO.

Also, we can apply whatever we want in this local patch script; it will be removed once the upstream fix is in place, so let's just keep the code that has already been tested. :)

mook-as:
For future reference, this is upstreamed as cloudfoundry/cloud_controller_ng#1060

PR #1296, opened 2018-01-18
Bump ruby buildpack

jimmykarily:
to fix the non working jruby 2.3.3


aduffeck:
Shouldn't we bump the version of the buildpack so that it's clear if you're using the fixed version or not?

jimmykarily:
After discussing it, we decided to fix this now. I'm closing this PR. I will open a new one with proper versioning.

PR #1308, opened 2018-01-22
Split up the test statuses

aarondl:


Jenkinsfile
```diff
@@ -13,6 +13,31 @@ String jobBaseName() {
...
+        case 'pending':
+            description = 'Tests running'
+            break
+        case 'success':
+            description = 'Tests passed'
+            break
+        case 'failure':
+            description = 'Tests failed'
+            break
+        default:
```
mook-as: This could just be the initial value, but 🤷‍♂️

PR #1334, opened 2018-02-08
Remove diego cell volume

aarondl:


viovanov:
Why keep `grootfs-data` ?

aarondl:
We didn't remove the other volume because it's not necessary, and it seems to scale with the number of apps in the cluster (when pushing `goenv` it adds 1.1GB to the volume) so it seems to be smarter to have it not in ephemeral storage for the node? Not sure. We can remove it later if we want?

PR #1338, opened 2018-02-09
Stop failing builds started by trigger

aarondl:


Jenkinsfile
```diff
@@ -31,11 +31,15 @@ void setBuildStatus(String context, String status) {
...
     }
 
+    try {
     githubNotify credentialsId: 'creds-github-suse-cf-ci-bot',
                  context: "jenkins/${context}",
                  description: description,
                  status: status.toUpperCase(),
                  targetUrl: env.BUILD_URL
+    } catch (IllegalArgumentException e) {
+      echo "Can't notify github status (can't infer git data)"
```
jimmykarily: What is causing this? I thought it had something to do with credentials but in that case we would rather fix it than ignore it.

mook-as: Credentials; we're doing this for now to get builds working again, expecting a follow-up PR to do this correctly.

PR #1341, opened 2018-02-09
sWAKVtIS Bump SSG, UFR for versioned secrets

andreas-kupries:
Ref: https://trello.com/c/sWAKVtIS/551-3-use-versioned-secrets-to-avoid-race-condition
For all related PRs (fissile, scf-secret-generator, uaa-fissile-release, scf)

Note: This needs the fissile PR#323 to work.

The new `make/uaa-upgrade` script is a helper to `helm upgrade` the uaa sub-cluster, like `make/uaa-run` is a helper to start it.


andreas-kupries:
Addressed withs commits 17fe3c49b7ac714c7885bbdcc97081d1189540b0 and baf74982049a385de4633220236f19cd3414662b

andreas-kupries:
Yes. Currently working on this.

make/stop
```diff
@@ -12,5 +12,6 @@ has_namespace() {
 stampy "${GIT_ROOT}/scf_metrics.csv" "$0" make-stop start
 if has_namespace ; then
     kubectl delete namespace "${NAMESPACE}"
+    helm delete $(helm list --namespace  "${NAMESPACE}"|grep -v NAME| awk '{ print $1 }')
```
mook-as: This would be a lot easier if we just added `--name=${NAMESPACE}` to `make/run`.
Otherwise, use `helm list | awk " \$NF == ${NAMESPACE} { print \$1 }"` maybe?

make/tests
```diff
@@ -38,16 +38,28 @@ stampy "${METRICS}" "$0" "make-tests::${POD_NAME}::setup" end
...
+
+remove_test_config () { rm test.yaml ; }
+sed < "output/kube/bosh-task/${POD_NAME}.yaml" \
+    > test.yaml \
+    "s|name: \"secret\"|name: \"${capsecret}\"|"
+trap remove_test_config EXIT
+
 # Delete left-over pod/definition from previous runs, then create/run
-kubectl delete --namespace="${NAMESPACE}" --filename="output/kube/bosh-task/${POD_NAME}.yaml" \
+kubectl delete --namespace="${NAMESPACE}" --filename="test.yaml" \
```
mook-as: We should be able to delete with the old file with the incorrect secrets (that's not used to match what to delete, only namespace/name I think).  Otherwise this wouldn't work if you upgraded CF in the middle...

make/tests
```diff
@@ -38,16 +38,28 @@ stampy "${METRICS}" "$0" "make-tests::${POD_NAME}::setup" end
 
 stampy "${METRICS}" "$0" "make-tests::${POD_NAME}::create" start
 
+# Replace the fixed secret in the relevant task definition with the
+# actual name as pulled from the cluster under test.
+
+capsecret="$(kubectl get pod api-0 --namespace "${NAMESPACE}" -o jsonpath='{@.spec.containers[0].env[?(@.name=="MONIT_PASSWORD")].valueFrom.secretKeyRef.name}')"
+#echo cap-secret = $capsecret
+
+remove_test_config () { rm test.yaml ; }
```
mook-as: Should that use `mktemp`? If the developer happens to be playing with things and has a temporary file named `test.yaml` it'll get deleted...

make/upgrade
```diff
@@ -0,0 +1,64 @@
...
+get_uaa_secret () {
+    kubectl get secret secret-$(release_version "${UAA_NAMESPACE}") \
+	--namespace "${UAA_NAMESPACE}" \
+	-o jsonpath="{.data['$1']}"
+}
+
+# Argument? (**)
+RELEASE=$(helm list --namespace "${NAMESPACE}" | awk '{ print $1 }' | tail -n 1)
+echo Upgrading ${NAMESPACE} release \"${RELEASE}\" ...
+#exit
```
mook-as: Delete line?

make/upgrade
```diff
@@ -0,0 +1,64 @@
...
+
+stampy "${GIT_ROOT}/scf_metrics.csv" "$0" make-run start
+stampy "${GIT_ROOT}/scf_metrics.csv" "$0" make-run::upgrade start
+
+DOMAIN=${DOMAIN:-}
+NETWORK_ENV=bin/settings/network.env
+if test -n "${DOMAIN}"; then
+    TMP=$(mktemp -d)
```
mook-as: Why isn't it just a file? Why make it a directory?

I _think_ if we tried harder we could make this not have a temporary file.

make/upgrade
```diff
@@ -0,0 +1,64 @@
...
+
+helm upgrade ${RELEASE} output/helm \
+     --namespace ${NAMESPACE} \
+     --set "env.CLUSTER_ADMIN_PASSWORD=$CLUSTER_ADMIN_PASSWORD" \
+     --set "env.DOMAIN=${DOMAIN}" \
+     --set "env.TCP_DOMAIN=${TCP_DOMAIN}" \
+     --set "env.UAA_ADMIN_CLIENT_SECRET=${UAA_ADMIN_CLIENT_SECRET}" \
+     --set "env.UAA_HOST=${UAA_HOST}" \
+     --set "env.UAA_PORT=${UAA_PORT}" \
+    --set "env.UAA_CA_CERT=${UAA_CA_CERT}" \
```
mook-as: Indent looks off

andreas-kupries:
SSG now hs https://github.com/SUSE/scf-secret-generator/pull/12

make/upgrade
andreas-kupries: It is actually the one line which had the indent correct.
Fixed the others with commit 207594b6c66964441319db433f170d5fbba7a046

make/upgrade
andreas-kupries: This part of the script is inherited from `make/run`. I saw no reason to fiddle with parts which worked.
Not changing.

make/upgrade
andreas-kupries: Yes. Done, commit 4e77b5e4929e24b557084a8e28dae24636069a58

make/tests
andreas-kupries: Yes, will fix.

make/tests
andreas-kupries: Wait, you are talking about running an upgrade while the test are running ?
Is that something we have ever supported ?

make/stop
andreas-kupries: You mean in the Makefile ?
I actually can simplify this now that I know about `-q` for helm list.

andreas-kupries: Done with commit 25e5e602dcd288ff89137c946d06066bba931000

make/tests
andreas-kupries: Done with commit 619e92637782a7acd2566211ff3fec2cb0f238b3

make/tests
```diff
@@ -47,7 +47,19 @@ while has_pod ; do
     sleep 1
 done
 
-kubectl create --namespace="${NAMESPACE}" --filename="output/kube/bosh-task/${POD_NAME}.yaml"
+# Replace the fixed secret in the relevant task definition with the
+# actual name as pulled from the cluster under test.
+
+capsecret="$(kubectl get pod api-0 --namespace "${NAMESPACE}" -o jsonpath='{@.spec.containers[0].env[?(@.name=="MONIT_PASSWORD")].valueFrom.secretKeyRef.name}')"
+#echo cap-secret = $capsecret
```
mook-as: Remove?

make/tests
```diff
@@ -47,7 +47,19 @@ while has_pod ; do
...
+
+CONFIG=$(mktemp)
+sed < "output/kube/bosh-task/${POD_NAME}.yaml" \
+    > "${CONFIG}" \
+    "s|name: \"secret\"|name: \"${capsecret}\"|"
+
+kubectl create --namespace="${NAMESPACE}" --filename="${CONFIG}"
+rm "${CONFIG}"
```
mook-as: We have `errexit` in effect here; would that not have a chance to not delete the temp file?

make/tests
andreas-kupries: Ok

make/tests
andreas-kupries: Oh, so you think we still need to trap ?

make/tests
andreas-kupries: Done with commit a48233d7d1a27f867a07a4cd138bd4248e05cd43

make/tests
andreas-kupries: Re-added with commit a48233d7d1a27f867a07a4cd138bd4248e05cd43

PR #1342, opened 2018-02-09
Update for indexed roles

alexjh:
This is an intermediate step until the new fissile stemcells are generated. New 'clustered' roles will be added once they are ready.


alexjh:
Deleting as this intermediate step isn't required.

PR #1365, opened 2018-02-23
Allow more than one IP address for external IPs

mook-as:
This updates SCF itself to use multiple IPs (one of which is not real and is just there to test the change).

### DIFF
--- a/make/run
+++ b/make/run
@@ -83,7 +83,8 @@ helm install output/helm \
      --set "env.UAA_HOST=${UAA_HOST}" \
      --set "env.UAA_PORT=${UAA_PORT}" \
      --set "env.UAA_CA_CERT=${CA_CERT}" \
-     --set "kube.external_ip=$(dig +short ${DOMAIN})" \
+     --set "kube.external_ips[0]=$(dig +short ${DOMAIN})" \
+     --set "kube.external_ips[1]=192.0.2.42" \
      "$@"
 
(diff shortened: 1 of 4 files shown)


alexjh:
I think https://github.com/SUSE/scf/blob/4deef677e6d18bac0839344abd99e7d0301b8f75/Jenkinsfile#L473 needs updating as well.

And UAA startup scripts should change too.

mook-as:
I'm deliberately _not_ changing UAA to ensure that the fallback path works. But yes, Jenkinsfile should change.

make/run
```diff
@@ -83,7 +83,8 @@ helm install output/helm \
      --set "env.UAA_HOST=${UAA_HOST}" \
      --set "env.UAA_PORT=${UAA_PORT}" \
      --set "env.UAA_CA_CERT=${CA_CERT}" \
-     --set "kube.external_ip=$(dig +short ${DOMAIN})" \
+     --set "kube.external_ips[0]=$(dig +short ${DOMAIN})" \
+     --set "kube.external_ips[1]=192.0.2.42" \
```
jandubois: If this is just for testing, why does it have to be checked into the repo and not just added as a comment on the Trello card?

If this is a permanent addition, then it should have a comment in the script and not just the commit message.

mook-as: This is a permanent change (to ensure that we don't regress). But I'll add comments to both places.

mook-as: Fixed in  1ccea81

PR #1380, opened 2018-03-05
Memory requests based on measuring max me of the roles in a long-running vagrant.

andreas-kupries:
Ref: https://trello.com/c/UEvqSutH/576-sane-memory-limits-for-all-pods

The requests are `max times 3`.
Note: While `fissile` transforms the request into data in values.yml it does not generate clauses using them yet.

You need https://github.com/SUSE/fissile/pull/329


alexjh:
I'm not sure if this is a new event or related to the changes here, but I'm seeing this message. I haven't tested on a `develop` branch to confirm. @andreas-kupries, do you know if it's a concern?

```
vagrant@vagrant-kube:~/scf> k get events
kubectl get events
LASTSEEN   FIRSTSEEN   COUNT     NAME        KIND      SUBOBJECT   TYPE      REASON                             SOURCE                MESSAGE
13s        3h          226       127.0.0.1   Node                  Warning   FailedNodeAllocatableEnforcement   {kubelet 127.0.0.1}   Failed to update Node Allocatable Limits "": failed to set supported cgroup subsystems for cgroup : Failed to set config for supported subsystems : failed to write 16826359808 to memory.limit_in_bytes: write /sys/fs/cgroup/memory/memory.limit_in_bytes: invalid argument
```

andreas-kupries:
I believe I may have seen this message as well.

I can imagine that with us going to actually use memory requests and limits that we are triggering code paths in kube which we did not before.

I do not know if the event is an issue or not.

Googling for a part of the message I find

https://github.com/kubernetes/kubernetes/issues/55867

PR #1397, opened 2018-03-15
Add anti affinity to new syslog roles

alexjh:
All roles should have anti affinity with themselves.


alexjh:
There's a ticket in Trello for this default... it's sitting in `Not Sure` right now. :)

jandubois:
> There's a ticket in Trello for this default... it's sitting in Not Sure right now. :)

Has been moved by @troytop. I assume because this is not a customer-facing change, but an internal simplification of the rolemanifest? Maybe let Troy chime in...

PR #1401, opened 2018-03-20
Bump fissile version for node affinity support

alexjh:


andreas-kupries:
Note, this also pulls in the fix `marky/no-nil-limits`.

PR #1402, opened 2018-03-20
Redo how secrets are generated

jandubois:


Jenkinsfile
```diff
@@ -63,7 +63,9 @@ void runTest(String testName) {
                         value = "tcp.#{domain}" if env['name'] == 'TCP_DOMAIN'
                         env['value'] = value.to_s
 
-                        env['valueFrom']['secretKeyRef']['name'] = capsecret if env['valueFrom'] && env['valueFrom']['secretKeyRef']
+                        if env['name'] != 'CLUSTER_ADMIN_PASSWORD' && env['name'] != 'UAA_ADMIN_CLIENT_SECRET' && env['name'] != 'UAA_CA_CERT'
+                            env['valueFrom']['secretKeyRef']['name'] = capsecret if env['valueFrom'] && env['valueFrom']['secretKeyRef']
```
mook-as: Having a `foo if bar` inside a `if …` seems kind of weird. (But I would have also done something like `unless %w(CLUSTER_ADMIN_PASSWORD UAA_ADMIN_CLIENT_SECRET UAA_CA_CERT).include? env['name']` or a `case` statement)

bin/dev/install_tools.sh
```diff
@@ -7,7 +7,7 @@ set -vx
 SCF_BIN_DIR="${SCF_BIN_DIR:-output/bin}"
 
 # Tool locations
-s3="https://cf-opensusefs2.s3.amazonaws.com/fissile"
+s3="https://cf-opensusefs2.s3.amazonaws.com/fissile/checks"
```
mook-as: (Reminder to not merge with the `checks` in place)

Jenkinsfile
jandubois: Addressed in https://github.com/SUSE/scf/commit/c6581db26701a2b0e77503b74c7f25501f3bbbde

PR #1405, opened 2018-03-21
Fix make kube source env file issue for helm

zhangtbj:
Right now, `make kube` only source FISSILE_DEFAULTS_FILE for kube, but when we use `make kube helm`, there is no any source file. 

So I move `FISSILE_DEFAULTS_FILE="${FISSILE_DEFAULTS_FILE},bin/settings/settings.env,${NETWORK_ENV}"`
out of `if [ "${BUILD_TARGET}" = "kube" ]; then`

### DIFF
--- a/make/kube
+++ b/make/kube
@@ -37,9 +37,9 @@ fi
 if [ "${BUILD_TARGET}" = "kube" ]; then
     # Overrides when generating kube config files instead of helm charts.
     FISSILE_OUTPUT_DIR="${PWD}/output/kube"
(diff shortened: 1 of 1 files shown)


zhangtbj:
For example, I change to use other blobstore type such as aws. I need some additional info. And I didn't set default value in role-manifest.yml for these info, then it will report error during make validate said these info are empty. I would like to use settings.env to add these then render them as before. I remember SCF supports this for `helm` before.

PR #1415, opened 2018-03-24
Make sure docker login always specifies the FISSILE_DOCKER_REGISTRY

jandubois:


jandubois:
@mook-as said on 🚀 chat:
> And yes that change looks sane

Which is the best I can get for code review right now.

PR #1425, opened 2018-03-28
Admin password is in secrets now

alexjh:


jandubois:
> However, I question the practice of echoing credentials to STDOUT and would like to see confirmation from @ttopnik that this is really what we want to do.

That should have been @troytop

alexjh:
Removed references to admin creds instead.

Issue #1433, opened 2018-04-02
Timestamp failed of rsyslog_file in some component


zhanggbj:
In some cf component, like nats or doppler etc, there is a `Timestamp failed` of rsyslog_file if you do `monit summary` after the pod is started for a while.

```
nats/1:/# monit summary
File 'rsyslog_file'                 Timestamp failed
```

The error is caused by file /var/log/message which has not been updated for more than 65 minutes, according to the monit definition, if timestamp is greater than 65 minutes, there will be an alert.
/var/log/message is used for general syslog messages, there isn't so much log as VM do. This config `/etc/monit/monitrc.d/rsyslog` is baked in stemcell .
```
check file rsyslog_file with path /var/log/messages
   group rsyslogd
   if timestamp > 65 minutes then alert
```

zhanggbj:
I'll close it as the code change has been merged. Thanks

PR #1434, opened 2018-04-02
Add a patch to remove timestamp check for rsyslog

zhanggbj:
Hi,
In nats or doppler, only limited logs will be recorded to /var/log/messages, then this file is no longer updated. If /var/log/messages is outdated for more than 65 mins, there will be an alert named "timestamp failed". So add a patch for these component to remove the check.

[issue 1433](https://github.com/SUSE/scf/issues/1433)
Thanks!

### DIFF
--- a/container-host-files/etc/scf/config/role-manifest.yml
+++ b/container-host-files/etc/scf/config/role-manifest.yml
@@ -5,6 +5,7 @@ roles:
   - scripts/configure-HA-hosts.sh
   scripts:
   - scripts/forward_logfiles.sh
+  - scripts/patches/fix_monit_rsyslog.sh
   jobs:
   - name: global-properties # needs to be first so images use it for processing monit templates
     release_name: scf-helper
@@ -37,6 +38,7 @@ roles:
   - scripts/configure-HA-hosts.sh
   scripts:
   - scripts/forward_logfiles.sh
+  - scripts/patches/fix_monit_rsyslog.sh
   jobs:
   - name: global-properties # needs to be first so images use it for processing monit templates
     release_name: scf-helper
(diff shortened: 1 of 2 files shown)


jandubois:
> script scripts/patches/fix_monit_syslog.sh is not used in the role manifest

Yes, because the inside the role manifest it is called `rsyslog` and not `syslog`.

jandubois:
> I'd like a sanity check from @jandubois, but I think this is fine?

Once the patch file is renamed to match the role manifest name, I'm fine with the PR.

zhanggbj:
@mook-as @jandubois  Thanks for your review! I've fixed the path file name.
Jenkins is still failing but I can't see the detials result `zhanggbj is missing the Overall/Read permission`.
Is there any way I can see it?

zhanggbj:
Just checked the Jenkins log, it seems that there's not enough storage.

```
Stack trace
java.io.IOException: No space left on device
```

jandubois:
> I've manually triggered another build for this PR to see if it works now.

That build has now failed as well, but for a different reason:

For security reasons Jenkins will not use the `Jenkinsfile` from the PR branch unless the PR author has *write* access to the base repo. If they don't, then Jenkins will use the `Jenkinsfile` from the HEAD of the branch the PR is to be merged into.

Now there has been some refactoring in the `Jenkinsfile` that relies on an external file `make/include/secrets` that doesn't exist in your PR branch yet.

Could you please rebase your PR on top of HEAD of `develop` and do a `git push --force` of your PR? That should also automatically trigger another Jenkins build.

Sorry for this, but I don't see an easier way of getting your PR to build on this Jenkins setup right now.

zhanggbj:
@jandubois Thanks for your help and info!
I just rebased it and now it should be updated with `develop` branch 9f565811c70c42175c4993eba2ff50dc5e14d5fc.

PR #1435, opened 2018-04-02
Update upgrade script for the new layout of secrets.

andreas-kupries:


make/upgrade
```diff
@@ -8,16 +8,24 @@ set -o errexit -o nounset
...
 NAMESPACE="cf"
 UAA_NAMESPACE="uaa"
-UAA_CA_CERT=internal-ca-cert
 
 GIT_ROOT=${GIT_ROOT:-$(git rev-parse --show-toplevel)}
 cd "${GIT_ROOT}"
 
-# Determine the kube revision of the chart controlling the specified namespace.
-release_version() { helm list --date --reverse --max 1 --namespace "$1" | awk '{ print $2 }' | tail -n 1 ; }
+get_uaa_secret_name() {
```
jandubois: I think this function should take an argument for `INTERNAL_CA_CERT` and not have it hard-coded.

make/upgrade
```diff
@@ -8,16 +8,24 @@ set -o errexit -o nounset
...
+    kubectl get pods \
+	--namespace "${UAA_NAMESPACE}" \
+	-o jsonpath='{.items[*].spec.containers[?(.name=="uaa")].env[?(.name=="INTERNAL_CA_CERT")].valueFrom.secretKeyRef.name}'
+}
+
+get_uaa_secret_key() {
```
jandubois: Same here

make/upgrade
```diff
@@ -8,16 +8,24 @@ set -o errexit -o nounset
...
+    kubectl get pods \
+	--namespace "${UAA_NAMESPACE}" \
+	-o jsonpath='{.items[*].spec.containers[?(.name=="uaa")].env[?(.name=="INTERNAL_CA_CERT")].valueFrom.secretKeyRef.key}'
+}
 
 get_uaa_secret () {
```
jandubois: This function also should take the env variable name as an argument, and not the name of the secret key. Instead it should call `$(get_uaa_secret_key $ENV_NAME)` itself.

make/upgrade
```diff
@@ -8,16 +8,24 @@ set -o errexit -o nounset
...
-    kubectl get secret secret-$(release_version "${UAA_NAMESPACE}") \
+    kubectl get secret $(get_uaa_secret_name) \
```
jandubois: This call should move inside `get_uaa_secret`.

make/upgrade
```diff
@@ -44,7 +52,7 @@ fi
 source bin/settings/settings.env
 source "${NETWORK_ENV}"
 
-UAA_CA_CERT="$(get_uaa_secret ${UAA_CA_CERT} | base64 -d -)"
+UAA_CA_CERT="$(get_uaa_secret $(get_uaa_secret_key) | base64 -d -)"
```
jandubois: The argument here should be `INTERNAL_CA_CERT`

make/upgrade
jandubois: Hmm, this comment looks like it got attached to the wrong line

make/upgrade
andreas-kupries: Done with commit 405a6bdb262d41482dfb7d8e5045cf3f406aa183

make/upgrade
andreas-kupries: Done with commit 405a6bdb262d41482dfb7d8e5045cf3f406aa183

make/upgrade
andreas-kupries: Done with commit 405a6bdb262d41482dfb7d8e5045cf3f406aa183

make/upgrade
andreas-kupries: Done with commit 405a6bdb262d41482dfb7d8e5045cf3f406aa183

make/upgrade
andreas-kupries: Done with commit 405a6bdb262d41482dfb7d8e5045cf3f406aa183

PR #1448, opened 2018-04-10
Reduce the number of immutable secrets

mook-as:
We can have fewer immutable secrets (if we upgrade cf-mysql); with just 4 left we can still run an upgrade (with a new secret generation number) and still have the cluster settle to a usable state.

UAA changes are still pending; it is not yet obvious how we'll be able to carry the UAA changes to SCF.

### DIFF
--- a/container-host-files/etc/scf/config/scripts/patches/fix_mysql_advertise_ip.sh
+++ b/container-host-files/etc/scf/config/scripts/patches/fix_mysql_advertise_ip.sh
@@ -9,17 +9,17 @@ fi
 
 cd "$PATCH_DIR"
 
-patch --force -p3 <<'PATCH'
-diff --git jobs/mysql/templates/my.cnf.erb jobs/mysql/templates/my.cnf.erb
-index 73645618..7d94807e 100644
---- jobs/mysql/templates/my.cnf.erb
-+++ jobs/mysql/templates/my.cnf.erb
-@@ -49,13 +49,14 @@ nice                            = 0
+patch --force -p4 <<'PATCH'
+diff --git a/jobs/mysql/templates/my.cnf.erb b/jobs/mysql/templates/my.cnf.erb
+index 514acdd5..80bb6b0e 100644
+--- a/jobs/mysql/templates/my.cnf.erb
++++ b/jobs/mysql/templates/my.cnf.erb
+@@ -58,13 +58,14 @@ nice                            = 0
  # GALERA options:
  wsrep_on                        = ON
  wsrep_provider                  = /var/vcap/packages/mariadb/lib/plugin/libgalera_smm.so
--wsrep_provider_options          = "gcache.size=<%= p('cf_mysql.mysql.gcache_size') %>M;pc.recovery=TRUE;pc.checksum=TRUE"
-+wsrep_provider_options          = "gcache.size=<%= p('cf_mysql.mysql.gcache_size') %>M;pc.recovery=TRUE;pc.checksum=TRUE;ist.recv_addr=<%= discover_external_ip %>:4568"
+-wsrep_provider_options          = "gcache.size=<%= p('cf_mysql.mysql.gcache_size') %>M;pc.recovery=FALSE;pc.checksum=TRUE"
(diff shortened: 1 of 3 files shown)


container-host-files/etc/scf/config/scripts/patches/fix_mysql_advertise_ip.sh
```diff
@@ -9,16 +9,16 @@ fi
...
+diff --git a/jobs/mysql/templates/my.cnf.erb b/jobs/mysql/templates/my.cnf.erb
+index 514acdd5..80bb6b0e 100644
+--- a/jobs/mysql/templates/my.cnf.erb
++++ b/jobs/mysql/templates/my.cnf.erb
+@@ -58,13 +58,14 @@ nice                            = 0
  # GALERA options:
  wsrep_on                        = ON
  wsrep_provider                  = /var/vcap/packages/mariadb/lib/plugin/libgalera_smm.so
--wsrep_provider_options          = "gcache.size=<%= p('cf_mysql.mysql.gcache_size') %>M;pc.recovery=TRUE;pc.checksum=TRUE"
+-wsrep_provider_options          = "gcache.size=<%= p('cf_mysql.mysql.gcache_size') %>M;pc.recovery=FALSE;pc.checksum=TRUE"
```
jandubois: How is this change related to making secrets mutable?

jandubois: Oops, sorry, I guess this was the removed line, so it is an upstream change. But shouldn't we keep the change to `pc.recovery` in our replacement line instead of setting it back to `TRUE`?

mook-as: That's needed to apply the patch due to changes in cf-mysql-release (which we needed to upgrade to pick up resetting passwords).

Maybe? I'm not sure what we might use that for.

mook-as: Looking at the relevant upstream change: https://github.com/cloudfoundry/cf-mysql-release/commit/8056e2939fac2f5e8ca42de0a50e821ea45c2f6a
I thought @alexjh did some work to deal with recovery in similar circumstances…

jandubois: > Looking at the relevant upstream change: cloudfoundry/cf-mysql-release@8056e29

Yes, that upstream commit is changing the value for `pc.recovery` from `TRUE` to `FALSE`, and you adjusted our patch to match.

But you did not adjust the replacement line as well, so our patch is now reverting this upstream change.  I suspect this is unintentional, and you just need to update your patch so that the replacement line is:

```
+wsrep_provider_options          = "gcache.size=<%= p('cf_mysql.mysql.gcache_size') %>M;pc.recovery=FALSE;pc.checksum=TRUE;ist.recv_addr=<%= discover_external_ip %>:4568"
```

instead of

```
+wsrep_provider_options          = "gcache.size=<%= p('cf_mysql.mysql.gcache_size') %>M;pc.recovery=TRUE;pc.checksum=TRUE;ist.recv_addr=<%= discover_external_ip %>:4568"
```

But if you do indeed want to revert that upstream change, then it needs to be explained in the commit message why you are reverting `pc.recovery` from the upstream setting back to `TRUE`.

alexjh: Yes, we need to keep the `ist.recv_addr=<%= discover_external_ip %>:4568` because the default gets the IP from the wrong interface, so it's not reachable from the other nodes. It would never bootstrap.

Their change sounds like a fix for a cluster that's already been bootstrapped, so I don't think they'll affect each other.

mook-as: Fixed in 576b6c6b0a28a05f21be3ba97a2be17a896db167

mook-as:
This is now blocked by https://github.com/SUSE/fissile/pull/349 (to pass on Jenkins, which has kube 1.7).

mook-as:
Rebased; since UAA is tip-of-develop now, I ended up not bumping it here.

PR #1456, opened 2018-04-13
Bump fissile to version required explicit tagging of roles for statefulset

andreas-kupries:
__Attention__: Currently using check/ version. Fix when build becomes master.
Modified role manifest to tag blobstore and diego-cell properly.

Depends on: https://github.com/SUSE/fissile/pull/347

https://trello.com/c/0zylWw07/606-2-fissile-statefulsets-should-be-explicitly-defined


container-host-files/etc/scf/config/role-manifest.yml
```diff
@@ -755,6 +755,7 @@ roles:
   - name: blobstore_url_signer
   - name: route_registrar
   - name: metron_agent
+  tags: [indexed]
```
mook-as: Same as UAA, we should have a comment here on why it's indexed. Also the other tag.

andreas-kupries: I am willing to add the note. Unfortunately I do not know the _why_ myself.
I will ping Vlad.

andreas-kupries: Addressed with commit f6660b9523f40d0308d0af11ae0ac08a8f4decd8

container-host-files/etc/scf/config/role-manifest.yml
```diff
@@ -755,6 +755,9 @@ roles:
   - name: blobstore_url_signer
   - name: route_registrar
   - name: metron_agent
+  tags:
+  - indexed
+  # Role needs LB => indexed
```
mook-as: The comment should be describing _why_ it needs an LB.

andreas-kupries: Added info in commit 5ece0bad56d95b4351fcf8c95cf559af9d89fcfa

PR #1479, opened 2018-04-25
Remove architecture tag from release archive

thardeck:
because it only provides helm charts which are hardware independent.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -377,7 +377,7 @@ pipeline {
                     source ${PWD}/.envrc
                     set -x
                     unset SCF_PACKAGE_COMPILATION_CACHE
-                    rm -f output/scf-*amd64*.zip
+                    rm -f output/scf-*.zip
                     make helm bundle-dist
                 '''
             }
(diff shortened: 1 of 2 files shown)


Jenkinsfile
```diff
@@ -393,22 +398,25 @@ pipeline {
...
+                    fi
+
                     kubectl delete storageclass hostpath || /bin/true
                     kubectl create -f - <<< '{"kind":"StorageClass","apiVersion":"storage.k8s.io/v1","metadata":{"name":"hostpath"},"provisioner":"kubernetes.io/host-path"}'
 
                     # Unzip the bundle
                     rm -rf output/unzipped
                     mkdir -p output/unzipped
-                    unzip -e output/scf-*amd64*.zip -d output/unzipped
+                    unzip -e output/scf-$OS-*.zip -d output/unzipped
```
mook-as: This is in a `"""` block, so `${}` is being interpreted by Jenkins instead of `sh`. You need `\${OS}` here.

thardeck: THX. Ok, that explains why the params. variable worked in one case and not the other.

PR #1490, opened 2018-04-26
Tell make/uaa-images about DOCKER_APP_VERSION

mook-as:
Fixes bustage from #1478


andreas-kupries:
While it fails in Jenkins,  it apparently failed (per Mark) because https://github.com/SUSE/scf/pull/1456 hasn't landed yet. So, merge this, then rebase 1456, and review/merge that.

PR #1493, opened 2018-04-27
Add a step to Jenkins to trigger secret rotation

alexjh:
Add a dummy `--set` to helm upgrade to trigger the rotation.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -128,6 +128,11 @@ pipeline {
             description: 'Enable publishing to amazon s3',
         )
         booleanParam(
+            name: 'TEST_ROTATE',
+            defaultValue: true,
+            description: 'Trigger secret rotation via helm upgrade',
+        )
+        booleanParam(
             name: 'TEST_SMOKE',
             defaultValue: true,
             description: 'Run smoke tests',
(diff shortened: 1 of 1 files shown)


jandubois:
@AlHarford The corresponding Trello card is in `Code Review`, but this PR still has the WIP label. Can you please either remove the WIP label or move the card back to `In Progress`?

mook-as:
@jandubois Sorry about that, label removed. (I took over from Alex at some point.)

jandubois:
@mook-as The PR looks good; I'm fine with merging it.

However, when I looked through the Jenkins output I was wondering if we shouldn't log the value of one of the secrets before and after rotation, just so you can verify from the log alone that the secrets actually have been rotated. Maybe even make the rotation step fail when the secret is still the same after rotation? Right now we rely on the specific parameters to helm to trigger the rotation, but we would never notice if they turned into a no-op in the future.

What do you think?

mook-as:
I'd be worried about keeping a _specific_ secret, because that secret might end up going away later. I suppose we could attempt to keep the first non-immutable generated secret, but then we'd have to figure out a way to find that…

jandubois:
> I'd be worried about keeping a specific secret, because that secret might end up going away later.

Why is that a concern? It means either the fetching of the no-longer-existing secret will fail, or it will return a constant value (empty string?), which will be unchanged after rotation and throw an error.

So either way this will throw an error as soon as the secret is removed, at which point we can update the `Jenkinsfile` to use a different one.

We already have an obvious choice for UAAA, as we already fetch the `internal-ca-cert` anyways, so it makes sense to check that it has changed value after rotation.

Just pick something for SCF that you expect won't go away for a while. Or pick something randomly. Or use `blobstore-password` and be done with it. 😄

Jenkinsfile
```diff
@@ -491,6 +491,14 @@ pipeline {
 
                     . make/include/secrets
 
+                    # Get the last updated secret
+                    secret_resource="\$(kubectl get secrets --namespace="${jobBaseName()}-${BUILD_NUMBER}-scf" --output=jsonpath='{.items[-1:].metadata.name}' --sort-by=.metadata.resourceVersion)"
+
+                    # Get a random secret
+                    secret_name="\$(kubectl get secret --namespace="${jobBaseName()}-${BUILD_NUMBER}-scf" "\${secret_resource}" -o json | jq -r '.data | to_entries | .[].key | select(contains(".generator") | not) ' | sort --random-sort | head -n 1)"
```
jandubois: I don't think this will work because you don't exclude immutable secrets (which I don't think is possible without parsing the role manifest again).

Issue #1492, opened 2018-04-27
Statefulset nodes fail to start on Kubernetes 1.9


HeavyWombat:
Hey there, @qu1queee and I investigated an issue colleagues of us had with a version of SCF we have deployed. In that cluster,  `nats` is a statefulset with two nodes. The first one came up just fine, the second one failed to start with some weird error messages.

We debugged it so far that we think that [configure-HA-hosts.sh](https://github.com/SUSE/scf/blob/2.8.0/container-host-files/etc/scf/config/scripts/configure-HA-hosts.sh#L44) relies on a certain structure of Kubernetes API result JSON that changes between version 1.8 and 1.9 of Kubernetes.

In 1.8, the JSON result has `kubernetes.io/created-by`:
```json
{
  "checksum/config": "[...]",
  "kubernetes.io/created-by": "{[...],\"apiVersion\":\"v1\",\"reference\":{\"kind\":\"StatefulSet\",\"name\":\"nats\", [...]}"
}
```

From what we can see, our 1.9 Kubernetes cluster API result does not have anything like that. There seems to be a new section with details about the name that could be used.

We wanted to check with you if you encountered a similar issue, too? Thanks in advance for your feedback.

@zhangtbj @zhanggbj @ScarletTanager FYI

Flask:
@HeavyWombat That feature was deprecated in Kubernetes 1.8 and removed in 1.9.

You still can edit the statefulset for mysql and nats manually to add the annotation `kubernetes.io/created-by: '{"reference":{"name":"mysql"}}'` (nats for the other statefulset. 

Another issue we ran into was the configgin-role for scf, where the "list" verb was missing.

Besides those two things it seems to run quite nice on 1.9.6

Cheers

Xiaohua-Shen:
another way to fix this is using `.metadata.ownerReferences[0].name`, which have same output for 1.8 and 1.9.  

```
# kubectl get pod nats-0 -n cf -o json | jq .metadata.ownerReferences[0].name
"nats"
# kubectl get pod nats-1 -n cf -o json | jq .metadata.ownerReferences
[
  {
    "apiVersion": "apps/v1beta1",
    "blockOwnerDeletion": true,
    "controller": true,
    "kind": "StatefulSet",
    "name": "nats",
    "uid": "6b779637-495f-11e8-8381-16336b7583f8"
  }
]
```

alexjh:
@HeavyWombat @Xiaohua-Shen ,

[...]

PR #1494, opened 2018-04-28
fix scripts to support both k8s v1.8 and v1.9

Xiaohua-Shen:
it's try to fix the issue https://github.com/SUSE/scf/issues/1492

### DIFF
--- a/container-host-files/etc/scf/config/scripts/configure-HA-hosts.sh
+++ b/container-host-files/etc/scf/config/scripts/configure-HA-hosts.sh
@@ -45,7 +45,7 @@ find_cluster_ha_hosts() {
         # Find the number of replicas we have
         local statefulset_name replicas i
         for ((i = 0 ; i < 5 ; i ++)) ; do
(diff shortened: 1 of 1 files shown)


mook-as:
This is a dumb question (doubly dumb because I wrote the original, terrible code): Why are we not using `statefulset_name="$(k8s_api api/v1 "/pods/${HOSTNAME}" | json_get '["metadata"]["labels"]["skiff-role-name"]')"`?

I was dumb and didn't explain in b18c830490c1b5b546b32c8c2c4d74b2b0d6653c from #1016…

Xiaohua-Shen:
@mook-as yes,  `statefulset_name="$(k8s_api api/v1 "/pods/${HOSTNAME}" | json_get '["metadata"]["labels"]["skiff-role-name"]')"` seems a better choice, since there is no array list.    My original PR is just based on existing logic to try to find the statefulset's name, not aware the sikff-role-name is same for the statefulset name.

I will update the pr based on it.

PR #1495, opened 2018-04-30
Add cf-networking-release

mook-as:
Attempting to get rid of the CATS failure


edwardstudy:
@mook-as Does scf support cf-networking function currently? It is nice performance enhancement feature we want to.

PR #1512, opened 2018-05-07
Fix string interpolation in Jenkinsfile

aduffeck:
With single quotes the os part of the glob is not interpolated resulting
in an empty file array and a subsequent build failure.


thardeck:
@jeff-hobbs I think the issue is that `tar_sources` runs a `make compile-clean` which removes `output`. Is this really necessary because `vagrant-prep` then is run a second time?

So we can either change the upload position of s3 or get rid of the `compile-clean` for tar_sources.

PR #1518, opened 2018-05-07
Jenkinsfile: troubelshooting why our zip files are not uploading

mook-as:
Not for merging

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -392,15 +392,11 @@ pipeline {
             when {
                 expression { return false }
             }
-            steps {
-            }
         }
 
         stage('smoke') {
             when {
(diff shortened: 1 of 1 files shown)


thardeck:
I think the issue is that tar_sources runs a `make compile-clean` which removes output. Is this really necessary because vagrant-prep then is run a second time.

So we can either change the upload time to s3 or get rid of the compile-clean for tar sources.

mook-as:
Yep, it looks like you're right: putting the `tar_sources` back in makes `publish_s3` not find any files. (It didn't fail because I didn't try to do anything the files, but it did print an empty list.)

Closing, the fix is in #1521

PR #1521, opened 2018-05-08
Jenkins: Don't run `make clean` when tarring up sources

mook-as:
That kinda kills the output


mook-as:
I branched this off `master`, if we want to cherry pick it into that. @gaktive gets to decide, I guess :D

Issue #1528, opened 2018-05-10
Failed to get secret for multi uaa pods


edwardstudy:
Hi, 

Function failed to get secret when I provisioned multi uaa pods. This [command](https://github.com/SUSE/scf/blob/develop/make/include/secrets#L17-L19) failed:
```
Error from server (NotFound): secrets "secrets-270.28.0-1 secrets-270.28.0-1" not found
Error from server (NotFound): secrets "secrets-270.28.0-1 secrets-270.28.0-1" not found
Error from server (NotFound): secrets "secrets-270.28.0-1 secrets-270.28.0-1" not found
Error from server (NotFound): secrets "secrets-270.28.0-1 secrets-270.28.0-1" not found
Error from server (NotFound): secrets "secrets-270.28.0-1 secrets-270.28.0-1" not found
```

Because the method to get secret name/key received duplicate value:
```
bjxzi$ kubectl get pods --namespace uaa
NAME                        READY     STATUS      RESTARTS   AGE
postgres-0                  1/1       Running     0          18h
secret-generation-1-tmbg5   0/1       Completed   0          18h
uaa-77448b8688-m89fb        1/1       Running     0          18h
uaa-77448b8688-nwzct        1/1       Running     0          18h

bjxzi$ kubectl get pods --namespace uaa -o 'jsonpath={.items[*].spec.containers[?(.name=="uaa")].env[?(.name=="INTERNAL_CA_CERT")].valueFrom.secretKeyRef.name}'
secrets-270.28.0-1 secrets-270.28.0-1
```

mook-as:
Looks like #1529 was merged, and that fixes this; thanks for your change!

PR #1529, opened 2018-05-10
Fetch secret name/key from first matched pod

edwardstudy:
This PR fixed issue [1528](https://github.com/SUSE/scf/issues/1528).


jandubois:
Thank you for your patch!

PR #1531, opened 2018-05-10
Restore scf specific jobs

alexjh:


src/scf-release/jobs/acceptance-tests/spec
```diff
@@ -0,0 +1,130 @@
...
+  acceptance_tests.include_detect:
+    description: Flag to run tests in the detect suite
+  acceptance_tests.use_http:
+    description: Flag for using HTTP when making api and application requests rather than the default HTTPS
+  acceptance_tests.include_privileged_container_support:
+    description: Flag for running tests that depend on privileged container support being enabled in your Cloud Foundry.
+  acceptance_tests.include_container_networking:
+    description: Flag for running tests that depend on container networking support being enabled in your Cloud Foundry.
+  acceptance_tests.include_isolation_segments:
+    description: Flag for running tests that depend on having isolation segments in your Cloud Foundry.
```
mook-as: Huh, I don't think we've been keeping up on the `include` flags. We're missing [`include_routing_isolation_segments`](https://github.com/cloudfoundry/cf-acceptance-tests/blob/cf1.21/helpers/config/config_struct.go#L94), for example.

Not in this PR, though; made [Trello card](https://trello.com/c/VQNQMVnl/) instead.

PR #1550, opened 2018-05-22
Bump scf-secrets-generator to make cert expiration configurable

jandubois:
Update role manifest validation to let the secrets generator use
KUBE_SERVICE_DOMAIN_SUFFIX without configuring HA itself.

[trello#PGGWWyht]

### DIFF
--- a/bin/config-validator.rb
+++ b/bin/config-validator.rb
@@ -301,6 +301,8 @@ def check_clustering(manifest, bosh_properties)
       @has_errors += 1
     else
       next if has_script(role, @configure_ha)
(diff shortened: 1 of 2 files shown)


bin/config-validator.rb
```diff
@@ -301,6 +301,7 @@ def check_clustering(manifest, bosh_properties)
       @has_errors += 1
     else
       next if has_script(role, @configure_ha)
+      next if role['name'] == 'secret-generation'
```
mook-as: Why skip secrets-generation? (Add a comment please)

PR #1587, opened 2018-06-08
Save logs of failed builds for post-mortem analysis

andreas-kupries:
Ref: https://trello.com/c/CSmv90Cr/696-3-jenkins-database-of-logs-to-track-failures

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -13,6 +13,10 @@ String jobBaseName() {
     return env.JOB_BASE_NAME.toLowerCase()
 }
 
+String getBuildLog() {
+    return currentBuild.rawBuild.getLogFile().getText()
+}
+
 void setBuildStatus(String context, String status) {
     def description = null
     switch (status) {
@@ -178,6 +182,11 @@ pipeline {
             name: 'S3_BUCKET',
             description: 'AWS S3 bucket to publish to',
             defaultValue: 'cap-release-archives',
+        )
+        string(
+            name: 'S3_LOG_BUCKET',
+            description: 'AWS S3 bucket to publish to',
+            defaultValue: 'cap-jenkins-logs',
         )
         string(
             name: 'S3_PREFIX',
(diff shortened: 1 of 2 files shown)


bin/clean-jenkins-log
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+cat the-build.log \
+    | sed \
+    -e 's|ha:////[a-zA-Z0-9/+][a-zA-Z0-9/+]*||' \
```
mook-as: If I'm reading the cleaned logs correctly, the `ha:////` thing is actually followed by a base64 string­ — which means _some_ lines also have a trailing `=` or `==`.

bin/clean-jenkins-log
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+cat the-build.log \
```
mook-as: Articles in things just tend to be confusing (why is this _the_ build log, as opposed to just _a_ build log?)

bin/clean-jenkins-log
andreas-kupries: Checking.

andreas-kupries: Confirmed. Fixed with commit 42b293167d3ef6f65cdf8926e2387410cc0d58c5

bin/clean-jenkins-log
andreas-kupries: No problem, will change.

andreas-kupries: Changed with commit 6b2b3ee83b2f572db54dec888f316d963d84b06b

bin/clean-jenkins-log
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+cat build.log \
+    | sed \
+    -e 's|ha:////[a-zA-Z0-9/+=][a-zA-Z0-9/+=]*||' \
```
mook-as: This actually matches more than required (since `=` can only appear at the end of the string, not in the middle), but I think we're fine with it.

Jenkinsfile
```diff
@@ -757,6 +766,30 @@ pass = ${OBS_CREDENTIALS_PASSWORD}
                     }
                 }
             }
-        }
+	    // Save logs of failed builds to s3 - we want to analyze where we may have jenkins issues.
+	    script {
+                if (env.BRANCH_NAME == 'develop' || env.BRANCH_NAME == 'master') {
+		    writeFile(file: 'build.log', text: getBuildLog())
+		    sh 'bin/clean-jenkins-log'
```
mook-as: Just realized that we should be passing the name of the log in as a parameter, instead of hard-coding `build.log` in the shell script. But whatever, we don't need to care about it.

PR #1590, opened 2018-06-11
Use groot-btrfs as an image plugin

viovanov:


container-host-files/etc/scf/config/role-manifest.yml
```diff
@@ -2004,6 +2002,12 @@ configuration:
   - name: HTTP_PROXY
     required: false
     description: Location of the proxy to use for regular web access.
+  - name: INSECURE_DOCKER_REGISTRIES
```
mook-as: Didn't we get support for `previous_names` or something? We should use that here.  Of course, having two things merge into one makes it trickier… but we should list at least one old name.

viovanov: Not sure that's useful for non-secrets.

jandubois: I agree. The `previous_names` mechanism is mostly needed to carry forward generated immutable secrets.

Mutual generated secrets would effectively just be rotated, and user specified values need to be re-specified on each upgrade anyways.

PR #1594, opened 2018-06-13
add app-autoscaler to scf

qibobo:


qibobo:
@mook-as Thank you for you comments. We will update the PR later.

container-host-files/etc/scf/config/role-manifest.yml
```diff
@@ -1618,6 +1786,494 @@ configuration:
...
+  - name: AUTOSCALER_SERVICE_BROKER_DB_CONFIG_MIN_CONNECTIONS
+    default: 0
+    description: ""
+ 
+  - name: AUTOSCALER_SERVICE_BROKER_HTTP_REQUEST_TIMEOUT
+    default: 5000
+    description: ""
+  - name: AUTOSCALER_SERVICE_BROKER_PASSWORD
+    default: "password"
+    description: ""
```
mook-as: This should be using a password generator

container-host-files/etc/scf/config/role-manifest.yml
```diff
@@ -2576,6 +3233,197 @@ configuration:
...
+    properties.autoscaler.api_server.scheduler.client_cert: '"((AUTOSCALER_ASACTORS_CLIENT_CERT))"'
+    properties.autoscaler.api_server.scheduler.client_key: '"((AUTOSCALER_ASACTORS_CLIENT_KEY))"'
+    properties.autoscaler.api_server.scheduler.host: '"autoscaler-actors.((KUBERNETES_NAMESPACE)).svc.((KUBERNETES_CLUSTER_DOMAIN))"'
+    properties.autoscaler.api_server.scheduler.port: '((AUTOSCALER_SCHEDULER_PORT))'
+    properties.autoscaler.api_server.server_cert: '"((AUTOSCALER_ASAPI_SERVER_CERT))"'
+    properties.autoscaler.api_server.server_key: '"((AUTOSCALER_ASAPI_SERVER_KEY))"'
+    properties.autoscaler.api_server.service_broker.ca_cert: '"((INTERNAL_CA_CERT))"'
+    properties.autoscaler.api_server.service_broker.client_cert: '"((AUTOSCALER_ASAPI_CLIENT_CERT))"'
+    properties.autoscaler.api_server.service_broker.client_key: '"((AUTOSCALER_ASAPI_CLIENT_KEY))"'
+    properties.autoscaler.api_server.service_broker.host: '"autoscaler-api.((KUBERNETES_NAMESPACE)).svc.((KUBERNETES_CLUSTER_DOMAIN))"'
```
mook-as: I believe nothing actually registers the broker to CF, right? Shouldn't it?

qibobo:
@mook-as 
this is the test guide of app-autoscaler: https://github.com/cloudfoundry-incubator/app-autoscaler-release/blob/master/src/acceptance/README.md

container-host-files/etc/scf/config/role-manifest.yml
mook-as: I actually needed `cf create-service-broker autoscaler p('autoscaler.service_broker.username') p('autoscaler.service_broker.password') http://<%= p('autoscaler.service_broker.api_server.host') %>:<%= p('autoscaler.service_broker.api_server.port') %>`, but yes.

I'm trying to get some tests in so that we don't accidentally break things; is there a smoke test for the autoscaler available? I got the acceptance tests integrated into SCF (will PR soon), but that took an hour on my machine.  I'm looking for something more like one minute or less.

mook-as:
Hi! Thanks for the changes.  I made a PR _into_ your branch to implement a smoke test of sorts, to ensure that we don't accidentally break the autoscaler in the future.  Three minutes was a bit longer than what I was looking for, but it looks like the default configuration doesn't let me lower it more.  Hopefully that shows you the sort of thing I'm looking for :)

Note that we might, in the future, end up defaulting the autoscaler to be opt-in (as in, zero instances of the relevant pods by default); that depends on how the vagrant boxes respond to having more things running on them — they're already a bit sluggish running CATS.  But that will happen in a subsequent PR.

qibobo:
@mook-as 
`cf create-service-broker autoscaler p('autoscaler.service_broker.username') p('autoscaler.service_broker.password') http://<%= p('autoscaler.service_broker.api_server.host') %>:<%= p('autoscaler.service_broker.api_server.port') %>`

Actually we should not use "autoscaler.service_broker.api_server.host" because it is the autoscaler apiserver host. But as apiserver and servicebroker are deployed in the same pod, it is ok. 
And it is better to use cf create-service-broker autoscaler {username} {password} https://autoscalerservicebroker.cf-dev.io as the autoscaler service broker has been registered to router.

And for acceptance test, to run all the tests parallel it will take 20-40 minutes, and if serially, it will take more time.
The acceptance tests can be divided into 3 types:

**1. service binding&unbinding**
       It takes about 2 or 3 minutes to test create service instance, bind service instance to an application, unbind service instance from the application and delete service instance.
**2. public api**
       It takes about 5 minutes to test all the public api of autoscaler.
**3. scaling in/out**
       It takes 20-30 minutes(run parallel) to test all the scaling in/out features.

As you mentioned, you want a test only takes a short time, I think you could just test the binding&unbinding.

mook-as:
@qibobo Thanks for the feedback! I'll update qibobo/scf#1 with your comments.

PR #1600, opened 2018-06-15
Changes to make it easier to run a subset of 🐱

mook-as:
These changes make it easier to run a subset of :cat: in order to attempt to reproduce failures.

Usage:
`make/tests acceptance-tests env.CATS_SUITES="=v3" env.CATS_FOCUS="health"`

### DIFF
--- a/bin/kube_overrides.rb
+++ b/bin/kube_overrides.rb
@@ -43,27 +43,29 @@ obj['spec']['containers'].each do |container|
         env['valueFrom']['secretKeyRef']['name'] = current_secrets_name
       end
     end
-    overrides.each do |k, v|
-      child = container
-      k[0...-1].each do |elem|
-        child[elem] ||= {}
-        child = child[elem]
+  end
+
+  overrides.each do |k, v|
+    child = container
+    k[0...-1].each do |elem|
+      child[elem] ||= {}
+      child = child[elem]
+    end
+    if k[0...1] == %w(env)
+      # Deal with the environment list specially, because the syntax isn't what
+      # humans normally want.
+      # The environment is actually in a list of hashes with "name" and "value"
+      # keys.  Erase any elements with the same name, and then append it.
+      child.reject! do |elem|
+        elem['name'] == k.last
       end
-      case child
-      when Array
-        # Deal with the environment list specially
-        child.reject! do |elem|
-          elem['name'] == k.last
-        end
-        child << {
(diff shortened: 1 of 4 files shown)


bin/kube_overrides.rb
```diff
@@ -37,6 +43,28 @@
         env['valueFrom']['secretKeyRef']['name'] = current_secrets_name
       end
     end
+    overrides.each do |k, v|
```
jandubois: Why is this code inside the `container['env'].each do |env|` loop?

For the `Array` case it doesn't matter, as it will delete and add-back the override over and over, but for a `Hash` it would add the override multiple times, once for each `env` entry.

bin/kube_overrides.rb
```diff
@@ -37,6 +43,28 @@
...
+      child = container
+      k[0...-1].each do |elem|
+        child[elem] ||= {}
+        child = child[elem]
+      end
+      case child
+      when Array
+        # Deal with the environment list specially
```
jandubois: Why does this mention "the environment list specially" when it then applies the filtering to all arrays?

bin/kube_overrides.rb
```diff
@@ -37,6 +43,28 @@
...
+        child.reject! do |elem|
+          elem['name'] == k.last
+        end
+        child << {
+          'name'  => k.last,
+          'value' => v,
+        }
+      when Hash
```
jandubois: What is the use case for the hash here?

I think it would be helpful if the comments would include an example of each kind of override.

bin/kube_overrides.rb
jandubois: > it would add the override multiple times

Well no, it wouldn't. So it makes no difference, after all, but I still would like to see the code pulled out of the loop. 😄

mook-as: Fixed in 7b3c931f0c7ec9d0f15e7fa29968b6b2a1baaa98 because it's just kind of embarrassing :smile:

bin/kube_overrides.rb
mook-as: Fixed in 7b3c931f0c7ec9d0f15e7fa29968b6b2a1baaa98; changed the condition to actually only work with the environment list.

bin/kube_overrides.rb
mook-as: Added example (image pull policy).

PR #1607, opened 2018-06-19
Use namespace for helm install name

alexjh:


make/upgrade
```diff
@@ -15,8 +15,7 @@ cd "${GIT_ROOT}"
 . make/include/secrets
 
 # Argument? (**)
-RELEASE=$(helm list --date --reverse --max 1 --namespace "${NAMESPACE}" -q | tail -n 1)
-echo Upgrading ${NAMESPACE} release \"${RELEASE}\" ...
+echo Upgrading ${NAMESPACE} release \"${NAMESPACE}\" ...
```
mook-as: This would be clearer, I think, if we do a `RELEASE="${NAMESPACE}" (with appropriate comment) instead?

make/upgrade
```diff
@@ -38,15 +37,16 @@ source "${NETWORK_ENV}"
 
 UAA_CA_CERT="$(get_secret "${UAA_NAMESPACE}" "uaa" "INTERNAL_CA_CERT")"
 
-helm upgrade ${RELEASE} output/helm \
-    --namespace ${NAMESPACE} \
+helm upgrade "${NAMESPACE}" output/helm \
+    --name "${NAMESPACE}" \
```
mook-as: Does `helm upgrade` take a `--name`?

make/upgrade
alexjh: :+1: Fixed and rebased to simplify the diff.

make/upgrade
alexjh: Oops, no, it doesn't. Fixed in the rebased branch.

PR #1611, opened 2018-06-22
Add cluster teardown instructions

alexjh:


README.md
```diff
@@ -351,6 +352,42 @@ make stop
...
+operation more likely to succeed:
+
+```bash
+kubectl delete statefulsets --all --namespace ${CF_NAMESPACE}
+helm delete --purge ${CF_RELEASE_NAME}
+kubectl delete namespace ${CF_NAMESPACE}
+kubectl delete statefulsets --all --namespace ${UAA_NAMESPACE}
+helm delete --purge ${UAA_RELEASE_NAME}
+kubectl delete namespace ${UAA_NAMESPACE}
+```
```
andreas-kupries: Is this something we should put into our `make/stop` scripts ?

alexjh: I don't think so... I've been testing this yesterday and I haven't been able to get the Vagrant box to break, even under high load. And I think this will probably get fixed by helm / kube in the future, so I'd rather not have the workaround running all the time if we don't need it.

PR #1613, opened 2018-06-25
Add ROUTER_TLS_PEM structured property

jandubois:
It is an alternative (and overrides) ROUTER_SSL_CERT (and KEY) and allows specification of multiple cert/key pairs.


jandubois:
WIP because fissile still points to a checks build.

PR #1619, opened 2018-06-27
Set persistent storage class as default

alexjh:
Note that https://github.com/SUSE/uaa-fissile-release/pull/112 is the real PR and this should point to UAA master again after it's merged.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -429,7 +429,7 @@ pipeline {
                     fi
 
                     kubectl delete storageclass hostpath || /bin/true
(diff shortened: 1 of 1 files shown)


alexjh:
I fixed the Jenkinsfile and added a check for the hostpath, and https://jenkins-new.howdoi.website/job/scf/view/change-requests/job/PR-1619/lastSuccessfulBuild/consoleFull ran successfully:

```
02:01:17 + kubectl get storageclass
02:01:17 NAME                 PROVISIONER
02:01:17 hostpath (default)   kubernetes.io/host-path
```

I just force-pushed the branch to remove the print statement.

PR #1625, opened 2018-06-29
Fix internal ca key exposed

jimmykarily:
`env | grep CA_KEY` doesn't return the key in the pods anymore. My `api-0` didn't come up when testing this because monit wasn't running in the container but it could be a random (irrelevant) issue. Let's wait for CI to see if it is all green.

### DIFF
--- a/bin/config-validator.rb
+++ b/bin/config-validator.rb
@@ -340,7 +340,8 @@ def check_rm_variables(manifest)
   end
 
   manifest['configuration']['variables'].each do |variable|
-    next if Common.special_use(variable['name'])
+    # "internal" variables are defined but not used in the role manifest. They are referenced directly in scripts.
+    next if variable['internal']
     found = templates.any? do |template|
       Common.parameters_in_template(template).include?(variable['name'])
     end
--- a/container-host-files/etc/scf/config/role-manifest.yml
+++ b/container-host-files/etc/scf/config/role-manifest.yml
@@ -3267,7 +3267,9 @@ configuration:
     ip: '"((IP_ADDRESS))"'
     networks.default.dns_record_name: '"((#KUBE_COMPONENT_NAME))((KUBE_COMPONENT_NAME))((/KUBE_COMPONENT_NAME))((^KUBE_COMPONENT_NAME))((DNS_RECORD_NAME))((/KUBE_COMPONENT_NAME))"'
     networks.default.ip: '"((IP_ADDRESS))"'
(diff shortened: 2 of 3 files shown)


jandubois:
Sorry, I did not get around to reviewing it today. Will do so on Tuesday...

bin/config-validator.rb
```diff
@@ -340,7 +340,7 @@ def check_rm_variables(manifest)
   end
 
   manifest['configuration']['variables'].each do |variable|
-    next if Common.special_use(variable['name'])
+    next if variable['internal']
```
jandubois: I think we keep the comment from the `special_use` function here. Something like

```
# "internal" variables are defined but not used in the role manifest. They are referenced directly in scripts.
```

container-host-files/etc/scf/config/role-manifest.yml
```diff
@@ -2023,7 +2023,6 @@ configuration:
       type: CACertificate
       value_type: private_key
     description: PEM-encoded CA key.
-    required: true
```
jandubois: Why is the `required` attribute removed? It is kind of redundant for secrets with generators, but there should be no reason to remove the setting.

jimmykarily:
@jandubois thanks for the review. There is another card to let users set the KEY (since it need some discussion first as you pointed out). This one just prevents exposing it for now.

container-host-files/etc/scf/config/role-manifest.yml
jimmykarily: This is just cleanup. Agreed, random cleanup but has to happen anyway. Want me to move it to a separate commit?

jandubois: I don't understand your comment. How is this "cleanup"? The setting is required, so should keep the attribute, even though it doesn't do anything because the generator will always make sure this setting exist.  All other certs have the `required: true` attribute, so I don't know why this "has to happen anyway".

I don't actually care that much about having the attribute itself, but I do want to understand why it has to be removed here, as that would indicate an issue elsewhere.

viovanov: @jandubois I'll PR a cleanup for the role manifest.
the reason to remove it is "it doesn't do anything".

jimmykarily: ok I removed the removal

jandubois: > @viovanov: I'll PR a cleanup for the role manifest.

Can you please hold back on that until the role re-arrangement has been merged to avoid merge conflicts...

PR #1633, opened 2018-07-04
Realigning our role composition back with cf upstream

jandubois:
This PR has been replaced by #1647 and can be deleted once we are sure I didn't make any mistakes during the rebasing and cherry-picking... 😄


jandubois:
> This PR has been replaced by #1647

PR #1638, opened 2018-07-05
Cleanup role manifest and extra validation.

viovanov:
Remove "process" key from all roles. It's not used.
Generated variables don't need the "required" key. Also validated.
Variables must be documented.

### DIFF
--- a/bin/config-validator.rb
+++ b/bin/config-validator.rb
@@ -350,6 +350,25 @@ def check_rm_variables(manifest)
     STDOUT.puts "role manifest variable #{variable['name'].red} was not found in any role manifest template"
     @has_errors += 1
   end
+
+  manifest['configuration']['variables'].each do |variable|
+    no_description = variable["description"].to_s.strip.empty? 
+
+    next unless no_description
+
+    STDOUT.puts "role manifest variable #{variable['name'].red} has an empty description"
+    @has_errors += 1
+  end
+
+  manifest['configuration']['variables'].each do |variable|
+    is_generated = !variable["generator"].nil? 
+    is_marked_required = variable["required"]
+
+    next unless is_generated && is_marked_required
+
+    STDOUT.puts "role manifest variable #{variable['name'].red} is generated and has 'required: true'"
+    @has_errors += 1
+  end
 end
 
 def global_variables(manifest)
--- a/container-host-files/etc/scf/config/opinions.yml
+++ b/container-host-files/etc/scf/config/opinions.yml
@@ -55,6 +55,20 @@ properties:
       port: 5432
     scheduler_db:
       port: 5432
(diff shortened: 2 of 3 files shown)


viovanov:
WIP until validation passes.

bin/config-validator.rb
```diff
@@ -350,6 +350,25 @@ def check_rm_variables(manifest)
     STDOUT.puts "role manifest variable #{variable['name'].red} was not found in any role manifest template"
     @has_errors += 1
   end
+
+  manifest['configuration']['variables'].each do |variable|
```
mook-as: We should put any new validation in fissile instead of here (because that gets better tested, and also automatically applies to UAA too).

container-host-files/etc/scf/config/opinions.yml
```diff
@@ -40,6 +40,78 @@ properties:
...
+  autoscaler:
+    api_server:
+      db_config:
+        idle_timeout: 100
+        max_connections: 10
+        min_connections: 0
+      info:
+        build: "beta"
+        description: "Provides autoscaling capabilities based on policies, for apps deployed on Cloud Foundry"
+        name: "SCF App Autoscaler"
```
mook-as: Hmm, we'll need to figure out how to make white-labelling this easier in the future.

Of course, that might just be ops files…

qibobo:
@viovanov @mook-as noticed that many autoscaler properties are removed from the role manifest, why?

bin/config-validator.rb
viovanov: I agree, but I think it's out of scope for this.
I've created https://trello.com/c/TD9VSPOl/430-fissile-move-validation-code-from-scf-to-fissile

container-host-files/etc/scf/config/opinions.yml
viovanov: We could expose them in the manifest again if they're ever required.

viovanov:
@qibobo our approach is to only expose something if it's _required_.
Many of the settings are really opinions - things users never need to change, that's why they've been moved to opinions.
The second reason is lack of descriptions.

If you think you need one of those settings to be exposed, please create a PR, that contains an appropriate description and a use case.

bin/config-validator.rb
mook-as: Hmm? No, I'm not asking you to move the existing checks; only move any _new_ ones you create (hence would be in scope).

qibobo:
@viovanov @mook-as  Thank  you！There are many properties for app-autoscaler in role manifest and they should be configurable in deployment by the admin according to the deploy environment to get the best performance. I'd like to raise a PR to add descriptions for the app-autoscaler properties.

PR #1645, opened 2018-07-20
Implement a better mechanism for waiting on mysql.

andreas-kupries:
Ref: https://trello.com/c/nG2H79ig/743-cf-usb-exit-with-panic-during-brain-test-after-upgrade-from-2101-2110

- Installing the mysql client in the testbrain package.
- Using the client to wait for the database to become ready.
- Note, dropped `wait_on_port` using `curl`.
  The new `wait_on_database` using `mysql` is enough.

### DIFF
--- a/src/scf-release/packages/acceptance-tests-brain/packaging
+++ b/src/scf-release/packages/acceptance-tests-brain/packaging
@@ -78,3 +78,7 @@ mv "${BOSH_COMPILE_TARGET}/github.com" "${BOSH_COMPILE_TARGET}/src/github.com"
 
 # We currently have no way of getting a valid version number, as we have no git repo
 GOBIN="${BIN_DIR}" go install -ldflags="-X main.version=0.0.0" github.com/SUSE/testbrain
+
+# Install the mysql client into the package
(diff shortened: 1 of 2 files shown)


src/scf-release/packages/acceptance-tests-brain/packaging
```diff
@@ -78,3 +78,7 @@ mv "${BOSH_COMPILE_TARGET}/github.com" "${BOSH_COMPILE_TARGET}/src/github.com"
 
 # We currently have no way of getting a valid version number, as we have no git repo
 GOBIN="${BIN_DIR}" go install -ldflags="-X main.version=0.0.0" github.com/SUSE/testbrain
+
+# Install the mysql client into the package
+zypper install -y mariadb-client
+cp /usr/bin/mysql ${BIN_DIR}
```
mook-as: How do we make sure we pull in the dependencies?

andreas-kupries: The mysql client depends only on the system libraries like libc. No other dependencies.
Which is why I can get away with copying just that binary.

PR #1649, opened 2018-08-01
Keep GARDEN_ROOTFS_DRIVER

viovanov:
This is simpler for users and doesn't require a change when upgrading.

### DIFF
--- a/container-host-files/etc/scf/config/role-manifest.yml
+++ b/container-host-files/etc/scf/config/role-manifest.yml
@@ -1315,6 +1315,7 @@ roles:
   - scripts/configure-HA-hosts.sh
   - scripts/set-diego-cell-memory-limits.sh
   - scripts/go_log_level.sh
+  - scripts/configure-garden-filesystem.sh
   scripts:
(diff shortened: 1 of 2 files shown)


container-host-files/etc/scf/config/role-manifest.yml
```diff
@@ -2665,9 +2666,14 @@ configuration:
...
     description: 'Override DNS servers to be used in containers; defaults to the same as the host.'
-  - name: GARDEN_USE_BTRFS
+  - name: GARDEN_ROOTFS_DRIVER
+    default: btrfs
+    description: The filesystem driver to use (btrfs or overlay-xfs).
+    required: true
+  - name: GARDEN_DISABLE_BTRFS
     default: ""
-    description: An empty value means BTRFS is used. Set to "false" to use overlay-xfs.
+    description: An empty value means BTRFS is used. It's set to "false" by a script if GARDEN_ROOTFS_DRIVER is set to overlay-xfs.
```
HartS: It's set to "true"

PR #1653, opened 2018-08-03
autoscaler: Always use client_credentials

mook-as:
In the context of SCF (/other fissilized CF deployment), it is possible for the administrator to change the administrator password and then do something that requires the autoscaler to be restarted.  In that case the password embedded in the helm release will no longer be correct, which will cause the autoscaler to cease working.

Instead, use client credentials for the autoscaler (which would be generated anyway).  This ensures that we always have an auth method that will work even when the secrets are rotated.

This is independent of the autoscaler being used outside of SCF (where it might still be a good idea to use username / password authentication).

/cc @qibobo Could you see if this would make sense?  Longer term, we are planning on moving fissile to use BOSH ops files instead, in which case none of this matters (since the operator would then be able to change any BOSH property they please).

### DIFF
--- a/container-host-files/etc/scf/config/opinions.yml
+++ b/container-host-files/etc/scf/config/opinions.yml
@@ -40,6 +40,12 @@ properties:
     user: admin
     org: test-brain-org
     space: test-brain-space
+  autoscaler:
+    cf:
+      client_id: app_autoscaler
+      grant_type: client_credentials
+      password: "" # unused
+      username: "" # unused
   capi:
     tps:
       watcher:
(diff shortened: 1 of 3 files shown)


qibobo:
@mook-as We prefer to use client_credentials for autoscaler and we always deploy autoscaler in our local by client_credentials. In the previous PR, there is no uaa client for autoscaler so we use user/password. It is good for us. Please go ahead. Thanks.

mook-as:
To test this PR:
* Deploy as normal, check that the scaler test pass etc.
* Change the admin password, via `cf passwd`
* Run the test again, as `make/tests scaler-smoke env.CF_CLUSTER_ADMIN_PASSWORD=<new password>`

andreas-kupries:
Damn. Needs another rebase :arrow_left: @mook-as (or conflict resolution here in the web interface. It looks trivial enough)

PR #1656, opened 2018-08-07
Enable HA for UAA mysql-proxy

alexjh:
https://github.com/SUSE/uaa-fissile-release/pull/115 should be merged first.


alexjh:
This isn't necessary now that the db and proxy roles have been merged into the `database` role.

PR #1666, opened 2018-08-13
versions: bump fissile stemcell for configgin 0.16.2

mook-as:
For some reason the openSUSE one was using `develop` builds instead of `release` builds; moved that back to match the SLE one.

### DIFF
--- a/bin/common/versions.sh
+++ b/bin/common/versions.sh
@@ -24,9 +24,9 @@ export STAMPY_MAJOR=$(echo "$STAMPY_VERSION" | sed -e 's/\.g.*//' -e 's/\.[^.]*$
 
 if [ "${USE_SLE_BASE:-false}" == "false" ]
 then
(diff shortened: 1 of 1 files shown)


jandubois:
But you are reverting the openSUSE stemcell now to the version that doesn't include the monit patch; is that what you wanted? Shouldn't it be updated for the new fissile version too?

jandubois:
> the version that doesn't include the monit patch

Or maybe not; I guess I don't understand the format of the version tag: `42.3-5.g9215f81-30.31`

What is the final `30.31` part?

mook-as:
I don't actually know; I just grabbed the last `fissile:release` stemcell, instead of the `fissile:develop` one that it was using before.  If that's causing issues, we had issues in the SLE builds (which was already using `fissile:release` before my changes).

andreas-kupries:
Info from Jan on :rocket: chat:

>  My initial assumption about reverting the monit change was incorrect, but I have not looked further if the latest tags are correct or not

PR #1671, opened 2018-08-14
Remove DNS_HEALTH_CHECK_HOST variable

viovanov:
The health check host can be the API role, and it can be automatically determined.
This also fixes issues with Core DNS.


jandubois:
Has been replaced by #1744

PR #1681, opened 2018-08-16
Remove superfluous repositories from the SLE stemcell

aduffeck:


aduffeck:
The pr SLES build just failed in the same way as the openSUSE master build, so the fix for the broken download from internal repositories works as expected.

PR #1700, opened 2018-08-24
Allow the dev to disable the host compilation cache

viovanov:


.envrc
```diff
@@ -88,7 +88,9 @@ export PATH
 export FISSILE_CONFIG_PREFIX=scf
 
 # We will persist compilation results to the VM here
-export SCF_PACKAGE_COMPILATION_CACHE="${SCF_PACKAGE_COMPILATION_CACHE:-${PWD}/.fissile/compilation}"
+if [ -z ${SCF_PACKAGE_COMPILATION_CACHE+x} ]; then
```
jandubois: What is the `+x` at the end of the variable name?

.envrc
```diff
@@ -88,7 +88,9 @@ export PATH
...
+  export SCF_PACKAGE_COMPILATION_CACHE="${PWD}/.fissile/compilation"
+fi
```
jandubois: Isn't the idiomatic way to write this:

```
export SCF_PACKAGE_COMPILATION_CACHE="${SCF_PACKAGE_COMPILATION_CACHE:-${PWD}/.fissile/compilation}"
```

Finally, the functionality doesn't seem to match the PR description: you don't disable the cache, you just point it to a different location.

.envrc
viovanov: Returns true if the var is unset, false if it's a string of length >= 0

jandubois: Ok, thanks!

Maybe add a comment above that explains that you want to set the default for an unset variable, but want to keep the empty string because that disables caching?

However, I think I would prefer to use a separate variable `SCF_DISABLE_PACKAGE_COMPILATION_CACHE` instead of making the empty string a magic value distinct from the null value.

.envrc
jandubois: The fact that I was confused about this shows that at least additional documentation would be helpful. Using a separate variable to disable caching would be even clearer.

viovanov: I agree, bash is making not making this intuitive to read.
There's a `SCF_DISABLE_PACKAGE_COMPILATION_CACHE` now.

.envrc
```diff
@@ -88,7 +88,12 @@ export PATH
 export FISSILE_CONFIG_PREFIX=scf
 
 # We will persist compilation results to the VM here
-export SCF_PACKAGE_COMPILATION_CACHE="${SCF_PACKAGE_COMPILATION_CACHE:-${PWD}/.fissile/compilation}"
+# Caching is disabled if SCF_DISABLE_PACKAGE_COMPILATION_CACHE is set
+if [ -z ${SCF_DISABLE_PACKAGE_COMPILATION_CACHE+x} ]; then
```
jandubois: Why aren't you checking for `SCF_DISABLE_PACKAGE_COMPILATION_CACHE` directly in https://github.com/SUSE/scf/blob/develop/make/compile#L28 instead of setting `SCF_PACKAGE_COMPILATION_CACHE` to a magic value here, and then testing for the magic value later?

viovanov: I think it's fine, it's what the `.envrc` is supposed to do - calculate your env.

PR #1701, opened 2018-08-24
Rename database role back to mysql

jandubois:
We need additional support to migrate data from previous installation during upgrade if we want to change the names of volume claims.

[trello#aOhLhzFF]


gaktive:
LGTM

PR #1702, opened 2018-08-27
 Fix role manifest missing descriptions and bump fissile

viovanov:
#1700 should be merged first
Depends on https://github.com/SUSE/fissile/pull/385

### DIFF
--- a/bin/common/versions.sh
+++ b/bin/common/versions.sh
@@ -9,7 +9,7 @@ set -o errexit -o nounset
 
 export BOSH_CLI_VERSION="fcaa9c6caff58ab8da8c56481320681cdea492ee"
 export CFCLI_VERSION="6.37.0"
-export FISSILE_VERSION="6.0.0+0.g711088f"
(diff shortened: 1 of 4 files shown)


viovanov:
[(5) fissile: move validation code from scf to fissile](https://trello.com/c/TD9VSPOl/769-5-fissile-move-validation-code-from-scf-to-fissile)

bin/common/versions.sh
```diff
@@ -9,7 +9,7 @@ set -o errexit -o nounset
 
 export BOSH_CLI_VERSION="fcaa9c6caff58ab8da8c56481320681cdea492ee"
 export CFCLI_VERSION="6.37.0"
-export FISSILE_VERSION="6.0.0+0.g711088f"
+export FISSILE_VERSION="fissile-6.0.0+8.g54a0c07.linux-amd64.tgz"
```
mook-as: There's too much stuff in the version (it breaks the URL)

PR #1716, opened 2018-08-31
Add check that only the CHANGELOG.md was changed

alexjh:


Jenkinsfile
```diff
@@ -243,6 +243,32 @@ pipeline {
...
+                  for (int k = 0; k < files.size(); k++) {
+                    def file = files[k]
+                    println file.path
+                    all_files << file.path
+                  }
+                }
+              }
+		  
+	      if (all_files.length == 1 && all_files[0] == 'CHANGELOG.md') {
+	        error "Only CHANGELOG.md was modified"
```
mook-as: We _probably_ don't need a rebuild for `README.md` (and other markdown files), either?  Up to you; maybe there are situations where we do…

Jenkinsfile
```diff
@@ -243,6 +243,32 @@ pipeline {
     }
 
     stages {
+        stage('check_for_changed_files') {
+          steps {
+            script {
+	      def all_files = new ArrayList()
+
+              def changeLogSets = currentBuild.changeSets
+              for (int i = 0; i < changeLogSets.size(); i++) {
```
mook-as: Does `for (changeSet  in currentBuild.changeSets)` not work? (I don't actually know)

alexjh: I think it might... if I could figure out how to get a changeSet :(

Jenkinsfile
```diff
@@ -332,6 +348,35 @@ pipeline {
                 checkout scm
             }
         }
+        stage('check_for_changed_files') {
+          steps {
+            script {
+	      if (env.BRANCH_NAME == 'master') {
```
mook-as: Do we want to use a `when` here too? This is more optional, but it would be nice to see that they're skipped on purpose on master, instead of just having no output.

Jenkinsfile
```diff
@@ -332,6 +348,35 @@ pipeline {
...
+	        return
+	      }
+
+	      def all_files = new HashSet<String>()
+
+              for (set in currentBuild.changeSets) {
+                def entries = set.items
+                for (entry in entries) {
```
mook-as: Do we need an intermediate, or is `for (entry in set.items)` good enough?

Jenkinsfile
```diff
@@ -332,6 +348,35 @@ pipeline {
...
+                  for (file in entry.affectedFiles) {
+                    all_files << file.path
+                  }
+                }
+              }
+
+	      echo "All files changed since last build: ${all_files}"
```
mook-as: Nit: indent is off, it seems.  It's also off in the `if` block below.

Jenkinsfile
```diff
@@ -17,6 +17,22 @@ String getBuildLog() {
     return currentBuild.rawBuild.getLogFile().getText()
 }
 
+boolean areIgnoredFiles(HashSet<String> changedFiles) {
+  HashSet<String> ignoredFiles = [
+    "CHANGELOG.md",
+    "README.md"
```
mook-as: Add a comment that this is the full path to the files, relative to the root of the repo?

Jenkinsfile
```diff
@@ -332,6 +348,35 @@ pipeline {
...
+        stage('check_for_changed_files') {
+          steps {
+            script {
+	      if (env.BRANCH_NAME == 'master') {
+	        return
+	      }
+
+	      def all_files = new HashSet<String>()
+
+              for (set in currentBuild.changeSets) {
```
mook-as: This is checking for changes since the last build, which is… probably correct. Might want a comment though.

(That is: if no relevant files changed since the last build on the same branch happened, this wouldn't build)

Jenkinsfile
andreas-kupries: You mean something like
```
         when {
            expression { env.BRANCH_NAME != 'master' }
         }
         steps {
            script {
	      def all_files = new HashSet<String>()
```
Note, I am not sure if my syntax is correct here.

Jenkinsfile
andreas-kupries: I have no idea here about what groovy need/wants/balks at.

Jenkinsfile
andreas-kupries: leading tab instead spaces

Jenkinsfile
andreas-kupries: Handled with commit e3ecd5fe538df3862b54e194b822fdd4f27cd987

Jenkinsfile
andreas-kupries: Handled with commit e3ecd5fe538df3862b54e194b822fdd4f27cd987

Jenkinsfile
mook-as: Yep, that is correct.

andreas-kupries: Handled with commit 25b6ffb7832d599209b8673be4d57589c7d0582b

Jenkinsfile
andreas-kupries: Lets leave it in for now. Does not hurt I believe.

PR #1717, opened 2018-09-04
Bump scf-helper-release

viovanov:


viovanov:
The conclusion of https://github.com/SUSE/scf-helper-release/pull/13

jandubois:
Shouldn't there be a corresponding change to `uaa-fissile-release` as well?

PR #1722, opened 2018-09-07
Ensure master builds are started with a trigger

alexjh:


Jenkinsfile
```diff
@@ -243,6 +248,16 @@ pipeline {
     }
 
     stages {
+        stage('ensure_triggered_master') {
```
mook-as: This should probably use a `when { expression { return env.BRANCH_NAME == 'master' } }` so that we don't see this run on non-master builds.
(Also, the indenting looks odd)

alexjh: Oh, I didn't realize that's what `when` is for.

Fixed in 41ddc723

PR #1694, opened 2018-08-22
One kube service per bosh job

alexjh:
This PR will have to go through several rounds of updates for:

* fissile
* configgin
* fissile stemcell
* scf-helper-release
* uaa-fissile-release

PR #1732, opened 2018-09-12
Translate to new format

aduffeck:
This pull request adapts the role manifest and the according validation code to the latest format changes in fissile.

**Note**: This requires changes in fissile which are in the `develop` branch, but `develop` also includes commits which require additional changes regarding the one-service-per-job change. This PR should thus currently **NOT** be merged until this is resolved.

**Note 2**: There's another pull request regarding this format change for [uaa-fissile-release](https://github.com/SUSE/uaa-fissile-release/pull/121). The submodule in `scf` or in this PR should be bumped once this is merged.

### DIFF
--- a/README.md
+++ b/README.md
@@ -475,7 +475,7 @@ You can access any URL or endpoint that references this address from your host.
 3. Modify the `role-manifest.yml`:
 
     1. Add new roles or change existing ones
-    1. Add exposed environment variables (`yaml path: /configuration/variables`).
+    1. Add exposed environment variables (`yaml path: /variables`).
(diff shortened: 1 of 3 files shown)


jandubois:
This PR cannot be made to work on it's own. I have rebased #1694 on top of this PR, so when it gets merged, then this PR will be merged as well.

jandubois:
> I have rebased #1694 on top of this PR

The rebased PR currently lives in a separate branch under `jandubois/one-kube-service-per-bosh-job`. I will push it to the original PR branch once it no longer hangs during startup because I don't want it to block CI.

PR #1752, opened 2018-09-24
Submit obs buildpack dependencies for legal review

thardeck:
Submit OBS built buildpack dependencies for legal review. For the old buildpacks the sources are committed via Concourse which was not the case for the buildpacks built in obs.

### DIFF
--- a/bin/dev/legalscan.sh
+++ b/bin/dev/legalscan.sh
@@ -1,10 +1,6 @@
 #!/usr/bin/env bash
 
-for PACKAGE in $(osc ls Cloud:Platform:buildpacks:dependencies); do
-  osc submitrequest --yes --message=legal-review Cloud:Platform:buildpacks:dependencies $PACKAGE Cloud:Platform:sources:buildpacks:reviewed
-done
-
-PROJECTS="Cloud:Platform:sources:scf Cloud:Platform:sources:buildpacks Cloud:Platform:sources:sidecars"
+PROJECTS="Cloud:Platform:sources:scf Cloud:Platform:sources:buildpacks Cloud:Platform:buildpacks:dependencies Cloud:Platform:sources:sidecars"
 for PROJECT in ${PROJECTS}; do
   for PACKAGE in $(osc ls $PROJECT); do
     osc submitrequest --yes --message=legal-review $PROJECT $PACKAGE ${PROJECT}:reviewed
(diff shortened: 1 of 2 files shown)


bin/dev/legalscan.sh
```diff
@@ -1,4 +1,9 @@
 #!/usr/bin/env bash
+
+for PACKAGE in $(osc ls Cloud:Platform:buildpacks:dependencies); do
+  osc submitrequest --yes --message=legal-review Cloud:Platform:buildpacks:dependencies $PACKAGE Cloud:Platform:sources:buildpacks:reviewed
```
dragonchaser: Why move the dependencies to `Cloud:Platform:sources:buildpacks:reviewed` instead of `Cloud:Platform:buildpacks:dependencies:reviewed`? This would make more sense IMHO?

bin/dev/legalscan.sh
```diff
@@ -1,4 +1,9 @@
 #!/usr/bin/env bash
+
+for PACKAGE in $(osc ls Cloud:Platform:buildpacks:dependencies); do
```
dragonchaser: Are `Cloud:Platform:sources:scf`, `Cloud:Platform:sources:buildpacks` and `Cloud:Platform:sources:sidecars` no longer reviewed?

bin/dev/legalscan.sh
dragonchaser: IMHO it would make sense to move the project names to variables.

bin/dev/legalscan.sh
thardeck: I have wrote a simple script to accept them as soon as legaldb is done. This prevents tarballs from being recomitted even though they were not changed.

bin/dev/legalscan.sh
thardeck: I did not move them, we just also submit them there so they get review by legaldb.

thardeck: At the moment we only use them in one script. What variable names would you suggest?

dragonchaser: Is it much of an effort to rename the target repository?

dragonchaser: `SOURCE_OBS_PROJECT` and maybe `TARGET_OBS_PROJECT`

thardeck: We need submit requests, which are only possible from one repo to another. So we need two anyway afaik.

PR #1767, opened 2018-09-27
Merge from master as 2.14.0

gaktive:
The changes tied to the one kube server per job need to be clearer in the release notes but should be OK as is here. Let me know if I'm wrong.


CHANGELOG.md
```diff
@@ -2,6 +2,25 @@
 
 ## [Unreleased]
 
+## [2.14.0] - 2018-09-27
+
+### Added
+- Credhub introduced as a user-accessible component (independent of future cf-deployment requirements)
+- Exposed SMTP variable to allow for account creation & password reset
```
jandubois: Why not specify the exact name of the new variable, and where it is set (in UAA)?

Just looked it up, and it is not one, but a while bunch of variables: https://github.com/SUSE/uaa-fissile-release/commit/382b70b0e30a465f7c355bf377f2afbfb0881394, but `SMTP_HOST` and `SMTP_FROM_ADDRESS` might be just the minimum that need to be set, depending on the server.

CHANGELOG.md
```diff
@@ -2,6 +2,25 @@
...
+
+### Changed
+- One Kubernetes service per job now, which will require changes to role manifest YAML
```
jandubois: More importantly, the service name has changed to include both the instance_group (was role) name, *and* the job name.

CHANGELOG.md
```diff
@@ -2,6 +2,25 @@
...
+- Credhub introduced as a user-accessible component (independent of future cf-deployment requirements)
+- Exposed SMTP variable to allow for account creation & password reset
+
+### Changed
+- One Kubernetes service per job now, which will require changes to role manifest YAML 
+- Bumped python & Ruby buildpacks
+- Bumped SLE12 & openSUSE stacks
+
+### Fixed
+- Kubernetes readiness check looks for both apiserver & kubelet now
```
jandubois: We always checked for those, but previous only found them when started by hyperkube. Now we also find them when started by other means.

CHANGELOG.md
gaktive: Revised to: `Exposed SMTP_HOST & SMTP_FROM_ADDRESS variables to allow for account creation & password reset`

CHANGELOG.md
gaktive: Reworded to `One Kubernetes service per job now, whereby the service names will include both the instance group (previously the role) and job name, which impacts the role manifest YAML`

CHANGELOG.md
gaktive: New wording: `Kubernetes readiness check no longer looks for hyperkube explicitly`

PR #1783, opened 2018-10-03
[#DVNMaxd0] Update role sizing information from the Azure testing.

andreas-kupries:
Ref: https://trello.com/c/DVNMaxd0/735-5-evaluate-sizing-tests-and-update-docs-helm-charts-for-13

### DIFF
--- a/container-host-files/etc/scf/config/role-manifest.yml
+++ b/container-host-files/etc/scf/config/role-manifest.yml
@@ -19,7 +19,7 @@ instance_groups:
             max: 3
             ha: 2
           capabilities: []
-          memory: 69
+          memory: 128
           virtual-cpus: 2
           affinity:
             podAntiAffinity:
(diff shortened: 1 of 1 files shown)


andreas-kupries:
Snapshot data attached to the referenced ticket.

container-host-files/etc/scf/config/role-manifest.yml
```diff
@@ -190,7 +190,7 @@ instance_groups:
             type: persistent
             tag: mysql-data
             size: 20
-          memory: 2950
+          memory: 2500
```
mook-as: `mysql-proxy`, by itself, takes 2.5GiB? And `mysql` (the actual database) has no limits?
Something sounds really wrong. (Probably in how we do limits)

The nearest round number is 2560.

container-host-files/etc/scf/config/role-manifest.yml
```diff
@@ -1103,7 +1103,7 @@ instance_groups:
           - path: /sys/fs/cgroup
             type: host
             tag: host-cgroup
-          memory: 4677
+          memory: 2800
```
mook-as: This probably wants a round number

PR #1798, opened 2018-10-13
Bump fissile to get kube labels for services

jandubois:


jandubois:
This fissile bump was already included in #1794

PR #1832, opened 2018-10-31
Turns out this is more generic

viovanov:
The FQDN won't work on latest GKE.

### DIFF
--- a/container-host-files/etc/scf/config/role-manifest.yml
+++ b/container-host-files/etc/scf/config/role-manifest.yml
@@ -480,7 +480,7 @@ instance_groups:
     release: loggregator-agent
   configuration:
     templates:
-      properties.dns_health_check_host: 'mysql-set.((KUBERNETES_NAMESPACE)).svc.((KUBERNETES_CLUSTER_DOMAIN))'
+      properties.dns_health_check_host: 'mysql-set'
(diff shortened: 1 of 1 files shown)


container-host-files/etc/scf/config/role-manifest.yml
```diff
@@ -1968,7 +1968,7 @@ configuration:
     properties.router.balancing_algorithm: '"((ROUTER_BALANCING_ALGORITHM))"'
     properties.router.client_cert_validation: '"((ROUTER_CLIENT_CERT_VALIDATION))"'
     # We use mysql because it comes up without any other dependencies and its critical for the cluster
-    properties.router.dns_health_check_host: 'mysql-set.((KUBERNETES_NAMESPACE)).svc.((KUBERNETES_CLUSTER_DOMAIN))'
+    properties.router.dns_health_check_host: 'mysql-set((#KUBERNETES_NAMESPACE))/((KUBERNETES_NAMESPACE))'
```
jandubois:
```suggestion
    properties.router.dns_health_check_host: 'mysql-set((#KUBERNETES_NAMESPACE))((/KUBERNETES_NAMESPACE))'
```

Not sure what the point of an empty conditional is here though.

PR #1848, opened 2018-11-02
Patch diego cells so they respond to AZ

andreas-kupries:
Ref: https://trello.com/c/UXZXgy8V/843-patch-diego-cells-so-they-respond-to-az

- Script `configure-az.sh`
- Parameter `env.AZ_LABEL_NAME`

Trello has documentation attached, both how it works and how it was tested.

### DIFF
--- a/container-host-files/etc/scf/config/scripts/configure-az.sh
+++ b/container-host-files/etc/scf/config/scripts/configure-az.sh
@@ -2,8 +2,8 @@
 
 if test -z "${AZ_LABEL_NAME}"
 then
-    echo AZ: No label configured
-    echo AZ: Skipping
+    echo "AZ: No label configured"
+    echo "AZ: Skipping"
 else
     # AZ override processing is in effect.
 
@@ -23,8 +23,8 @@ else
 
     node="$($kubectl get pod $(hostname) -o jsonpath='{.spec.nodeName}')"
 
-    echo AZ: Configured ${AZ_LABEL_NAME}
-    echo AZ: Node...... ${node}
+    echo "AZ: Configured ${AZ_LABEL_NAME}"
+    echo "AZ: Node...... ${node}"
 
     # Determine the AZ of the kube worker node and make this
     # information available to the container, scripts, and binaries of
@@ -35,12 +35,12 @@ else
 
     if test -z "${NODE_AZ}"
     then
-	echo AZ: No information found
-	echo AZ: Skipping
+	echo "AZ: No information found"
+	echo "AZ: Skipping"
     else
 	# Propagate the found AZ information into cloudfoundry
 
-	echo AZ: Found..... ${NODE_AZ}
+	echo "AZ: Found..... ${NODE_AZ}"
         export KUBE_AZ="${NODE_AZ}"
     fi
 fi
--- a/container-host-files/etc/scf/config/scripts/configure-az.sh
+++ b/container-host-files/etc/scf/config/scripts/configure-az.sh
@@ -13,7 +13,7 @@ else
     # Determine the name of the kube worker node this container is
     # executing on.
 
-    node="$($kubectl get pod $(hostname) -o jsonpath='{.spec.nodeName}')"
+    node="$("${kubectl}" get pod "$(hostname)" -o jsonpath='{.spec.nodeName}')"
 
     echo "AZ: Configured ${AZ_LABEL_NAME}"
     echo "AZ: Node...... ${node}"
@@ -23,16 +23,16 @@ else
     # the diego-cell instance group.
 
     QUERY="jsonpath={.metadata.labels.${AZ_LABEL_NAME}}"
-    NODE_AZ=$($kubectl get node $node -o "${QUERY}" || true)
+    NODE_AZ=$("${kubectl}" get node "${node}" -o "${QUERY}" || true)
 
     if test -z "${NODE_AZ}"
     then
-	echo "AZ: No information found"
-	echo "AZ: Skipping"
+        echo "AZ: No information found"
+        echo "AZ: Skipping"
     else
-	# Propagate the found AZ information into cloudfoundry
+        # Propagate the found AZ information into cloudfoundry
 
-	echo "AZ: Found..... ${NODE_AZ}"
+        echo "AZ: Found..... ${NODE_AZ}"
         export KUBE_AZ="${NODE_AZ}"
     fi
 fi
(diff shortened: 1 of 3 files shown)


container-host-files/etc/scf/config/scripts/configure-az.sh
```diff
@@ -0,0 +1,46 @@
...
+    # AZ override processing is in effect.
+
+    # Locate the kubectl binary.
+    ##
+    # Attention! This only works because of us having/using a common
+    # layer which contains all the packages, regardless of their use
+    # within a specific role. This gives the diego-cell access to
+    # things it normally is not using at all. Here this is the
+    # `kubectl` cli originally added for use by the
+    # `acceptance-tests-brain`.
```
mook-as: That's… not awesome.

Can we make a dummy job in `scf-release` that pulls in the package, or something along those lines? So that there's an actual dependency to the correct package?

(I was originally thinking pre-start script, but this needs to run before `configgin` does, I think, and that's too late.)

We also make kube calls in [`configure-HA-hosts.sh`](https://github.com/SUSE/scf/blob/develop/container-host-files/etc/scf/config/scripts/configure-HA-hosts.sh), but that uses the HTTP API directly.  `curl` should be in the stemcell (so at least we don't get the magic action-at-a-distance thing).

Also, it looks like we can't use the [downwards API](https://kubernetes.io/docs/tasks/inject-data-application/environment-variable-expose-pod-information/) because that only exposes `nodeName` (not any of the labels on the node).

container-host-files/etc/scf/config/scripts/configure-az.sh
```diff
@@ -0,0 +1,46 @@
...
+    NODE_AZ=$($kubectl get node $node -o "${QUERY}" || true)
+
+    if test -z "${NODE_AZ}"
+    then
+	echo AZ: No information found
+	echo AZ: Skipping
+    else
+	# Propagate the found AZ information into cloudfoundry
+
+	echo AZ: Found..... ${NODE_AZ}
```
mook-as: For consistency, we should probably quote this

andreas-kupries: Sure.

container-host-files/etc/scf/config/scripts/configure-az.sh
andreas-kupries: Hm. ... A dummy job might actually work. Thank you. Will have a look.

jandubois: Also, on a general note we should not reference the `packages-src` directory in SCF at all. It is a fissile implementation detail that may already be broken by https://github.com/cloudfoundry-incubator/fissile/pull/439/files

It *moves* the directories from `packages-src` to `packages` to make them available to processes running under `bpm` control.

I'm already working on a PR that removes the 2 references we have to `packages-src` in SCF and UAA, so please don't create any new ones. 😄

andreas-kupries: Right. I wrote the big note as reminder that this would likely break the moment we use separate images per release ... Or something. ... Already broken, good enough to find something less of a hack, or no hack.

jandubois: The `cfdot` job in `diego-release` is an example of a job used to just install a binary (and update a few config files): https://github.com/cloudfoundry/diego-release/tree/develop/jobs/cfdot

andreas-kupries: ty for the reference.

container-host-files/etc/scf/config/scripts/configure-az.sh
andreas-kupries: Handled the easy thing with commit 917589c8c88d72da13b15d9b37cb691b7affd8e6.

container-host-files/etc/scf/config/scripts/configure-az.sh
andreas-kupries: Done with commit 8b7361f102776fb116cbce38b94475d58df2b6f8

container-host-files/etc/scf/config/scripts/configure-az.sh
```diff
@@ -0,0 +1,38 @@
...
+    NODE_AZ=$($kubectl get node $node -o "${QUERY}" || true)
+
+    if test -z "${NODE_AZ}"
+    then
+	echo "AZ: No information found"
+	echo "AZ: Skipping"
+    else
+	# Propagate the found AZ information into cloudfoundry
+
+	echo "AZ: Found..... ${NODE_AZ}"
```
mook-as: Nit: tab for indent here

src/scf-release/jobs/get-kubectl/spec
```diff
@@ -0,0 +1,11 @@
+---
+name: get-kubectl
+
+description: This job exists only to ensure the presence of the lubetcl binary in the role referencing it.
```
mook-as: Typo: `lubetcl`

container-host-files/etc/scf/config/scripts/configure-az.sh
```diff
@@ -0,0 +1,38 @@
...
+else
+    # AZ override processing is in effect.
+
+    # Locate the kubectl binary.
+    kubectl="/var/vcap/packages/kubectl/bin/kubectl"
+
+    # Determine the name of the kube worker node this container is
+    # executing on.
+
+    node="$($kubectl get pod $(hostname) -o jsonpath='{.spec.nodeName}')"
```
mook-as: Just realized we never quoted `$kubectl` and `$(hostname)` here…

container-host-files/etc/scf/config/scripts/configure-az.sh
andreas-kupries: Done with commit 158b9b51fe310632299747b6a4e80eef87d80457

src/scf-release/jobs/get-kubectl/spec
andreas-kupries: Done with commit 158b9b51fe310632299747b6a4e80eef87d80457

container-host-files/etc/scf/config/scripts/configure-az.sh
andreas-kupries: Done with commit 158b9b51fe310632299747b6a4e80eef87d80457

container-host-files/etc/scf/config/scripts/configure-az.sh
andreas-kupries: Handled the easy thing with commit 917589c8c88d72da13b15d9b37cb691b7affd8e6.

src/scf-release/jobs/get-kubectl/spec
andreas-kupries: And fully fixed with commit a381e6d3903a823a22499bf4fad50b3697b1c5eb

PR #1863, opened 2018-11-08
configure metron-agent as a colocated-container

qu1queee:
Hello,

This should cover the item https://www.pivotaltracker.com/story/show/160013733

Basically this includes:
- Moving the `loggregator-agent` as an `instance_group` of the type `colocated-container`
- Remove the `loggregator_agent` as an item of the `jobs` for all `instance_groups` using it, and replace it with the `properties.bosh_containerization.colocated_containers["loggregator-agent"]`

worked together with @HeavyWombat

### DIFF
--- a/container-host-files/etc/scf/config/role-manifest.yml
+++ b/container-host-files/etc/scf/config/role-manifest.yml
@@ -13,6 +13,8 @@ instance_groups:
     release: cf-syslog-drain
     properties:
       bosh_containerization:
+        colocated_containers:
+        - loggregator-agent
         run:
           scaling:
             min: 1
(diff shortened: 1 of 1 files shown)


viovanov:
@prabalsharma can you please take a look and see if this would break any of your scripts?
This will make it so that all pods have 2 containers now.

jandubois:
I'm not convinced that including `configure-HA-hosts.sh` in the metron role makes sense: as far as I can tell it is only used to set `KUBE_NATS_CLUSTER_IPS`; all other side-effects are not used anywhere.

But I'm going to merge this PR as-is once I have the ok from QA, and will then try to cleanup the `configure-HA-hosts.sh` stuff in a separate PR (it needs changes to the validator as well).

prabalsharma:
> @prabalsharma can you please take a look and see if this would break any of your scripts?
> This will make it so that all pods have 2 containers now.

Should be okay for `develop`. We will adjust our scripts if required.
thanks for pointing out

PR #1871, opened 2018-11-13
Remove configure-HA-hosts.sh from roles that don't need it

jandubois:
The script sets `KUBE_NATS_CLUSTER_IPS`. `KUBE_MYSQL_CLUSTER_IPS`, and `KUBE_CONSUL_CLUSTER_IPS`, but only the first one is still being used.

The validator was also checking for `KUBERNETES_CLUSTER_DOMAIN`, but that one isn't being set by this script, it is set by run.sh in the fissile stemcell.

The only remaining properties that require the script are:

```
diego.route_emitter.nats.machines: ((KUBE_NATS_CLUSTER_IPS))((#KUBE_SIZING_NATS_COUNT))((/KUBE_SIZING_NATS_COUNT))
nats.machines: ((KUBE_NATS_CLUSTER_IPS))((#KUBE_SIZING_NATS_COUNT))((/KUBE_SIZING_NATS_COUNT))
```

In addition to the obvious suspects, this is being used by all roles that use the route_registrar.

### DIFF
--- a/bin/config-validator.rb
+++ b/bin/config-validator.rb
@@ -146,7 +146,7 @@ def check_clustering(manifest, bosh_properties)
     block.call(lambda do |job|
       bosh_properties[job.release][job.name].each_key do |property|
         rparams.fetch("properties.#{property}", []).each do |param|
-          next unless /^(KUBERNETES_CLUSTER_DOMAIN|KUBE_.*_CLUSTER_IPS)$/ =~ param
+          next unless param == "KUBE_NATS_CLUSTER_IPS"
           collected_params[param] << "Job #{job.name.red} in release #{job.release.red}"
         end
       end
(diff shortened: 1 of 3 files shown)


bin/config-validator.rb
```diff
@@ -146,7 +146,7 @@ def check_clustering(manifest, bosh_properties)
     block.call(lambda do |job|
       bosh_properties[job.release][job.name].each_key do |property|
         rparams.fetch("properties.#{property}", []).each do |param|
-          next unless /^(KUBERNETES_CLUSTER_DOMAIN|KUBE_.*_CLUSTER_IPS)$/ =~ param
+          next unless param == "KUBE_NATS_CLUSTER_IPS"
```
mook-as: Silly rubocop warning: use `''` instead of `""` for constants.

PR #1874, opened 2018-11-15
autoscaler: move operator to new role

mook-as:
On some kube systems hairpin connections (going to the same container via the service) is not supported; as the operator connects to the scheduler and the scalingengine this can be problematic.  Move it to a new role that has nothing else to work around the issue.

The servicebroker also must not be co-located with apiserver.

The eventgenerator cannot be co-located with metricscollector.

@qibobo I'm picking where to put the jobs more-or-less randomly; please let me know if there are better choices. Unfortunately GitHub doesn't want to let me pick you as a reviewer…

### DIFF
--- a/container-host-files/etc/scf/config/role-manifest.yml
+++ b/container-host-files/etc/scf/config/role-manifest.yml
@@ -1330,17 +1330,13 @@ instance_groups:
             min: 0
             max: 3
             ha: 2
-          capabilities: []
           volumes:
           - path: /var/vcap/store
             type: persistent
             tag: postgres-data
             size: 100
-          shared-volumes: []
           memory: 1024
           virtual-cpus: 2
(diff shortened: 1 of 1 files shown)


qibobo:
as
@mook-as The picture above is the call-flow of the autoscaler components. So according to the picture, for consume the least containers, there should be 4 containers for autoscaler:
1 autoscaler_postgres
2 apiserver,operator
3 servicebroker,scheduler,eventgenerator
4 metricscollector,scalingengine

It is strange that we put some components  in a container because they are not similar in feature and it is not easy to give the role a name that can best describe its feature.
So maybe it's better that each component consumes a container?

mook-as:
Thanks! I guess then it's better for me to split out each job to a separate instance group.  I'll work on that, then.

PR #1907, opened 2018-12-04
role manifest: ask for SYS_RESOURCE caps where necessary

mook-as:
These instance groups include scripts which attempt to set `ulimit -n`, which requires `SYS_RESOURCE` to do.

### DIFF
--- a/container-host-files/etc/scf/config/role-manifest.yml
+++ b/container-host-files/etc/scf/config/role-manifest.yml
@@ -251,6 +251,7 @@ instance_groups:
                       values:
                       - nats
                   topologyKey: "beta.kubernetes.io/os"
+          service-account: withsysresource
         ports:
         - name: nats
           protocol: TCP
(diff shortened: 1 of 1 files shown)


jandubois:
The changes look fine to me, but SCF isn't starting on Jenkins:

```
04:14:32 api-group-0                     1/2       Running            0          52m
```

Not sure if this is related to the config change though, as even UAA took 10 minutes to get ready.

mook-as:
> `/usr/local/rvm/gems/ruby-2.3.1/gems/kubeclient-2.5.2/lib/kubeclient/common.rb:117:in `rescue in handle_exception': pods "diego-api-0" is forbidden: unable to validate against any pod security policy: [capabilities.add: Invalid value: "SYS_RESOURCE": capability may not be added] (KubeException)`

Yeah, this breaks the build, thanks for catching that

f0rmiga:
Rebased the branch. I will take over this PR.

PR #1911, opened 2018-12-05
Bump fissile to include manifest in role images:

jimmykarily:
https://github.com/cloudfoundry-incubator/fissile/commit/a19161d0d3cef0cc505f60c3d606f583eef3d8a1


thardeck:
Somehow it seems my download was wrong, because i got a different checksum. The original was ok, so I removed my commit regarding the checksum change.

jimmykarily:
Created PR to consume the final release in uaa-fissile-release repo as well: https://github.com/SUSE/uaa-fissile-release/pull/134

jimmykarily:
The problem was fixed and Jenkins is happy. We still have to define the process on how to make changes in the future but we could do that outside this PR. I think all comments are now addressed (in [uaa-fissile-release](https://github.com/SUSE/uaa-fissile-release/pull/134) as well except for the master vs develop issue).

PR #1912, opened 2018-12-05
Bump nfs-volume-release

andreas-kupries:
Ref https://trello.com/c/dOiHIcaP/847-5-update-nfs-volume-release-from-120-to-152-or-later

Blocked on CF bump. See last commit and its message, or the referenced ticket.


andreas-kupries:
@mook-as Ok. This looks good to me, and it passes Jenkins post-rebase, on the 2nd try. I would like to approvie, however I cannot, being the original author. As such I ask you to approve. At that point you will be able to merge to. Or throw it back to me and will then do that.

PR #1913, opened 2018-12-06
make: build *.tgz as well as *.zip for bundle archives

mook-as:


make/bundle-dist
```diff
@@ -19,20 +19,23 @@ else
...
 
 # "Am I Ok" for kube
 cp ${GIT_ROOT}/bin/dev/kube-ready-state-check.sh ${tmp_dir}/
 
 # Package the assembly. This directly places it into the output
-( cd ${tmp_dir} ; zip -r9 $ARCHIVE * )
+# For now, include both zip and tgz in order to transition smoothly
+cd "${tmp_dir}"
+tar czvf "${ARCHIVE_ROOT}.tgz" -- *
+zip -r9 "${ARCHIVE_ROOT}.zip" -- *
```
jandubois: Doesn't this include the `*.tgz` file inside the `*.zip`?

jandubois: Oops, I forgot that `ARCHIVE_ROOT` includes the full path and is outside the temp directory.

PR #1927, opened 2018-12-12
Refactored and fixed Kube ready state check script

f0rmiga:
## Description

Makes the check script more portable, as well as more robust since `ps aux` shows the full command line, while `pgrep` only shows the first 15 characters. Also, `pgrep` is not available on every system.

### DIFF
--- a/bin/dev/kube-ready-state-check.sh
+++ b/bin/dev/kube-ready-state-check.sh
@@ -25,14 +25,14 @@ shift $((OPTIND-1))
 category="${1:-all}"
 case ${category} in
     all|api|kube|node)
-	: # ok, nothing to do
+	: # OK, nothing to do.
 	;;
     *) usage
 	exit 1
 	;;
 esac
 
-#Script to determine is the K8s host is "ready" for cf deployment
+# Script to determine is the K8s host is "ready" for cf deployment.
 FAILED=0
 SCF_DOMAIN=${SCF_DOMAIN:-cf-dev.io}
 
@@ -66,7 +66,7 @@ function status() {
 }
 
 function having_category() {
-    # `all` matches always
+    # `all` matches always.
     set -- all "$@"
     case "$@" in
 	*${category}*)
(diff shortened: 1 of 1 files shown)


bin/dev/kube-ready-state-check.sh
```diff
@@ -78,62 +78,63 @@ function having_category() {
...
 if having_category kube ; then
     kubectl get pods --namespace=kube-system --selector name=tiller 2> /dev/null | grep -Eq '([0-9])/\1 *Running'
     status "tiller should be running (1/1 ready)"
 fi
 
-# ntp or systemd-timesyncd is installed and running
+# ntp or systemd-timesyncd is installed and running.
 if having_category api node ; then
-    pgrep -x ntpd >& /dev/null || pgrep -x chronyd >& /dev/null || systemctl is-active systemd-timesyncd >& /dev/null
+    (ps x | grep ntpd >& /dev/null) || (ps x | grep chronyd >& /dev/null) || systemctl is-active systemd-timesyncd >& /dev/null
```
mook-as: Oh wow, didn't realize `pgrep` just completely ignored the executable name after 15 characters.  But `grep` here means it will match commands with that substring (e.g. `/usr/sbin/ntpdate`).  And none of the things we're testing for are 15 characters long.

bin/dev/kube-ready-state-check.sh
```diff
@@ -78,62 +78,63 @@ function having_category() {
...
 
-# At least one storage class exists in K8s
+# At least one storage class exists in K8s.
 if having_category kube ; then
     test ! "$(kubectl get storageclasses 2>&1 | grep "No resources found.")"
     status "A storage class should exist in K8s"
 fi
 
-# privileged pods are enabled in K8s
+# Privileged pods are enabled in K8s.
```
mook-as: All of our command line argument checks are probably outdated, given that they seem to prefer using config files these days…

bin/dev/kube-ready-state-check.sh
```diff
@@ -78,62 +78,63 @@ function having_category() {
...
-    pgrep -ax 'hyperkube|kubelet' | grep kubelet | grep --silent -- --allow-privileged
+    ps ax | grep -E 'hyperkube|kubelet' | grep kubelet | grep -- --allow-privileged >& /dev/null
     status "Privileged must be enabled in 'kubelet'"
 fi
 
-# override tasks infinity in systemd configuration
+# Override tasks infinity in systemd configuration.
 if having_category node ; then
     if has_command systemctl ; then
+        # TODO(f0rmiga): Check, also, for infinity string. Reference https://github.com/moby/moby/blob/v17.04.0-ce/contrib/init/systemd/docker.service#L20-L22.
```
mook-as: Actually, no; systemd maps `infinity` to a big number (that's not actually infinite) :D

PR #1929, opened 2018-12-12
Dist: upload helm charts without separately

mook-as:
Helm can't install tgz charts if there are directories in the archive. This also means we can go back to uploading zip files for the release tool, since that needs to be manually extracted anyway.

### DIFF
--- a/make/bundle-dist
+++ b/make/bundle-dist
@@ -18,7 +18,7 @@ elif [[ ${FISSILE_STEMCELL} =~ ^.*fissile-stemcell-sle.*$ ]] ; then
     stemcell_os="sle"
     suffix=""
 else
-    echo "Unkown stemcell operating system: ${FISSILE_STEMCELL}"
+    echo "Unknown stemcell operating system: ${FISSILE_STEMCELL}"
     exit 1
 fi
 ARCHIVE_ROOT="${GIT_ROOT}/output/scf-${stemcell_os}-${APP_VERSION}"
--- a/make/bundle-dist
+++ b/make/bundle-dist
@@ -13,8 +13,10 @@ trap "rm -rf '${tmp_dir}'" EXIT
 
 if [[ ${FISSILE_STEMCELL} =~ ^.*fissile-stemcell-opensuse.*$ ]] ; then
     stemcell_os="opensuse"
+    suffix="-opensuse"
 elif [[ ${FISSILE_STEMCELL} =~ ^.*fissile-stemcell-sle.*$ ]] ; then
     stemcell_os="sle"
+    suffix=""
 else
     echo "Unkown stemcell operating system: ${FISSILE_STEMCELL}"
     exit 1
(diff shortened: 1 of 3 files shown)


andreas-kupries:
This is still marked WIP

make/bundle-dist
```diff
@@ -24,18 +26,19 @@ ARCHIVE_ROOT="${GIT_ROOT}/output/scf-${stemcell_os}-${APP_VERSION}"
 echo Packaging, taking "${APP_VERSION}" ...
 
 # Assembling the pieces ...
-mkdir -p ${tmp_dir}/kube ${tmp_dir}/helm
+find "${tmp_dir}"
```
andreas-kupries: This change seems to indicate the the tmpdir is expected to be present and that the coming commands create the subdirs ?

make/kube-dist
```diff
@@ -6,33 +6,31 @@ GIT_ROOT="${GIT_ROOT:-$(git rev-parse --show-toplevel)}"
...
-    # The directory should match the chart name.
-    if [[ $FISSILE_STEMCELL =~ ^.*fissile-stemcell-opensuse.*$ ]] ; then
-      suffix="-opensuse"
-    fi
+if [[ ${FISSILE_STEMCELL} =~ ^.*fissile-stemcell-opensuse.*$ ]] ; then
+    stemcell_os="opensuse"
+elif [[ ${FISSILE_STEMCELL} =~ ^.*fissile-stemcell-sle.*$ ]] ; then
+    stemcell_os="sle"
+else
+    echo "Unkown stemcell operating system: ${FISSILE_STEMCELL}"
```
andreas-kupries: Typo `Unkown` vs `Unk>n<own`.

make/bundle-dist
andreas-kupries: Or, in light of the overall comment, no subdirs anymore at all ?

mook-as: `mkdir -d` creates the `tmpdir` (in a way that should avoid racing with other processes).  So, yes. (And line 33 does indeed create the subdirs; because it's `mkdir -p`, `tmpdir` technically doesn't even need to exist yet.)

make/kube-dist
mook-as: Thanks! Fixed in 00f35b008877c3f9bdc9bd79db510aab33d7833e along with `make/kube-dist` where I copied this code from

mook-as:
Hmm, this doesn't seem to work; it looks like I actually do need a subdirectory to hold the charts in for things to actually deploy correctly.  The error message is just confusingly backwards.

> $ helm install --dry-run https://s3.amazonaws.com/cap-release-archives/prs/PR-1929-uaa-opensuse-helm-2.14.5%2Bcf2.7.0.134.g916f0668.tgz
Error: chart yaml not in base directory

make/bundle-dist
```diff
@@ -13,29 +13,32 @@ trap "rm -rf '${tmp_dir}'" EXIT
...
 
-# kube configs
-tar xf "${GIT_ROOT}/output/${ARTIFACT_NAME}-kube-${APP_VERSION}.tgz" -C "${tmp_dir}/kube"
-# helm charts
-tar xf "${GIT_ROOT}/output/${ARTIFACT_NAME}-helm-${APP_VERSION}.tgz" -C "${tmp_dir}/helm"
+for project in uaa cf ; do
+    mkdir -p "${tmp_dir}/kube/${project}${suffix}"
+    tar xf "${GIT_ROOT}/output/${project}-${stemcell_os}-kube-${APP_VERSION}.tgz" -C "${tmp_dir}/kube/${project}${suffix}/"
+    mkdir -p "${tmp_dir}/helm"
```
f0rmiga: Do we want to scope this as well to the `$project`? If not, this command can be executed outside the loop.

make/kube-dist
```diff
@@ -1,38 +1,47 @@
 #!/usr/bin/env bash
 
 set -o errexit -o nounset
+set -x
```
f0rmiga: Do we want this?

make/bundle-dist
mook-as: Fixed in bc206b25d8f09322bffb700b7ba72da8783a823c

make/kube-dist
mook-as: Nope; removed in bc206b25d8f09322bffb700b7ba72da8783a823c

f0rmiga:
@mook-as After running helm using the file from the last build, I got:
```
$ helm install --dry-run https://s3.amazonaws.com/cap-release-archives/prs/PR-1929-uaa-opensuse-kube-2.14.5%2Bcf2.7.0.136.ge618b589.tgz
Error: chart metadata (Chart.yaml) missing
```

PR #1942, opened 2018-12-18
matchExpression has to match component name (api-group)

berchtos:
The predefined podAntiAffinity rule does have an outdated component name (api), causing the matchExpression to be invalid.
Previously the api-group was named api, hence the values filter did still work.

https://github.com/SUSE/scf/blob/develop/container-host-files/etc/scf/config/role-manifest.yml#L648


berchtos:
@jandubois I did went through the others too, api-group was the only one I did spot.

jandubois:
> I did went through the others too, api-group was the only one I did spot.

Yeah, @f0rmiga confirmed that as well. 😄

Thanks for spotting this!

PR #1957, opened 2019-01-03
v3 API use HTTPS instead of HTTP

### LINKED ISSUE #1900: requests to v3 api over http instead of https
I'm working with CAP 1.2 (chart version 2.13.3) and looking forward to the new charts for 2.14.5. Maybe this issue is already solved in newer versions.

CF API delivers v3 urls with http:// instead of https://. We are using a loadbalancer in front of CF und don't wanna have a http-available API so the loadbalancer makes a redirect to https. CF cli doesn't like that at all and leads to errors like this:

```
> CF_TRACE=1 cf org system
Getting info for org system as admin...  
[...]  
REQUEST: [2018-11-29T13:48:51+01:00]  
GET /v3/isolation_segments?organization_guids=679703cb-33c1-4848-b66a-974cf01f21c6 HTTP/1.1  
[...]  
RESPONSE: [2018-11-29T13:48:51+01:00]  
HTTP/1.1 401 Unauthorized  
[...]  
Authentication error  
FAILED  
```
I direct api call /v3/isolation_segments does work because it goes over https. Inside calls like the one above the client makes a get request on /v3 to receive the v3 urls/paths which seems to be used afterwards:
```
> cf curl /v3  
[...]  
      "isolation_segments": {  
         "href": "http://api.my.domain/v3/isolation_segments"  
      },  
[...]  
```

f0rmiga:
## Description

Changes the URLs for the v3 API to use HTTPS instead of HTTP.

## Test plan

Calling `cf curl /v3` shall report the URLs as HTTPS instead of HTTP.

## Issue tracking

Fixes #1900

### DIFF
--- a/container-host-files/etc/scf/config/opinions-common.yml
+++ b/container-host-files/etc/scf/config/opinions-common.yml
@@ -165,7 +165,6 @@ properties:
       blobstore_type: webdav
       webdav_config:
         username: 'blobstore_user'
-    external_protocol: http
     install_buildpacks:
     - name: staticfile_buildpack
       package: staticfile-buildpack


f0rmiga:
[CC protocol should be https](https://trello.com/c/NM5rPN21/902-cc-protocol-should-be-https)

mook-as:
:cat: failure:
```

   Downloaded app package (171B)
   Traceback (most recent call last):
     File "/tmp/buildpacks/73ffc9805b0d23e30c9e8a7681b943bd/scripts/detect.py", line 23, in <module>
       .user_config(step='detect')
     File "/tmp/buildpacks/73ffc9805b0d23e30c9e8a7681b943bd/lib/build_pack_utils/builder.py", line 144, in find_composer_path
       self._detecter = self._config(ComposerJsonSearch(self._ctx))
     File "/tmp/buildpacks/73ffc9805b0d23e30c9e8a7681b943bd/lib/build_pack_utils/detecter.py", line 52, in __init__
       self.extension_module = utils.load_extension(extension_path)
     File "/tmp/buildpacks/73ffc9805b0d23e30c9e8a7681b943bd/lib/build_pack_utils/utils.py", line 55, in load_extension
       fromlist=['extension'])
     File "/tmp/buildpacks/73ffc9805b0d23e30c9e8a7681b943bd/extensions/composer/extension.py", line 34, in <module>
       from semver import max_satisfying
   ImportError: No module named semver
   None of the buildpacks detected a compatible application
```

Yeah that's probably not related to the change here

PR #1964, opened 2019-01-07
Add ingress kube definitions to scf chart.

andreas-kupries:
Ref: https://trello.com/c/jgev7lC4/892-3-fissile-work-to-use-an-ingress-controller

Fissile: https://github.com/cloudfoundry-incubator/fissile/pull/459
UAA: https://github.com/SUSE/uaa-fissile-release/pull/135


jandubois:
I've run multiple Jenkins builds for this PR, and while each one failed during CATS, the failures where different between the runs and not related to the PR changes afaict.

PR #1967, opened 2019-01-09
Set up default PSPs

mook-as:
This uses the changes from https://github.com/cloudfoundry-incubator/fissile/pull/460 to explicitly describe the PSPs we need, and the node-reader cluster role we need for diego-cell.  With this, we should not need to manually set up cluster roles & bindings.

### DIFF
--- a/bin/common/versions.sh
+++ b/bin/common/versions.sh
@@ -10,7 +10,7 @@ set -o errexit -o nounset
 export BOSH_CLI_VERSION="fcaa9c6caff58ab8da8c56481320681cdea492ee"
 export CFCLI_VERSION="6.37.0"
 export FISSILE_FLAVOR="develop"
-export FISSILE_VERSION="7.0.0+212.g74809e0"
+export FISSILE_VERSION="7.0.0+220.g4e5711b"
 export HELM_VERSION="2.11.0"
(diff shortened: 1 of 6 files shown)


bin/common/versions.sh
```diff
@@ -9,8 +9,8 @@ set -o errexit -o nounset
 
 export BOSH_CLI_VERSION="fcaa9c6caff58ab8da8c56481320681cdea492ee"
 export CFCLI_VERSION="6.37.0"
-export FISSILE_FLAVOR="develop"
-export FISSILE_VERSION="7.0.0+195.ga19161d"
+export FISSILE_FLAVOR="checks"
+export FISSILE_VERSION="7.0.0+210.g1aa92ee2"
```
andreas-kupries: One character too long. `ee` not `ee2`.

mook-as: Bumped version (because of fissile changes).

mook-as:
Jenkins is falling over on UAA secret rotation:
```
/usr/local/rvm/gems/ruby-2.3.1/gems/kubeclient-2.5.2/lib/kubeclient/common.rb:117:in `rescue in handle_exception': pods "secret-generation-2-dnfqv" is forbidden: unable to validate against any pod security policy: [] (KubeException)
```

PR #1969, opened 2019-01-09
Revert "Bump php-buildpack release to 4.3.67.1"

jandubois:
This reverts commit daa418614d486032062387128f6ef86344294954.


jandubois:
@thardeck Just making sure Tim sees this, to investigate what caused the CATS failure in the buildpack bump. The failing test is:

```
[Fail] [detect] Buildpacks php [It] makes the app reachable via its bound route 
```

thardeck:
Thanks,I suppose it is the first time that a buildpack bump breaks the cats. I will look into it.

PR #1972, opened 2019-01-10
Restoring DEFAULT_STACK to users.

andreas-kupries:
Ref: https://trello.com/c/W0YvSEfG/917-2-add-back-defaultstack-setting

- Removed the associated properties from the opinions.

- Role manifest
  - Restored the DEFAULT_STACK parameter exposing this setting.
  - Restored the property assignments using this parameter.

- Modified the fissile wrapper script to assemble a role manifest like
  it does for the light opinions. This brings in the proper
  `the-stack-name` which got selected via USE_SLE_BASE, and makes it
  referable from the `default` key of the `DEFAULT_STACK`
  parameter. We use the same stack-specific part files as for the
  light opinions.

  Note: The new final role manifest is a sibling of the original role
  manifest which is now serving as parts. Needed to not break the
  script paths.

- This way of handling the changed default value for `DEFAULT_STACK`
  works with the existing system (re-use of information, kept
  central), and with the format itself (yaml anchor and references,
  instead of brittle text-manipulation).

### DIFF
--- a/bin/fissile
+++ b/bin/fissile
@@ -19,6 +19,8 @@ if [ -f ${FISSILE_GIT_ROOT}/Jenkinsfile ] ; then
     fi
 
     tmp_file=$(mktemp -t tmp.XXXXXXXXXXXXXXXXXXXX)
+    trap "rm -f $tmp_file" EXIT
+
     cat "${STACK_BASE}" "${FISSILE_LIGHT_COMMON}" > "${tmp_file}"
     mv "${tmp_file}" "${FISSILE_LIGHT_OPINIONS}"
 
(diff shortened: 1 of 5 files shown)


bin/fissile
```diff
@@ -13,14 +13,21 @@ if [ -f ${FISSILE_GIT_ROOT}/Jenkinsfile ] ; then
...
 
     if [ "${USE_SLE_BASE}" = "false" ]; then
-        SRC_BASE="${FISSILE_LIGHT_OPEN42}"
+        STACK_BASE="${FISSILE_LIGHT_OPEN42}"
     else
-        SRC_BASE="${FISSILE_LIGHT_SLE12}"
+        STACK_BASE="${FISSILE_LIGHT_SLE12}"
     fi
 
     tmp_file=$(mktemp -t tmp.XXXXXXXXXXXXXXXXXXXX)
```
mook-as: Huh we never bothered deleting the temp file if things fail did we?  Should probably fix that while we're here. (The usual `trap … EXIT` is fine)

andreas-kupries: Like 
```
trap "rm -f $tmp_file" EXIT
```

andreas-kupries: ok. waiting actually, and me redoing images etc to check the SSG manifest thing Jan mentioned

PR #2002, opened 2019-01-21
Revert "Set up Istio in Vagrant VM"

f0rmiga:
Reverts SUSE/scf#1962


mook-as:
Just so we're clear, we're reverting this temporarily because we're cutting a release soon; this will go back in after that.

jandubois:
> Just so we're clear, we're reverting this temporarily because we're cutting a release soon; this will go back in after that.

FYI @zhanggbj

PR #2011, opened 2019-01-23
Make nfs-broker properly optional, i.e. not registered when off

andreas-kupries:
Ref: https://trello.com/c/4ceOnKG6/913-3-find-out-why-non-functional-cf-volume-services-appear-in-stratos

:warning: Do not merge. Intended to support debugging some other issue. :warning: 

Move broker registration from `configure-scf` job to new `register-nfs-persi`, as a post-start script. Run the new job as colocated container of the broker to separate the job and broker readiness. Running the new job in the same container as the broker will generate a dependency cycle preventing the pod from becoming ready

(   post-start waits on the endpoint to accept traffic before completing
 -> endpoint waits on pod to become ready before accepting traffic
 -> pod waits on post-start to complete before becoming ready
)

### DIFF
--- a/container-host-files/etc/scf/config/role-manifest.yml
+++ b/container-host-files/etc/scf/config/role-manifest.yml
@@ -1191,9 +1191,11 @@ instance_groups:
     release: nfs-volume
     properties:
       bosh_containerization:
+        colocated_containers:
+        - nfs-broker-registration
         run:
           scaling:
-            min: 1
+            min: 0
             max: 3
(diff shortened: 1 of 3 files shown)


prabalsharma:
@andreas-kupries I guess we don't wan't to merge this PR. In that case this should be closed.

andreas-kupries:
Agreed. With the ticket now done there is no reason to keep it open.

PR #2025, opened 2019-01-28
Role manifest: Remove diego-cell readiness probe

mook-as:
Turns out we didn't actually _have_ a probe script; it previously worked because the role manifest had an error and the readiness probe was never used.  Now that we fixed the typo, we try to make it use a non-existent readiness probe instead, which always fails.

See also #2015 (merged in 04f0c59c6d55151d2a440c11e9508d16f424d668) where we merged the typo fix (and broke the build).

See also https://github.com/SUSE/scf-helper-release/tree/master/jobs/global-properties/templates where there is no `diego-cell` readiness probe.

### DIFF
--- a/container-host-files/etc/scf/config/role-manifest.yml
+++ b/container-host-files/etc/scf/config/role-manifest.yml
@@ -1284,10 +1284,6 @@ instance_groups:
             tag: host-cgroup
           memory: 2800
           virtual-cpus: 4
-          healthcheck:
-            readiness:
-              command:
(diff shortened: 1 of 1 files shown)


qu1queee:
@mook-as so basically the `diego-cell` was never intended to execute an script for the readiness-probe?

mook-as:
@qu1queee Yes; or rather, at some point we wanted one, but never actually had one (I guess we realized it wasn't necessary)?

If there's a probe that we _should_ be executing, we would want to put that back (and revert this PR again).  But I can't find any evidence of having that probe in the first place.

PR #2027, opened 2019-01-28
acceptance-tests-brain: add tests for syslog forwarding

mook-as:
This enables the log forwarding in Jenkins, and adds a brains test to check that the forwarding works correctly (by adding a new file, and checking that a unique string is reported in a new container started by kubernetes).


bin/kube_overrides.rb
```diff
@@ -1,7 +1,11 @@
 require 'yaml'
 require 'json'
 
-namespace, domain, kube_config = ARGV.shift(3)
+release, kube_config = ARGV.shift(3)
```
andreas-kupries: Should be `shift(2)`, no ? (I will do the fix as part of the review)
(I noticed because I wanted to run only the new brain test, and the INCLUDE setting I provided got lost)

mook-as: Correct; sorry about that!

bin/kube_overrides.rb
```diff
@@ -30,30 +34,20 @@
...
-
-            env['value'] = value.to_s
-          end
-
-          if env['valueFrom'] && env['valueFrom']['secretKeyRef']
-            name = env['name'].downcase.gsub('_', '-')
-            if generated.has_key?(name) && (secrets[name].nil? || secrets[name].empty?)
-              env['valueFrom']['secretKeyRef']['name'] = current_secrets_name
+          if env['valueFrom']
+            # Download API; check if it's from a secret
```
mook-as: Typo: `Downward API`

andreas-kupries: Will change

src/scf-release/src/acceptance-tests-brain/test-scripts/016_syslog_forwarding_test.rb
```diff
@@ -0,0 +1,124 @@
...
+    def status(fmt, *args)
+        printf "\n\e[0;32m#{fmt}\e[0m\n", *args
+    end
+
+    # Report problem to the user; use as printf
+    def trouble(fmt, *args)
+        printf "\n\e[0;31m#{fmt}\e[0m\n", *args
+    end
+
+    $RUN_SUFFIX = SecureRandom.hex(16)
```
andreas-kupries: I might reduce this to 8. The hex doubles the output, making it 32 currently.

The last jenkins reported

```
15:46:43 + kubectl run 11280387a7c84f769917f9c5b1214b90 --namespace pr-2027-5-scf --command --port 514 --expose --image=splatform/scf-acceptance-tests-brain:4dae05e3949d150f2acddeff10caf8a5edbaf226 --labels=brains=11280387a7c84f769917f9c5b1214b90.1f4e935891d083b1bbf46adb7cfaead4 -- /bin/sh -c zypper --non-interactive install /usr/bin/socat /usr/bin/logger && /usr/bin/socat TCP-LISTEN:514,fork 'EXEC:/usr/bin/logger --socket-errors=off --stderr --tag ""'
15:46:43 The Deployment "11280387a7c84f769917f9c5b1214b90" is invalid: 
15:46:43 * metadata.labels: Invalid value: "11280387a7c84f769917f9c5b1214b90.1f4e935891d083b1bbf46adb7cfaead4": must be no more than 63 characters
15:46:43 * spec.selector.matchLabels: Invalid value: "11280387a7c84f769917f9c5b1214b90.1f4e935891d083b1bbf46adb7cfaead4": must be no more than 63 characters
15:46:43 * spec.selector: Invalid value: v1.LabelSelector{MatchLabels:map[string]string{"brains":"11280387a7c84f769917f9c5b1214b90.1f4e935891d083b1bbf46adb7cfaead4"}, MatchExpressions:[]v1.LabelSelectorRequirement(nil)}: invalid label selector.
```

I get a failing kube command under vagrant as well somewhere, but later I think. It still has this long hex value in names, and I worry that this is the problem.

If not I will let jenkins run on the reduced form to see if that then matches what I see.

andreas-kupries: Yep. uuidgen yields 32 characters, plus 32 char suffix, plus dots => Too long.
My reduction of the suffix made that pass.
Still a problem, uuidgen output may start with a digit. The service name must start with an alphabetical char.

Pushed a fix, also reduced the random part to 16 chars too, to keep under the limits.

andreas-kupries:
@mook-as The brain test failure is minibroker mongodb, i.e. unrelated. The syslog test passes.
I am going with :white_check_mark: and will merge.

PR #2033, opened 2019-01-29
Added provisioning for loading the loop kernel module

f0rmiga:
## Description

Loads the loop kernel module on system startup, as well as during the Vagrant machine provisioning.
It prevents diego-cell-0 from crashing for the first time it tries to perform a mount using a file as a filesystem.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -89,6 +89,12 @@ Vagrant.configure(2) do |config|
     end
   end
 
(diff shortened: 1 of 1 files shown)


f0rmiga:
[When GARDEN_ROOTFS_DRIVER is btrfs, wait for the filesystem to be ready before calling mount](https://trello.com/c/V0KW1qCC/921-when-gardenrootfsdriver-is-btrfs-wait-for-the-filesystem-to-be-ready-before-calling-mount)

PR #2035, opened 2019-01-29
acceptance-tests-brain: 012_minibroker_redis: fix image pulling

mook-as:
The minibroker charts changed how the images are fetched; fix up the test scripts so we can find them again (by overriding the appropriate bits in the helm values).

Also, dump out relevant information when tests fail.


mook-as:
Merged a bit early because it already got past the relevant changes. Unbreaking `develop`.

andreas-kupries:
Checking ... Confirmed. Ok.

PR #2019, opened 2019-01-25
Add cflinuxfs3 to scf

lightlace:
supersedes #2013

PR #2038, opened 2019-01-30
Add cflinuxfs3 stack to scf.

lightlace:
supersedes #2019.


jimmykarily:
@lightlace we need to check how much the base image is going to increase with the additional stack and decide wether we merge or not. I'm assuming it's not too big of a difference but let's see.

jimmykarily:
I compared the size of the scf-api-group image between the 2.14.5 release and this branch (as reported by `docker inspect`)

the released image is `7.83G` in size and this branch produces an image of `8.27G` (or `8.41G` vs `8.88G` as reported by `docker images`). So we have a ~400Mb increase although cflinuxfs3 image is ~1G but there are more differences between the 2 commits that might affect the image size.

jimmykarily:
https://trello.com/c/ULH9IPEo/868-p9-3-cflinuxfs3

PR #2039, opened 2019-01-30
Enable mutual TLS between Cloud Controller and GoRouter

f0rmiga:
## Description

Use port 9024 instead of 9022 on CC for mutual TLS with GoRouter.

Reference https://bosh.io/jobs/cloud_controller_ng?source=github.com/cloudfoundry/capi-release&version=1.66.0#p%3dcc.public_tls

## Test plan

Deploy SCF and perform an operation with the cf CLI. Then get a shell session to the `router-0` and `cat /var/vcap/sys/log/gorouter/access.log`. It shall show it is using port 9024. E.g.:

```
api.cf-dev.io - [2019-03-05T03:33:33.282+0000] "GET /v2/jobs/... HTTP/1.1" 200 0 270 "-" "go-cli 6.40.0+07673feb9.2018-10-08 / linux" "172.16.0.1:56432" "172.16.1.182:9024" x_forwarded_for:"172.16.0.1" x_forwarded_proto:"https" vcap_request_id:"..." response_time:0.01207756 app_id:"-" app_index:"-"
```

### DIFF
--- a/container-host-files/etc/scf/config/role-manifest.yml
+++ b/container-host-files/etc/scf/config/role-manifest.yml
@@ -789,6 +789,9 @@ instance_groups:
         - name: api-tls
           protocol: TCP
           internal: 9023
+        - name: api-mutual-tls
+          protocol: TCP
+          internal: 9024
         - name: statsd
           protocol: TCP
           internal: 8125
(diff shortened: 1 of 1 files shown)


f0rmiga:
[(3) Enable mTLS between CC and gorouter](https://trello.com/c/epPCj29Z/927-3-enable-mtls-between-cc-and-gorouter)

f0rmiga:
@mook-as Apparently, `cc.mutual_tls` is for mutual TLS with other components. CC <-> GoRouter has its own cert as we can see here https://github.com/cloudfoundry/capi-release/commit/2093779934dd2915632bef7cd66b23dd48cfbef5. Although this commit introduces the `cc.cc_gorouter_tls`, we can see here https://github.com/cloudfoundry/capi-release/commit/5234dc5911deeb143eb6bd1e9ac108938721b186 that it was renamed from `cc.cc_gorouter_tls` to `public_tls`. Maybe @ericpromislow can illuminate us if I'm wrong, since he was part of this as we can see from the first referenced commit.

PR #2049, opened 2019-02-04
scf-release/acceptance-tests-brain: new minibroker mariadb test

mook-as:
Adds new brain test for minibroker mariadb.

Note that newer versions of the mariadb chart has changed the format for specifying parameters (in this case, only the database name is relevant).  To support the older version of minibroker we have, though, we need to specify it twice, once the old way and once the new way.


src/scf-release/src/acceptance-tests-brain/test-scripts/014_minibroker_mariadb_test.sh
```diff
@@ -0,0 +1,197 @@
...
+
+## # # ## ### Test-specific code ### ## # #
+## For custom cleanup retrap the signals EXIT & ERR to run a custom
+## function, and chain to login_cleanup inside. Remove if not needed.
+
+# We have some waits in this test, and want to clean things up properly
+# (especially the service broker) when we abort.  So in our waits we check that
+# nine minutes haven't elapsed; this gives us around a minute to do any clean up
+# we need.
+DEADLINE="$(date --date=9min +%s)"
```
f0rmiga: Let's make `9min` a const that is easier to control. If you feel like it should also be shared between other tests, let's move to a shared script. Otherwise, `check_deadline` should take it as a parameter.

PR #2054, opened 2019-02-06
Update cflinuxfs stacks to 0.263.0 and 0.57.0

thardeck:


thardeck:
We decided only to use the version officially released on github, not the latest bosh.io. Closing

PR #2063, opened 2019-02-06
acceptance-tests-brain: refactor test scripts

mook-as:
This converts the test scripts into ruby, so that code sharing is easier and hopefully easier to understand.


f0rmiga:
BRAIN tests pass. Merging.

PR #2064, opened 2019-02-07
Istio on Vagrant

f0rmiga:
## Description

- Brings back https://github.com/SUSE/scf/commit/27486c2dcff7359d5ac4c099aa27d0f19d547711.
- Adds minor changes.
- Adds missing stop.

## Test plan

On a Vagrant machine, run:

- `make istio-run`.
- Check that Istio was installed on k8s.
- `make istio-stop`.
- Check that Istio was uninstalled from k8s.
- Check that `make istio-stop` is idempotent.
- Check that `make istio-run` will perform an upgrade if Istio is already installed.
- `cat ~/.profile` and check for the exported `$PATH` with the Istio bin location.
- Change the version of Istio on `//bin/common/versions.sh` to `1.0.4` and run `make istio-run`. Notice it performs the download of the version and upgrades the installation.
- `cat ~/.profile` and check that `$PATH` with the Istio bin location was updated.


f0rmiga:
[Add Istio back to the Vagrant machine](https://trello.com/c/65SXpAD1/939-add-istio-back-to-the-vagrant-machine)

make/istio/run
```diff
@@ -0,0 +1,42 @@
...
+istio_path="${GIT_ROOT}/.istio/${ISTIO_VERSION}"
+if [ ! -d "$istio_path" ]; then
+  mkdir -p "${istio_path}"
+  curl -L "https://github.com/istio/istio/releases/download/${ISTIO_VERSION}/istio-${ISTIO_VERSION}-linux.tar.gz" \
+    | tar zx -C "${istio_path}" --strip-components=1
+fi
+
+path_with_istio="PATH=\"${istio_path}/bin:\${PATH}\""
+export_path_comment="# Add Istio binaries to PATH."
+if ! grep --silent "${export_path_comment}" ~/.profile 2> /dev/null; then
```
mook-as: Hmm, I am not sure I'm entirely comfortable with this modification of user data.
You can actually run `make/run`, etc. on your workstation (pointing at a kube), and things will work. So modifying `~/.profile` here, instead of when building vagrant (or at least, `vagrant up`), seems icky.

make/istio/stop
```diff
@@ -0,0 +1,24 @@
...
+source "${GIT_ROOT}/make/include/defaults"
+
+printf "Stopping and cleaning up %s release...\\n" "${ISTIO_HELM_RELEASE}"
+
+if has_release "${ISTIO_HELM_RELEASE}"; then
+  helm delete --purge "${ISTIO_HELM_RELEASE}"
+fi
+
+if has_namespace "${ISTIO_NAMESPACE}"; then
+  kubectl delete namespace "${ISTIO_NAMESPACE}"
```
mook-as: Do we need to wait for the NS to actually go away? We usually do elsewhere right?

make/istio/stop
```diff
@@ -0,0 +1,24 @@
...
+fi
+
+# Delete Custom Resource Definitions.
+custom_resource_definitions=$(kubectl get customresourcedefinitions -o=go-template='{{range .items}}{{.metadata.name}}{{"\n"}}{{end}}' | grep "istio.io" | cat)
```
mook-as: I don't think we need the `| cat` at the end?
(I also find it annoying that there isn't a built-in `strings.Contains` function)

make/istio/stop
```diff
@@ -0,0 +1,24 @@
...
+if [ -n "${custom_resource_definitions}" ]; then
+  echo "${custom_resource_definitions}" | xargs kubectl delete customresourcedefinitions
```
mook-as: A.k.a. `kubectl delete customresourcedefinitions ${custom_resource_definitions}`
Or if you really like `xargs`, `echo ${custom_resource_definitions} | xargs --no-run-if-empty kubectl delete customresourcedefinitions`

make/istio/run
```diff
@@ -0,0 +1,42 @@
+#!/usr/bin/env bash
+
+set -o errexit -o nounset
+
+: "${GIT_ROOT:=$(git rev-parse --show-toplevel)}"
+source "${GIT_ROOT}/bin/common/versions.sh"
+source "${GIT_ROOT}/make/include/defaults"
+
+# Download and install the Istio release.
+istio_path="${GIT_ROOT}/.istio/${ISTIO_VERSION}"
```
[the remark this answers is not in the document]
f0rmiga: Well... if you destroy a Vagrant vm and bring a new one up, if you run `make istio-run` it will check for the version and prevent from downloading again the same version. If we "unversion", it's going to be dumb in the sense that whatever version is installed inside `//output/istio` is going to be picked.

bin/common/install_tools.sh
```diff
@@ -53,8 +53,17 @@ if systemctl list-unit-files kube-apiserver.service | grep --quiet enabled ; the
...
-  echo "Installing tiller for helm ..."
-  ${do_as_vagrant} helm init
+  echo "Installing tiller for helm with service account ..."
+  tiller_sa_namespace="kube-system"
+  tiller_sa_name="tiller"
+  if ! kubectl get serviceaccount "${tiller_sa_name}" --namespace "${tiller_sa_namespace}" 1> /dev/null 2> /dev/null; then
+    kubectl create serviceaccount "${tiller_sa_name}" --namespace "${tiller_sa_namespace}"
+  fi
+  tiller_crb_name="tiller"
+  if ! kubectl get clusterrolebinding "${tiller_crb_name}" --namespace "${tiller_sa_namespace}" 1> /dev/null 2> /dev/null; then
```
mook-as: I don't think cluster role bindings are within a namespace

make/istio/run
```diff
@@ -0,0 +1,39 @@
...
+  --set "gateways.istio-ingressgateway.type=NodePort"
+  --set "gateways.istio-egressgateway.enabled=false"
+  --set "tracing.enabled=true"
+  --set "grafana.enabled=true"
+)
+
+helm upgrade --install "${args[@]}" "${chart}" "${values[@]}"
+
+path_with_istio="PATH=\"${istio_path}/bin:\${PATH}\""
+printf "Done. You might want to 'export %s' to get Istio binaries onto your \$PATH.\\n" "${path_with_istio}"
```
mook-as: Maybe add it in [`.envrc`](https://github.com/SUSE/scf/blob/340b8e6c511a59394110d2ba5409da12dc3f23e1/.envrc#L71) too? (With the version number in there)

make/istio/stop
```diff
@@ -0,0 +1,22 @@
...
+  helm delete --purge "${ISTIO_HELM_RELEASE}"
+fi
+
+if has_namespace "${ISTIO_NAMESPACE}"; then
+  kubectl delete namespace "${ISTIO_NAMESPACE}"
+fi
+
+# Delete Custom Resource Definitions.
+custom_resource_definitions=$(kubectl get customresourcedefinitions -o=go-template='{{range .items}}{{.metadata.name}}{{"\n"}}{{end}}' | grep "istio.io" || true)
+echo "${custom_resource_definitions}" | xargs --no-run-if-empty kubectl delete customresourcedefinitions
```
mook-as: Not necessary, just thought it might be nice:
```bash
kubectl get customresourcedefinitions -o=go-template='{{range .items}}{{.metadata.name}}{{"\n"}}{{end}}' \
  | grep "istio.io" \
  | xargs --no-run-if-empty kubectl delete customresourcedefinitions
```

bin/common/install_tools.sh
f0rmiga: Right. Why would it!?

make/istio/stop
f0rmiga: After the changes I made, I couldn't see it at the end. :)

PR #2074, opened 2019-02-07
Fix names for buildpack releases

jandubois:
The `name` and `version` keys are currently not used by fissile; those values are extracted from `release.MF` for final releases. The values should at least be validated though, but for now this commit just fixes the names of the buildpack releases.


jimmykarily:
We will have to change these values as well: https://github.com/SUSE/cloudfoundry/blob/master/ci/pipelines/final-releases/pipeline.yml.erb#L14

and we should keep an eye on all pipelines in case we missed something else.

thardeck:
@jimmykarily I have fixed the pipeline accordingly.

PR #2079, opened 2019-02-08
Small brain linting cleanup and in-order execution

f0rmiga:


src/scf-release/jobs/acceptance-tests-brain/templates/run.erb
```diff
@@ -49,7 +49,7 @@ for i in "${ASSETS}/plugins/"*; do
 done
 shopt -u nullglob
 
-PARAM="--timeout 600 --verbose"
+PARAM="--timeout 600 --verbose --in-order"
```
mook-as: We probably shouldn't trigger `--in-order` by default. After all, we wrote the test brain runner, and intentionally randomized the test order to catch issues with test dependencies…

f0rmiga:
> I have no idea why `wait_for_namespace2` exists

It was waiting for the namespace number two!?

src/scf-release/src/acceptance-tests-brain/test-scripts/012_minibroker_redis_test.rb
```diff
@@ -73,7 +55,7 @@ def wait_for_namespace2(namespace)
 
     run "kubectl get namespace #{MINIBROKER_NAMESPACE} || kubectl create namespace #{MINIBROKER_NAMESPACE}"
     run "helm init --client-only"
-    run *(%Q(helm upgrade #{HELM_RELEASE} minibroker
+    run(*(%Q(helm upgrade #{HELM_RELEASE} minibroker
```
mook-as: Hmm, I don't know why I used `%Q().split` instead of `%W()`.

Issue #2081, opened 2019-02-09
Lack of permission definitions for Istio automatic sidecar injection


zhanggbj:
If you try Istio automatic sidecar injection following the instructions here ->https://istio.io/docs/setup/kubernetes/sidecar-injection/, after the restart, pod cannot be created, there will be an error like below

```
$ kubectl get events
2m32s       Warning   FailedCreate              ReplicaSet   (combined from similar events): Error creating: pods "sleep-68648fd5b5-w5vh8" is forbidden: unable to validate against any pod security policy: [spec.initContainers[0].securityContext.privileged: Invalid value: true: Privileged containers are not allowed spec.initContainers[0].securityContext.capabilities.add: Invalid value: "NET_ADMIN": capability may not be added spec.containers[0].securityContext.allowPrivilegeEscalation: Invalid value: true: Allowing privilege escalation for containers is not allowed]
```

zhanggbj:
Istio needs to inject sidecar into the original deployment. And the sidecar requires extra permissions to `run privileged container` and `add Linux capability – NET_ADMIN`.

zhanggbj:
This issue still exists and @f0rmiga is helping investigation. Thanks.
Debugging details:
https://github.com/SUSE/scf/pull/2082#issuecomment-514922803

f0rmiga:
Closing due to inactivity.

PR #2082, opened 2019-02-09
Add permission definitions for Istio automatic sidecar injection

zhanggbj:
[2081](https://github.com/SUSE/scf/issues/2081)


zhanggbj:
Istio needs to inject sidecar into the original deployment. And the sidecar requires extra permissions to `run privileged container` and `add Linux capability – NET_ADMIN`.

zhanggbj:
Hi @viovanov @mook-as @f0rmiga ,

This is a fix for Istio sidecar injection, would you please help take a review? Thanks!

make/assets/istio/clusterrole.yaml
```diff
@@ -0,0 +1,14 @@
+---
```
f0rmiga: I know we already have the `make/assets` directory, but in this case I think it is better to have those files under `make/istio/assets`. I may eventually move the contents of `make/assets` to a more meaningful directory structure.

make/istio/run
```diff
@@ -6,6 +6,10 @@ set -o errexit -o nounset
 source "${GIT_ROOT}/bin/common/versions.sh"
 source "${GIT_ROOT}/make/include/defaults"
 
+# Add permissions to support Istio automatic sidecar injection.
+SC="${GIT_ROOT}/make/assets/istio"
```
f0rmiga: `SC` is not very clear. Maybe after moving to `make/istio/assets`, you could name this `assets="${GIT_ROOT}/make/istio/assets"`. Then the next line becomes `kubectl create -f "${assets}"`, which is more legible IMO.

make/istio/stop
```diff
@@ -21,3 +21,16 @@ fi
 kubectl get customresourcedefinitions -o=go-template='{{range .items}}{{.metadata.name}}{{"\n"}}{{end}}' \
   | grep "istio.io" \
   | xargs --no-run-if-empty kubectl delete customresourcedefinitions
+
+# Delete permission definitions.
+kubectl get podsecuritypolicy -o=go-template='{{range .items}}{{.metadata.name}}{{"\n"}}{{end}}' \
+  | grep "istio" \
+  | xargs --no-run-if-empty kubectl delete podsecuritypolicy
+
+kubectl get clusterrole -o=go-template='{{range .items}}{{.metadata.name}}{{"\n"}}{{end}}' \
```
mook-as: Don't need to do this here, but I just realized that all of this is easier as:
```bash
kubectl get CustomResourceDefinitions,PodSecurityPolicy,ClusterRole,ClusterRoleBinding -o name \
  | grep istio \
  | xargs --no-run-if-empty kubectl delete
```

(Of course this only works if the grep works for all of the resource types)

make/assets/istio/clusterrole.yaml
zhanggbj: Sure, so I will move to `make/istio/assets` in next patch. Thanks!

make/istio/run
zhanggbj: It make sense:-) I'll change to `istio_assets`in next patch.

zhanggbj:
Hi @mook-as
Good catch. This is incorrect here. I want to make every pod in any namespace have the permission to modify iptables for istio-init container. I changed it to ServiceAccount `default`. Does it make sense? Thanks!

zhanggbj:
@f0rmiga @mook-as 
I've addressed the review comments. Would you please help take a look? Thanks!

zhanggbj:
@mook-as 
Thanks for your review! Honestly I'm not so familar with the permission settings. I took some investigation and resubmit the PR today, would it be better if using `apiGroup` instead of `serviceAccount` here? Thanks!

mook-as:
@zhanggbj 
That would at least let all the system accounts bind to the `istio`, role, yes. I'm not sure that helps though; it means all service accounts can use the `istio` PSP, but still only one PSP will be chosen according to the [documented order](https://kubernetes.io/docs/concepts/policy/pod-security-policy/#policy-order).  And this doesn't fundamentally resolve the bigger concern about allowing too much privileges to everything…

f0rmiga:
@zhanggbj We have a story to track this in our queue.

zhanggbj:
Hi @f0rmiga,
Sorry for late response. Would you please help share the link, I didn't find it in the tracker for CF Containerization.
I was distracted by internal staff last month, and will also continue working on it this week :-)

jandubois:
@zhangtbj 

I've been looking into this for a while today to find a proper solution, but everything seemed rather complicated. But then I noticed that ever since we moved to cf-deployment 6.x, we use privileged service accounts for all instance groups anyways (because jobs are now managed by `bpm`):

```
$ grep serviceAccount * | awk '{print $3}' | sort | uniq
"default-privileged"
"eirini"
"eirini-privileged"
"garden-runc"
"privileged"
"secret-generator"
"withsysresource-privileged"
```

So if you rebase your istio branch on top of the latest `develop` branch, maybe you don't need any permissions changes and everything works automatically?  Please try and let me know!

zhanggbj:
hi @jandubois @f0rmiga 
I deployed latest scf(develop b5cf613) today but didn't get any luck, the problem still exists.

After scf deployment done, I tried to inject istio sidecar for uaa-0 pod.
```
29s    Warning   FailedCreate       StatefulSet   create Pod uaa-0 in StatefulSet uaa failed error: pods "uaa-0" is forbidden: unable to validate against any pod security policy: [spec.initContainers[0].securityContext.privileged: Invalid value: true: Privileged containers are not allowed spec.initContainers[0].securityContext.capabilities.add: Invalid value: "NET_ADMIN": capability may not be added spec.initContainers[0].securityContext.privileged: Invalid value: true: Privileged containers are not allowed spec.initContainers[0].securityContext.capabilities.add: Invalid value: "NET_ADMIN": capability may not be added spec.initContainers[0].securityContext.privileged: Invalid value: true: Privileged containers are not allowed spec.initContainers[0].securityContext.capabilities.add: Invalid value: "NET_ADMIN": capability may not be added]
```

jandubois:
> I tried to inject istio sidecar for uaa-0 pod.

Ah, I see; the UAA role manifest doesn't use privileged containers because UAA doesn't use BPM(yet?).

So I would suggest that you copy the corresponding privileged accounts and PSPs from the SCF role manifest into the UAA one and use those for the instance groups that are `istio_managed`. Given that almost all SCF roles are now privileged I don't think there should be much pushback to doing this for UAA as well.

I've created an (internal) card today to review the while service account setup for SCF because it seems too complex and unwieldy right now. So we may want to change this again after that review, but for now we should be fine just adding the privileged SA to UAA.

zhanggbj:
Hi @jandubois @f0rmiga ,

I checked `api-group-0` was set with serviceAccount `default-privileged`, this is expected, right?
But when I tried to inject Istio sidecar, it also report the same error like below
```
30m     Warning   FailedCreate       StatefulSet   create Pod api-group-0 in StatefulSet api-group failed error: pods "api-group-0" is forbidden: unable to validate against any pod security policy: [spec.initContainers[0].securityContext.privileged: Invalid value: true: Privileged containers are not allowed spec.initContainers[0].securityContext.capabilities.add: Invalid value: "NET_ADMIN": capability may not be added spec.containers[0].securityContext.privileged: Invalid value: true: Privileged containers are not allowed spec.containers[1].securityContext.privileged: Invalid value: true: Privileged containers are not allowed spec.initContainers[0].securityContext.privileged: Invalid value: true: Privileged containers are not allowed 
```

Recreate steps
1. scf is up
2. install Istio by `make istio-run`
3. label cf namespace `kubectl label namespace cf istio-injection=enabled`
4. delete api-group-0 pod and wait recreate.
5. check K8s log `kubectl get event -n cf`

zhanggbj:
Just an update. Detailed discussion about scf permissions is at https://github.com/cloudfoundry-incubator/fissile/wiki/ServiceAccount-and-PodSecurityPolicy-usage.

jandubois:
> Just an update. Detailed discussion about scf permissions is at https://github.com/cloudfoundry-incubator/fissile/wiki/ServiceAccount-and-PodSecurityPolicy-usage.

These changes have been implemented in fissile and SCF now; is this PR still necessary, or does istio works as expected out-of-the-box now?

PS: Is there some document somewhere about running SCF with istio, and what you can do when you have it enabled?

zhanggbj:
Hi @jandubois,
Great! So the PSP is exactly the same with here now? https://github.com/cloudfoundry-incubator/fissile/wiki/ServiceAccount-and-PodSecurityPolicy-usage.

I will try it again, with the right PSP then we can inject Istio envoy proxy sidecar to cf component pod and use Istio to manage it. There's no doc yet, if this part works fine, I'll raise PR to add some docs
. And if you're interested, here is some demo I did to use Istio to manage CFEE. 

Thanks a lot!

zhanggbj:
Hi @jandubois, 

I tried with the latest scf, looks like the issue still exists. When I tried to inject an Istio sidecar for api-group-0, the error is 
```
8m4s        Warning   FailedCreate   StatefulSet   create Pod api-group-0 in StatefulSet api-group failed error: pods "api-group-0" is forbidden: unable to validate against any pod security policy: [spec.initContainers[0].securityContext.capabilities.add: Invalid value: "NET_ADMIN": capability may not be added spec.containers[0].securityContext.privileged: Invalid value: true: Privileged containers are not allowed spec.containers[1].securityContext.privileged: Invalid value: true: Privileged containers are not allowed spec.initContainers[0].securityContext.capabilities.add: Invalid value: "NET_ADMIN": capability may not be added spec.containers[0].securityContext.privileged: Invalid value: true: Privileged containers are not allowed spec.containers[0].securityContext.allowPrivilegeEscalation: Invalid value: true: Allowing privilege escalation for containers is not allowed spec.containers[1].securityContext.privileged: Invalid value: true: Privileged containers are not allowed spec.containers[1].securityContext.allowPrivilegeEscalation: Invalid value: true: Allowing privilege escalation for containers is not allowed]
```

I took a look at the pod definition, so far the pod is using `kubernetes.io/psp: cf-psp-default` and it already has `privileged: true` and `allowedCapabilities:*`as below, so not sure why this happened, any thoughts? Thanks!
```
 k get psp --all-namespaces cf-psp-default -o yaml
kubectl get psp --all-namespaces cf-psp-default -o yaml
apiVersion: extensions/v1beta1
kind: PodSecurityPolicy
metadata:
  creationTimestamp: "2019-07-24T08:48:47Z"
  labels:
    app.kubernetes.io/component: cf-psp-default
    app.kubernetes.io/instance: cf
    app.kubernetes.io/managed-by: Tiller
    app.kubernetes.io/name: cf
    app.kubernetes.io/version: 1.4.1
    helm.sh/chart: cf-2.17.1
    skiff-role-name: cf-psp-default
  name: cf-psp-default
  resourceVersion: "10730"
  selfLink: /apis/extensions/v1beta1/podsecuritypolicies/cf-psp-default
  uid: da0b8209-adef-11e9-8db7-08002736ab9d
spec:
  allowPrivilegeEscalation: true
  allowedCapabilities:
  - '*'
  defaultAllowPrivilegeEscalation: true
  fsGroup:
    rule: RunAsAny
  hostPorts:
  - max: 65535
    min: 0
  privileged: true
  runAsUser:
    rule: RunAsAny
  seLinux:
    rule: RunAsAny
  supplementalGroups:
    rule: RunAsAny
  volumes:
  - configMap
  - secret
  - emptyDir
  - downwardAPI
  - projected
  - persistentVolumeClaim
  - nfs
```

jandubois:
@f0rmiga Could you take a look at this please; I was hoping to investigate it this week, but I won't get around to it in the near future.

zhanggbj:
I'll close this PR since it isn't valid now. Will **keep** issue https://github.com/SUSE/scf/issues/2081 for further discussion. Thanks!

PR #2085, opened 2019-02-11
Added postgres minibroker brain test

f0rmiga:
## Description

Brain test for minibroker postgres.

## Test plan

```
$ (export FISSILE_ROLES=acceptance-tests-brain ; make scf-release && fissile build packages && fissile build images && make/kube)
$ make/tests acceptance-tests-brain env.INCLUDE="postgres"
```


make/tests
```diff
@@ -92,7 +92,7 @@ do
     echo 1>&2 "${POD_NAME} container failed to reach Running state"
     exit 1
   fi
-  sleep 1
+  sleep 5
```
f0rmiga: @mook-as It's not a random change :)

mook-as: Great! Why is it here? :D

src/scf-release/src/acceptance-tests-brain/test-resources/postgres-example-app/README.md
```diff
@@ -0,0 +1,3 @@
+# Postgres example app
+
+This is an example app initially intended to be used by the Postgres minibroker brain tests.
```
mook-as: Do we not have an existing sample app that can talk postgres?

mook-as: Now that I think about it… we have https://github.com/scf-samples/rails-example. Why the heck didn't I use that one for mariadb? I'm dumb. Sorry.

make/tests
f0rmiga: Because 1 second it too fast. I noticed sometimes it failed before the pod was ready.

mook-as: :+1: (I just wanted _some_ sort of reason to show that it's not random :stuck_out_tongue:)

f0rmiga: I didn't run that beautiful sed command in this file this time.

PR #2086, opened 2019-02-12
Brain tests - match the helm client according to the server version

f0rmiga:
## Description

**For brain tests**

Perform a check whether the helm client matches the server version. If not, download the version to match the server.

## Test plan

Get a Tiller version different than 2.11.0 deployed to a supported k8s cluster and run a minibroker brain test.

### DIFF
--- a/src/scf-release/jobs/acceptance-tests-brain/templates/run.erb
+++ b/src/scf-release/jobs/acceptance-tests-brain/templates/run.erb
@@ -9,17 +9,13 @@ set -o errexit
 HELM_BIN_DIR=/var/vcap/packages/helm/bin
 HELM_VERSION="$(${HELM_BIN_DIR}/helm version | grep Server | sed -e 's/.*"v//' -e 's/".*//')"
 
-echo Packaged helm clients ...
-( cd ${HELM_BIN_DIR} ; find . -type f | sed -e 's/^/= /' )
 echo Helm server ${HELM_VERSION}
-echo Compare
 
-if test -f "${HELM_BIN_DIR}/${HELM_VERSION}/helm"
-then
-    echo Match
-else
-    echo "Helm server ${HELM_VERSION} not supported, only: $(cd ${HELM_BIN_DIR} ; ls | sort)"
-    exit 1
+if ! test -f "${HELM_BIN_DIR}/${HELM_VERSION}/helm"; then
+    printf "Helm v%s could not be found. Installing...\\n" "${HELM_VERSION}"
+    helm_url="https://storage.googleapis.com/kubernetes-helm/helm-v${HELM_VERSION}-linux-amd64.tar.gz"
+    mkdir -p "${HELM_BIN_DIR}/${HELM_VERSION}"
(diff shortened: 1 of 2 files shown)


f0rmiga:
[Minibroker: brain test for Redis fails due to client mismatch](https://trello.com/c/ZLlm1pbe/933-minibroker-brain-test-for-redis-fails-due-to-client-mismatch)

src/scf-release/packages/helm/packaging
```diff
@@ -4,26 +4,15 @@ set -e
 
 BIN_DIR="${BOSH_INSTALL_TARGET}/bin"
 
-VERSIONS=""
-VERSIONS="${VERSIONS} 2.11.0"
-VERSIONS="${VERSIONS} 2.8.2"
+VERSION="2.11.0"
```
mook-as: We should still pack a few versions so that the happy path is not installing extra things

andreas-kupries: I would be more happy with that as well.

f0rmiga: Done! I'm also happy with it!

src/scf-release/jobs/acceptance-tests-brain/templates/run.erb
```diff
@@ -9,17 +9,13 @@ set -o errexit
...
 echo Helm server ${HELM_VERSION}
-echo Compare
 
-if test -f "${HELM_BIN_DIR}/${HELM_VERSION}/helm"
-then
-    echo Match
-else
-    echo "Helm server ${HELM_VERSION} not supported, only: $(cd ${HELM_BIN_DIR} ; ls | sort)"
-    exit 1
+if ! test -f "${HELM_BIN_DIR}/${HELM_VERSION}/helm"; then
```
andreas-kupries: `! test` feels wrong compared to `test !` ... Checking bash manpage it looks to be valid.

src/scf-release/jobs/acceptance-tests-brain/templates/run.erb
```diff
@@ -9,17 +9,13 @@ set -o errexit
 HELM_BIN_DIR=/var/vcap/packages/helm/bin
 HELM_VERSION="$(${HELM_BIN_DIR}/helm version | grep Server | sed -e 's/.*"v//' -e 's/".*//')"
 
-echo Packaged helm clients ...
-( cd ${HELM_BIN_DIR} ; find . -type f | sed -e 's/^/= /' )
 echo Helm server ${HELM_VERSION}
-echo Compare
```
andreas-kupries: Dislike the removal of this, keeping would help future debugging. Still, can live with it.

src/scf-release/jobs/acceptance-tests-brain/templates/run.erb
f0rmiga: To avoid confusion, did `[ ! -f ]`.

src/scf-release/jobs/acceptance-tests-brain/templates/run.erb
f0rmiga: I added an informative statement inside the `if`. I'm not favorable of debugging statements, but informative statements that also help debugging I'm all up!

PR #2095, opened 2019-02-14
acceptance-tests-brain: use same app for mariadb & postgres tests

mook-as:
This drops [pong_matcher_go](https://github.com/cloudfoundry-samples/pong_matcher_go) and [postgres-example-app](https://github.com/SUSE/scf/tree/develop/src/scf-release/src/acceptance-tests-brain/test-resources/postgres-example-app) in favour of [rails-example](https://github.com/scf-samples/rails-example) (which supports multiple DB backends already).

This should wait for https://github.com/scf-samples/rails-example/pull/3 to get more extensive testing though.


f0rmiga:
Rebased with the develop branch.

PR #2097, opened 2019-02-14
Bump SCF to cf 5.1

andreas-kupries:
Ref: https://trello.com/c/jKAvgx6W/929-5-post-131-cf-big-bump-part-2-as-most-recent-as-possible
UAA = https://github.com/SUSE/uaa-fissile-release/pull/141


container-host-files/etc/scf/config/opinions-common.yml
```diff
@@ -357,8 +356,7 @@ properties:
     service_name: 'persi-nfs'
     service_id: '20e22be2-a4fb-4b17-a5ef-0c9dba8bdc74'
   router:
-    backends:
-      enable_tls: true
+    ca_certs: ''
```
mook-as: This should be customizable (or at least stick a Trello card in)

container-host-files/etc/scf/config/role-manifest.yml
```diff
@@ -4,21 +4,25 @@ releases:
...
-  version: "1.66.0"
-  url: "https://bosh.io/d/github.com/cloudfoundry/capi-release?v=1.66.0"
-  sha1: "4236495d4a047fb9396996aba3daea9e0cb3f1bf"
+  version: "1.70.0"
+  url: "https://bosh.io/d/github.com/cloudfoundry/capi-release?v=1.70.0"
+  sha1: "bf66504c6479cff3754fb1e114e7ea45767db04c"
+- name: cf-cli
+  version: "1.8.0"
+  url: "https://bosh.io/d/github.com/bosh-packages/cf-cli-release?v=1.8.0"
+  sha1: "15d0718bc82ff9394efd2c9798873a39bf7eba06"
```
mook-as: So we can drop the manual bit in the brains?

container-host-files/etc/scf/config/role-manifest.yml
```diff
@@ -174,10 +183,18 @@ instance_groups:
         - name: sched-health
           protocol: TCP
           internal: 8080
+  - name: bpm
+    release: bpm
+    properties:
+      bosh_containerization:
+        run:
+          capabilities: [ALL]
+        pod-security-policy: privileged
```
mook-as: For new manifests we should say `service-account` instead of `pod-security-policy`. (Because trying to do PSPs here directly is incorrect; PSPs are tied to service accounts)

container-host-files/etc/scf/config/role-manifest.yml
```diff
@@ -459,7 +491,7 @@ instance_groups:
                       values:
                       - diego-api
                   topologyKey: "beta.kubernetes.io/os"
-          service-account: withsysresource
+          service-account: privileged
```
mook-as: If we do `service-account` for BPM instead, we can then not touch this (and have it be merged to be `privileged`, at least in theory)

container-host-files/etc/scf/config/role-manifest.yml
```diff
@@ -906,13 +958,93 @@ instance_groups:
...
+    # defaults
+  - name: log-cache-expvar-forwarder
+    release: log-cache
+    # defaults
+    properties:
+      bosh_containerization:
+        colocated_containers:
+        - loggregator-agent
+        run:
+          service-account: privileged
```
mook-as: Why does this need to be privileged? (There's probably a good reason, but we should document it)

container-host-files/etc/scf/config/role-manifest.yml
```diff
@@ -959,13 +1091,41 @@ instance_groups:
...
...nters: '[{ "name": "egress", "source_id": "log-cache", "addr": "http://localhost:6060/debug/vars", "template": "{{.LogCache.Egress}}" }, { "name": "ingress", "source_id": "log-cache", "addr": "http://localhost:6060/debug/vars", "template": "{{.LogCache.Ingress}}" }, { "name": "expired", "source_id": "log-cache", "addr": "http://localhost:6060/debug/vars", "template": "{{.LogCache.Expired}}" }, { "name": "dropped", "source_id": "log-cache", "addr": "http://localhost:6060/debug/vars", "template": "{{.LogCache.Dropped}}" }, { "name": "promql-timeout", "source_id": "log-cache", "addr": "http://localhost:6060/debug/vars", "template": "{{.LogCache.PromQLTimeout}}" }, { "name": "egress", "source_id": "log-cache-nozzle", "addr": "http://localhost:6061/debug/vars", "template": "{{.Nozzle.Egress}}" }, { "name": "ingress", "source_id": "log-cache-nozzle", "addr": "http://localhost:6061/debug/vars", "template": "{{.Nozzle.Ingress}}" }, { "name": "err", "source_id": "log-cache-nozzle", "addr": "http://localhost:6061/debug/vars", "template": "{{.Nozzle.Err}}" }, { "name": "dropped", "source_id": "log-cache-nozzle", "addr": "http://localhost:6061/debug/vars", "template": "{{.Nozzle.Dropped}}" }]'
```
mook-as: Does anything here require expansion? Can we have this in the opinions?

If not, can we at least format this so it's not a giant line?
```
properties.counters: >
[
 {
   name: ergress
…
```

container-host-files/etc/scf/config/scripts/patches/fix_chown_blobstore_packages.sh
```diff
@@ -9,29 +9,27 @@ fi
...
    mkdir -p $data_tmp_dir
-
--  local dirs="$run_dir $log_dir $store_dir $store_tmp_dir $data_dir $data_tmp_dir $nginx_webdav_dir ${nginx_webdav_dir}/.."
+ 
+   chown vcap:vcap $store_dir
+-  local dirs="$run_dir $log_dir $store_tmp_dir $data_dir $data_tmp_dir $nginx_webdav_dir ${nginx_webdav_dir}/.."
 -  local num_needing_chown=$(find $dirs -not -user vcap -or -not -group vcap | wc -l)
-+  local dirs="$run_dir $log_dir $store_dir $store_tmp_dir $data_dir $data_tmp_dir $packages_dir"
++  local dirs="$run_dir $log_dir $store_tmp_dir $data_dir $data_tmp_dir $packages_dir"
 +  local num_needing_chown=$(find -L $dirs -not -user vcap -or -not -group vcap | wc -l)
```
mook-as: If we're no longer using their chown loop (why?), then we shouldn't patch their chown loop :)

container-host-files/etc/scf/config/scripts/patches/fix_haproxy_fd_requirements.sh
```diff
@@ -7,6 +7,10 @@ set -e
 # https://www.pivotaltracker.com/story/show/128789361
 # See HCF-1065 for the outcome of running performance tests on HCF.
 
+# It further contains a temporary patch needed to make kube-dns work
+# for BPM-managed jobs.  This will go away with the transition to the
+# cf-operator, which will not require BPM anymore.
```
mook-as: That starting comment block is wrong now (it's no longer patching a pre-start script); please reword the whole thing so it makes sense (and better describes why it needs to be in one patch file, rather than two separate patches).

Of course, it _would_ be preferable if this could be in two patch files instead (patching the same target file on disk), so that it's easier to remove one after we do the BPM thing

container-host-files/etc/scf/config/scripts/patches/fix_logcache_certs_forwarder.sh
```diff
@@ -0,0 +1,30 @@
+set -e
```
mook-as: Please add a comment on what this does (also, a shebang line).

container-host-files/etc/scf/config/scripts/patches/fix_logcache_certs_scheduler.sh
```diff
@@ -0,0 +1,30 @@
+set -e
```
mook-as: Comment + shebang please

tooling/lib/releases_diff.rb
```diff
@@ -1,24 +1,44 @@
...
+
+        # Path to the anchors for the current stack
+        if "#{ENV['USE_SLE_BASE']}" == "false"
+            stack = ENV['FISSILE_LIGHT_OPEN42']
+        else
+            stack = ENV['FISSILE_LIGHT_SLE12']
+        end
+
+        system("  cat #{stack}   #{@current_src_path}                     > #{@current_manifest}")
+        system("( cat #{stack} ; git show HEAD:#{current_manifest_path} ) > #{@old_manifest}")
```
mook-as: Using `system` seems a bit icky :(

tooling/lib/releases_diff.rb
```diff
@@ -30,26 +50,29 @@ def self.git_root()
     end
 
     # Gets the directory that contains all the final releases downloaded for SCF
-    def current_final_releases_work_dir()
-        current_manifest_path = File.join(ReleasesDiff.git_root, @manifest_path)
-        File.join(File.expand_path('../', current_manifest_path), '.final_releases')
+    # Note FISSILE_FINAL_RELEASES_DIR in `.envrc`.
+    def final_releases_work_dir()
+        File.join(File.expand_path('../', @current_src_path), '.final_releases')
```
mook-as: Maybe `return ENV['FISSILE_FINAL_RELEASES_DIR'] if ENV['FISSILE_FINAL_RELEASES_DIR']` (or something to that effect)?

container-host-files/etc/scf/config/opinions-common.yml
andreas-kupries: IMHO trello card is good enough for now.

container-host-files/etc/scf/config/role-manifest.yml
andreas-kupries: We will have to explicitly switch over, as smoke-tests did. I would not hold the merge about this, and rather add a trello card so to not forget.

container-host-files/etc/scf/config/role-manifest.yml
andreas-kupries: Ok. Can do that.

container-host-files/etc/scf/config/role-manifest.yml
andreas-kupries: Going to a separate SA likely requires a full round of re-testing.
I see how it would be a good thing, I also shudder at the above re-test.

container-host-files/etc/scf/config/role-manifest.yml
andreas-kupries: Will recheck.

container-host-files/etc/scf/config/role-manifest.yml
andreas-kupries: Opinions apply to all uses the same. We have different definitions for these expvar-forwarder properties in the two places the job occurs at (See roles doppler, log-cache-scheduler).

I can try to re-format, and hope that I did not break yaml.

container-host-files/etc/scf/config/scripts/patches/fix_chown_blobstore_packages.sh
andreas-kupries: I have to recheck. IIRC the patch as it was did not apply anymore and I simply made it apply without conflict again.

container-host-files/etc/scf/config/scripts/patches/fix_haproxy_fd_requirements.sh
andreas-kupries: Will check

container-host-files/etc/scf/config/scripts/patches/fix_logcache_certs_forwarder.sh
andreas-kupries: Will do

container-host-files/etc/scf/config/scripts/patches/fix_logcache_certs_scheduler.sh
andreas-kupries: Sure

tooling/lib/releases_diff.rb
andreas-kupries: IIRC was the fastest way to hammer it into a working thing without spending time on how to exec subprocesses from ruby, and/or assemble files from pieces.

Better code appreciated

tooling/lib/releases_diff.rb
andreas-kupries: If it will work, sure.

container-host-files/etc/scf/config/role-manifest.yml
mook-as: I'm okay with not doing that for now

tooling/lib/releases_diff.rb
mook-as: I'll try to do that as part of the follow-up bump.

container-host-files/etc/scf/config/opinions-common.yml
andreas-kupries: Ref: https://trello.com/c/2ZUa6mv4/625-make-routercacerts-customizable

container-host-files/etc/scf/config/role-manifest.yml
andreas-kupries: Ref: https://trello.com/c/reOzKNER/626-switch-brain-tests-to-cf-cli-release

container-host-files/etc/scf/config/role-manifest.yml
andreas-kupries: Done for all uses, with commit c5c618134e4d775c67f8d2b8b23fe9f450ca2873.
Separate commit, to make it easy to undo in case of trouble.

container-host-files/etc/scf/config/role-manifest.yml
andreas-kupries: Ref https://trello.com/c/gEuXlu5g/627-give-bpm-job-its-own-service-account

container-host-files/etc/scf/config/role-manifest.yml
andreas-kupries: IIRC without it the fact that the loggregator-agent became privileged was not properly accounted for in the output of fissile. And I do not remember in full why the privileged on the BPM in this role was not enough.

You might wish to try to undo this and see if my recollection is true.

container-host-files/etc/scf/config/role-manifest.yml
andreas-kupries: Done with commit d8a742961bdd0b5cd0e450f5e56117830c5bec02

container-host-files/etc/scf/config/scripts/patches/fix_chown_blobstore_packages.sh
andreas-kupries: I am not sure why you believe that `we're no longer using their chown loop` ... The only thing I see changed in the patch is that `$store_dir` was removed from the list of directories to handle.

container-host-files/etc/scf/config/scripts/patches/fix_haproxy_fd_requirements.sh
andreas-kupries: Done with commit d18bf53f7e914783360b615d9d38c2ffc3739cea

container-host-files/etc/scf/config/scripts/patches/fix_logcache_certs_forwarder.sh
andreas-kupries: Done with commit 9e821dab0bd8d8bb11609c134d32d97119ac85b7

container-host-files/etc/scf/config/scripts/patches/fix_logcache_certs_scheduler.sh
andreas-kupries: Done with commit 9e821dab0bd8d8bb11609c134d32d97119ac85b7

tooling/lib/releases_diff.rb
andreas-kupries: Done with commit c92adf2c720739e0e45b885d7ace25e7b88f75ad

container-host-files/etc/scf/config/scripts/patches/fix_chown_blobstore_packages.sh
mook-as: Ah, sorry, I got confused reading a patch to a patch.  We _didn't_ do it, upstream did (you're just fixing the context here).  Ignore this then.

container-host-files/etc/scf/config/role-manifest.yml
```diff
@@ -906,13 +958,102 @@ instance_groups:
...
+      properties.maps: >
+        [{ "name": "worker-assignments",
+	   "source_id": "log-cache-scheduler",
+	   "addr": "http://localhost:6064/debug/vars",
+	   "template": "{{.WorkerAssignments | jsonMap}}" },
+	 { "name": "worker-health",
+	   "source_id": "log-cache-scheduler",
+	   "addr": "http://localhost:6064/debug/vars",
+	   "template": "{{.WorkerHealth | jsonMap}}" }
+	]
```
mook-as: This will probably break (because YAML hates tabs).

container-host-files/etc/scf/config/role-manifest.yml
```diff
@@ -959,13 +1100,115 @@ instance_groups:
...
+      properties.cc.common_name: 'SCF CA'
+      properties.cc.cert: '"((LOG_CACHE_CERT))"'
+      properties.cc.key:  '"((LOG_CACHE_CERT_KEY))"'
+      properties.uaa.client_id: 'doppler'
+      properties.uaa.client_secret: '"((UAA_CLIENTS_DOPPLER_SECRET))"'
+      properties.uaa.ca_cert: '"((UAA_CA_CERT))((^UAA_CA_CERT))((INTERNAL_CA_CERT))((/UAA_CA_CERT))"'
+      properties.uaa.internal_addr: '"https://((KUBERNETES_NAMESPACE)).((UAA_HOST)):((UAA_PORT))"'
+      properties.counters: >
+        [{ "name": "egress",
+	   "source_id": "log-cache",
```
mook-as: Tabs here too.

container-host-files/etc/scf/config/role-manifest.yml
andreas-kupries: Oops. Sure, will fix.

andreas-kupries: Done with commit 35498d43eecd091908c6e7c4852e10b0509937bb

PR #2098, opened 2019-02-15
Ingress resource - add necessary annotations and small refactors

f0rmiga:
## Description

- Adds missing NGINX Ingress Controller annotations.
- Changes the name of the resources to more meaningful ones (not tied to a specific Ingress Controller).
- Injects the backend port.

## Test plan

- Deploy the NGINX Ingress Controller.
- Deploy SCF with the below additional values and no `kube.external_ips`:
```
services:
  ingress:
    class: nginx
    backends:
      router:
        port: 443
```
- Check GoRouter can be reached through the Ingress Controller.

### DIFF
--- a/chart-parts/ingress.yaml
+++ b/chart-parts/ingress.yaml
@@ -1,4 +1,4 @@
-{{- if .Values.services.ingress }}
+{{- if .Values.services.ingress -}}
 ---
 # The cert and the key cannot be filled automatically. If we could it
 # would roughly look like the lines below. As is it is the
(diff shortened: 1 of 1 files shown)


f0rmiga:
[(3) Fissile work to use an Ingress Controller](https://trello.com/c/jgev7lC4/892-3-fissile-work-to-use-an-ingress-controller)

chart-parts/ingress.yaml
```diff
@@ -31,14 +32,19 @@ data:
 apiVersion: "extensions/v1beta1"
 kind: "Ingress"
 metadata:
-  name: "ingress-scf"
-  namespace: {{ .Release.Namespace | quote }}
+  name: "{{ .Release.Name }}-{{ .Values.services.ingress.class }}"
+  namespace: "{{ .Release.Namespace }}"
```
andreas-kupries: Same question as for https://github.com/SUSE/uaa-fissile-release/pull/142/files, prefer to stay with `| quote` as the dominant way for us quoting a single value.

f0rmiga: Answered on the referenced PR.

PR #2110, opened 2019-02-19
Ingress resource - refactor make scripts and add ingress bits

f0rmiga:
## Description

Adds the ingress scripts and refactors the current scripts to use the Ingress Controller.

## Test plan

- `export INGRESS_CONTROLLER=nginx` on a Vagrant ssh session.
- `make run` and/or `make upgrade` to make use of the Ingress Controller.
- Check that only the Ingress Controller got an external ip `kubectl get services --all-namespaces`.


f0rmiga:
[(3) Fissile work to use an Ingress Controller](https://trello.com/c/jgev7lC4/892-3-fissile-work-to-use-an-ingress-controller)

.envrc
```diff
@@ -103,3 +103,5 @@ fi
 export FISSILE_AUTH_TYPE=rbac
 export FISSILE_OUTPUT_DIR="${FISSILE_OUTPUT_DIR:-${PWD}/output/helm}"
 export FISSILE_DEFAULTS_FILE=
+
+export INGRESS_CONTROLLER=${INGRESS_CONTROLLER:-}
```
mook-as: So this just sets it to empty if unset?
Seems unnecessary to put it in `.envrc` (since it has the side effect of resetting your value if you:

1. `cd ~/scf`
2. `set INGRESS_CONTROLLER=nginx` # forgot to do this
3. `do stuff …`
4. `cd ..` # for whatever reason
5. `cd scf` # oops is now unset

I believe step 4 could even be `cd src/uaa-fissile-release` because it reloads the `.envrc`.

make/ingress/run
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env bash
+
+: "${GIT_ROOT:=$(git rev-parse --show-toplevel)}"
+
+if [ -z "${INGRESS_CONTROLLER}" ]; then
```
mook-as: Wouldn't this look better as:

```bash
case "${INGRESS_CONTROLLER}" in
  "")
    echo "No Ingress Controller specified; set INGRESS_CONTROLLER"
    ;;
  nginx)
    "${GIT_ROOT}/make/ingress/nginx/run"
    ;;
  *)
    echo "Ingress controller '${INGRESS_CONTROLLER}' not recognized" >&2
    exit 1
esac
```

Alternatively:
```bash
if [ -z "${INGRESS_CONTROLLER}" ]; then
  echo "No Ingress Controller specified; set INGRESS_CONTROLLER"
  exit
fi

if [ -x "${GIT_ROOT}/make/${INGRESS_CONTROLLER}/run" ] ; then
  exec "${GIT_ROOT}/make/${INGRESS_CONTROLLER}/run" "$@"
fi

echo "Ingress controller '${INGRESS_CONTROLLER}' not recognized" >&2
exit 1
```

(Actually, do we even want the empty string message, if `make run` will run this by default with no controllers?)

make/ingress/stop
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env bash
+
+: "${GIT_ROOT:=$(git rev-parse --show-toplevel)}"
+
+if [ -z "${INGRESS_CONTROLLER}" ]; then
```
mook-as: See above.

make/upgrade
```diff
@@ -34,18 +34,22 @@ helm_args=(
...
-    --set "kube.external_ips[1]=$(getent hosts "${DOMAIN}" | awk 'NR=1{print $1}')"
-)
+if [ -n "${INGRESS_CONTROLLER}" ]; then
+    helm_args+=(
+        --set "services.ingress.class=${INGRESS_CONTROLLER}"
+        --set "services.ingress.backends.router.port=443"
+    )
+else
+    helm_args+=(
+        --set "kube.external_ips[0]=$(getent hosts "${DOMAIN}" | awk 'NR=1{print $1}')"
```
mook-as: You're dropping the check on multiple IPs (see the comment that is being deleted).

make/ingress/nginx/run
```diff
@@ -0,0 +1,32 @@
...
+  --set "tcp.20002=${CF_NAMESPACE}/tcp-router-tcp-router-public:20002"
+  --set "tcp.20003=${CF_NAMESPACE}/tcp-router-tcp-router-public:20003"
+  --set "tcp.20004=${CF_NAMESPACE}/tcp-router-tcp-router-public:20004"
+  --set "tcp.20005=${CF_NAMESPACE}/tcp-router-tcp-router-public:20005"
+  --set "tcp.20006=${CF_NAMESPACE}/tcp-router-tcp-router-public:20006"
+  --set "tcp.20007=${CF_NAMESPACE}/tcp-router-tcp-router-public:20007"
+  --set "tcp.20008=${CF_NAMESPACE}/tcp-router-tcp-router-public:20008"
+  --set "tcp.4443=${CF_NAMESPACE}/router-gorouter-public:4443"
+  --set "tcp.2222=${CF_NAMESPACE}/diego-ssh-ssh-proxy-public:2222"
+  --set "tcp.${UAA_PORT}=${UAA_NAMESPACE}/uaa-uaa-public:${UAA_PORT}"
```
mook-as: TODO: don't use 2793 for UAA? If we're doing nginx there's no reason to have a separate port for it. (I'm totally okay with making that a separate PR)

.envrc
f0rmiga: True. I initially thought it wouldn't hurt, but I guess you convinced me to take it out of here.

make/upgrade
f0rmiga: I saw the comment. Isn't this testing someone else's problem? We know kube's `externalIPs` work with multiple IP addresses.

make/ingress/nginx/run
f0rmiga: Initially I set port 443 for UAA, so it went through the Ingress Controller, but then some CATS failed due to some CLI using 2793. I guess it should be part of another card.

make/upgrade
mook-as: For the services, yes; but this is testing that we propagate it down correctly (because we had issues before where it didn't). It used to be called `kube.external_ip` or something, and only supported one.

make/ingress/nginx/run
mook-as: Yeah, I think it's another card to make UAA use 443 (so that we actually test that it works)

make/upgrade
f0rmiga: As discussed in person, Jenkins would catch any abnormal behavior. Leaving the code as is in the PR.

make/ingress/nginx/run
f0rmiga: Card created.

make/ingress/nginx/run
```diff
@@ -0,0 +1,32 @@
...
+  # https://kubernetes.github.io/ingress-nginx/user-guide/exposing-tcp-udp-services
+  --set "tcp.20000=${CF_NAMESPACE}/tcp-router-tcp-router-public:20000"
+  --set "tcp.20001=${CF_NAMESPACE}/tcp-router-tcp-router-public:20001"
+  --set "tcp.20002=${CF_NAMESPACE}/tcp-router-tcp-router-public:20002"
+  --set "tcp.20003=${CF_NAMESPACE}/tcp-router-tcp-router-public:20003"
+  --set "tcp.20004=${CF_NAMESPACE}/tcp-router-tcp-router-public:20004"
+  --set "tcp.20005=${CF_NAMESPACE}/tcp-router-tcp-router-public:20005"
+  --set "tcp.20006=${CF_NAMESPACE}/tcp-router-tcp-router-public:20006"
+  --set "tcp.20007=${CF_NAMESPACE}/tcp-router-tcp-router-public:20007"
+  --set "tcp.20008=${CF_NAMESPACE}/tcp-router-tcp-router-public:20008"
```
mook-as: Just realized… we don't have 20009 in the list?

make/ingress/run
```diff
@@ -0,0 +1,14 @@
+#!/usr/bin/env bash
+
+if [ -z "${INGRESS_CONTROLLER}" ]; then
```
mook-as: Is this always set? Do we need to use `if [ -z "${INGRESS_CONTROLLER:-}" ]` instead? (I guess we don't have `-o nounset` here… which then brings up the question, why don't we?)

make/ingress/update_secrets
```diff
@@ -0,0 +1,63 @@
...
+if [ -z "${INGRESS_CONTROLLER:-}" ]; then
+  exit
+fi
+
+: "${GIT_ROOT:=$(git rev-parse --show-toplevel)}"
+source "${GIT_ROOT}/make/include/defaults"
+source "${GIT_ROOT}/make/include/secrets"
+
+printf "Waiting for UAA secret generation job to finish..."
+until [[ $(kubectl get pods --namespace "${UAA_NAMESPACE}" --selector "app.kubernetes.io/component=secret-generation" -o jsonpath='{.items[?(@.status.phase=="Succeeded")].status.containerStatuses[?(@.state.terminated.reason=="Completed")].state.terminated.exitCode}') == 0 ]]; do
```
mook-as: This magic selector should be in a helper function somewhere I think.

make/ingress/nginx/run
mook-as: Clarification: The patch is correct as-is; it was just surprising that we do _not_ listen on port 20009 at all.

f0rmiga: 20009 is not a port listed on any public service we currently deploy.

make/ingress/run
f0rmiga: The script was simple enough for not using `-o nounset`, but I guess it is just better having it then? Pushing on the next commit.

make/include/wait_for
```diff
@@ -0,0 +1,39 @@
...
+        exit 1
+        ;;
+      :)
+        echo "Option -$OPTARG requires an argument." >&2
+        exit 1
+        ;;
+    esac
+  done
+
+  until [[ -n "$(kubectl get jobs --namespace "${namespace}" --selector "${selector}" -o jsonpath='{.items[?(@.status.succeeded==1)].status.conditions[?(@.type=="Complete")]}')" ]]; do
```
mook-as: Do we need to check that `-n`, `-l`, `-t` have been specified? (Though `-t` should probably have a default?)
Maybe build the command line as an array variable first, so that it's not one giant line?

make/ingress/run
```diff
@@ -1,6 +1,8 @@
 #!/usr/bin/env bash
 
-if [ -z "${INGRESS_CONTROLLER}" ]; then
+set -o nounset
```
mook-as: If you're going back to edit this later, maybe `-o errexit`? Not a big deal though.

make/include/wait_for
f0rmiga: The check is performed in the case of getopts already.

mook-as: As far as I can tell, nothing in [`getopts`](https://www.gnu.org/software/bash/manual/html_node/Bourne-Shell-Builtins.html#index-getopts) enforces that all arguments are provided?

make/include/wait_for
```diff
@@ -33,7 +33,27 @@ wait_for_job() {
     esac
   done
 
-  until [[ -n "$(kubectl get jobs --namespace "${namespace}" --selector "${selector}" -o jsonpath='{.items[?(@.status.succeeded==1)].status.conditions[?(@.type=="Complete")]}')" ]]; do
+  if [[ -z "${namespace:-}" ]]; then
+    echo "Missing namespace. '-n' must be specified." >&2
+    exit 1
```
mook-as: `return 1`, maybe (to bubble things up to the caller)? Not that it makes a difference…

make/include/wait_for
```diff
@@ -33,7 +33,27 @@ wait_for_job() {
...
+  fi
+
+  if [[ -z "${selector:-}" ]]; then
+    echo "Missing selector. '-l' must be specified." >&2
+    exit 1
+  fi
+
+  if [[ -z "${time_to_sleep:-}" ]]; then
+    time_to_sleep=3
```
mook-as: Just do `local time_to_sleep=3` on line 12?

make/include/wait_for
```diff
@@ -33,7 +33,27 @@ wait_for_job() {
...
+  fi
+
+  check_args=(
+    --namespace "${namespace}"
+    --selector "${selector}"
+    --output "jsonpath='{.items[?(@.status.succeeded==1)].status.conditions[?(@.type==\"Complete\")]}'"
```
mook-as: Huh, do you need the single quotes on the value? Doesn't `--output jsonpath={.items…` work?

make/include/wait_for
f0rmiga: -o errexit wouldn't do an exit 1?

mook-as: No, because this is a function, not a subshell; `return 1` means the caller can catch things, whereas `exit 1` means the shell exits.
(But we probably don't care)

f0rmiga: I did `return` anyways.

Issue #2100, opened 2019-02-15
Not able to fork processes as vcap user anymore


eyenx:
See moby Issue: https://github.com/moby/moby/issues/27444

SUSE SCF images seem to have a nproc limit set  for vcap user (inside the image itself):

```
# tail -5 /etc/security/limits.conf
vcap  soft  nproc  4096
vcap  hard  nproc  4096
*               hard    core            0
```

Because SUSE SCF Pods run processes as vcap user with UID 1000, sometimes it can happen, that too many processes are spawned by UID 1000 on one single Kubernetes node. This brings us  to a race condition where processes inside Pods cannot be started with user vcap anymore and stay in RUNNING state with an unhealthy status "0/1".

Can the limit be raised or can it be customized with HELM Values?

jandubois:
I was looking into user namespace remapping to see if that can help us, but it is not currently supported, and even the planned/proposed feature does not support different mappings per pod:

> Non-goal is to support pod/container level user namespace isolation. There can be images using different users but on the node, pods/containers running with these images will share common user namespace remapping configuration. In other words, all containers on a node share a common user-namespace range.

https://github.com/kubernetes/community/blob/master/contributors/design-proposals/node/node-usernamespace-remapping.md#non-goals

jandubois:
> See moby Issue: [moby/moby#27444](https://github.com/moby/moby/issues/27444)

Oh, should have followed that link first; even if it was supported the remapping wouldn't help.

eyenx:
Any updates on this? We would really welcome a possibility to change that limits.conf variable with helm values if possible.

f0rmiga:
@eyenx I have an open PR for this issue. I just wanted to confirm if there is any other role other than diego-cell that needs to have this configurable.

f0rmiga:
@eyenx You can use the charts generated by our CI to try it out:
uaa-opensuse -> https://s3.amazonaws.com/cap-release-archives/prs/PR-2155-uaa-opensuse-helm-2.15.2%2Bcf5.5.0.183.gaf596484.tgz
cf-opensuse -> https://s3.amazonaws.com/cap-release-archives/prs/PR-2155-cf-opensuse-helm-2.15.2%2Bcf5.5.0.183.gaf596484.tgz

eyenx:
Hi @f0rmiga

Thanks, will try them out and give feedback! 

Thanks, regards

f0rmiga:
[...]

PR #2155, opened 2019-03-04
Set vcap hard and soft nproc limits for every role

f0rmiga:
## Description

This PR resolves https://github.com/SUSE/scf/issues/2100.
It enables changing the hard and soft limits for the max processes for the vcap user inside the container for each role.

They can be set by passing the `env.VCAP_HARD_NPROC` and `env.VCAP_SOFT_NPROC`  values during deployment.

## Test plan

- Deploy SCF without specifying `VCAP_HARD_NPROC` and `VCAP_SOFT_NPROC`, and:
  - Get a shell session to any container and:
    - `su vcap`
    - `ulimit -u` -> should return 4096 (the default value).
- Perform an upgrade setting only `VCAP_HARD_NPROC`:
  - The containers should error. Grab the logs of the failed container with `kubectl logs -n cf diego-cell-0 -c diego-cell -p`. The error should say `VCAP_SOFT_NPROC must be set when VCAP_HARD_NPROC is set`.
- Repeat the previous point, but setting only `VCAP_SOFT_NPROC`. A similar error should be raised.
- Perform an upgrade setting both `VCAP_HARD_NPROC` and `VCAP_SOFT_NPROC`, but set `VCAP_SOFT_NPROC` to a value higher than `VCAP_HARD_NPROC`. It should fail with `VCAP_SOFT_NPROC (<val>) cannot be larger than VCAP_HARD_NPROC (<val>)`.
- Perform an upgrade setting both `VCAP_HARD_NPROC` and `VCAP_SOFT_NPROC` to 8192, and:
  - Get a shell session to any container and:
    - `su vcap`
    - `ulimit -u` -> should return 8192 (the provided value).
    - `exit`.
    - `cat /etc/security/limits.conf`. Check that both the hard and soft limits were set accordingly.

### DIFF
--- a/container-host-files/etc/scf/config/role-manifest.yml
+++ b/container-host-files/etc/scf/config/role-manifest.yml
@@ -151,6 +151,7 @@ releases:
 instance_groups:
 - name: syslog-scheduler
   scripts:
+  - scripts/set_vcap_limits.sh
   - scripts/forward_logfiles.sh
   - scripts/patches/fix_monit_rsyslog.sh
   post_config_scripts:
@@ -201,6 +202,7 @@ instance_groups:
         pod-security-policy: privileged
 - name: adapter
   scripts:
+  - scripts/set_vcap_limits.sh
   - scripts/forward_logfiles.sh
   - scripts/patches/fix_monit_rsyslog.sh
   post_config_scripts:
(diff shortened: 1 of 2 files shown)


f0rmiga:
[Allow bumping up the nproc limit for vcap user](https://trello.com/c/XvU6E5Vg/955-allow-bumping-up-the-nproc-limit-for-vcap-user)

f0rmiga:
[(3) brain: 009_backup: confirm backup and restore finish successfully](https://trello.com/c/fdFfW2nr/952-3-brain-009backup-confirm-backup-and-restore-finish-successfully)

PR #2156, opened 2019-03-04
Brain test - backup and restore without sleep

f0rmiga:
## Description

There is no need for `sleep 60` after running `cf backup restore` as it is not an async operation.

(Also fixing punctuation)

## Test plan

Brain tests should pass.


f0rmiga:
[(3) brain: 009_backup: confirm backup and restore finish successfully](https://trello.com/c/fdFfW2nr/952-3-brain-009backup-confirm-backup-and-restore-finish-successfully)

PR #2170, opened 2019-03-06
Jenkins: Grab and dump logs to s3 on failed runs

bikramnehra:
## Description

- Modified `Jenkinsfile` to execute the `klog.sh` script
- Modified `s3` bucket path to accommodate `klogs`
- Modified `klog.sh` to support args from Jenkins

## Test plan

- Introduce failures in tests and trigger build
- Check whether `klog.sh` was triggered even if the build failed
- Verify if both build logs and klogs were dumped in correct s3 buckets

e.g: `https://s3.console.aws.amazon.com/s3/buckets/cap-jenkins-log/prs/PR-2170-jenkins-scf-PR-2170-6/`

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -870,9 +870,12 @@ pass = ${OBS_CREDENTIALS_PASSWORD}
             }
             // Save logs of failed builds to s3 - we want to analyze where we may have jenkins issues.
             script {
-                if ((env.BRANCH_NAME == 'develop' || env.BRANCH_NAME == 'master') && fileExists('bin/clean-jenkins-log')) {
+                if (env.BRANCH_NAME == 'develop' || env.BRANCH_NAME == 'master') {
                     writeFile(file: 'build.log', text: getBuildLog())
-                    sh 'bin/clean-jenkins-log'
+                    sh """ 
+                         bin/clean-jenkins-log
+                         container-host-files/opt/scf/bin/klog.sh -f ${jobBaseName()}-${BUILD_NUMBER}-scf
+                       """
                     withAWS(region: params.S3_REGION) {
                         withCredentials([usernamePassword(
                             credentialsId: params.S3_CREDENTIALS,
@@ -880,12 +883,17 @@ pass = ${OBS_CREDENTIALS_PASSWORD}
                             passwordVariable: 'AWS_SECRET_ACCESS_KEY',
                         )]) {
                             script {
-                                def subdir = "${params.S3_PREFIX}${distSubDir()}"
                                 def prefix = distPrefix()
+                                def subdir = "${params.S3_PREFIX}${distSubDir()}${prefix}${env.BUILD_TAG}/"
                                 s3Upload(
                                     file: 'cleaned-build.log',
                                     bucket: "${params.S3_LOG_BUCKET}",
-                                    path: "${subdir}${prefix}${env.BUILD_TAG}",
+                                    path: "${subdir}",
+                                )
+                                s3Upload(
+                                    file: 'klog.tar.gz',
+                                    bucket: "${params.S3_LOG_BUCKET}",
+                                    path: "${subdir}",
                                 )
                             }
                         }
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -872,10 +872,8 @@ pass = ${OBS_CREDENTIALS_PASSWORD}
             script {
                 if (env.BRANCH_NAME == 'develop' || env.BRANCH_NAME == 'master') {
                     writeFile(file: 'build.log', text: getBuildLog())
-                    sh """ 
-                         bin/clean-jenkins-log
-                         container-host-files/opt/scf/bin/klog.sh -f ${jobBaseName()}-${BUILD_NUMBER}-scf
-                       """
+                    sh 'bin/clean-jenkins-log'
+                    sh 'container-host-files/opt/scf/bin/klog.sh -f ${jobBaseName()}-${BUILD_NUMBER}-scf'
                     withAWS(region: params.S3_REGION) {
                         withCredentials([usernamePassword(
                             credentialsId: params.S3_CREDENTIALS,
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -882,7 +882,8 @@ pass = ${OBS_CREDENTIALS_PASSWORD}
                         )]) {
                             script {
                                 def prefix = distPrefix()
-                                def subdir = "${params.S3_PREFIX}${distSubDir()}${prefix}${env.BUILD_TAG}/"
+                                prefix = prefix.substring(0, prefix.length() - 1)
+                                def subdir = "${params.S3_PREFIX}${distSubDir()}${prefix}/${env.BUILD_TAG}/"
                                 s3Upload(
                                     file: 'cleaned-build.log',
                                     bucket: "${params.S3_LOG_BUCKET}",
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -882,6 +882,7 @@ pass = ${OBS_CREDENTIALS_PASSWORD}
                         )]) {
                             script {
                                 def prefix = distPrefix()
+                                // Trimming trailing "-" from path as distPrefix() returns "PR-${CHANGE_ID}-".
                                 prefix = prefix.substring(0, prefix.length() - 1)
                                 def subdir = "${params.S3_PREFIX}${distSubDir()}${prefix}/${env.BUILD_TAG}/"
                                 s3Upload(
--- a/container-host-files/opt/scf/bin/klog.sh
+++ b/container-host-files/opt/scf/bin/klog.sh
@@ -21,7 +21,13 @@ if [ "$1" == "-f" ]; then
   FORCE=1
 fi
 
-NS=${1-scf}
+# read namespace val from args if provided
+if [ ! -z "$2" ]; then
+  NS=$2
+else
+  NS=${1-scf}
+fi 
+
 DONE="${KLOG}/${NS}/done"
 
 if [ "${FORCE}" == "1" ] ; then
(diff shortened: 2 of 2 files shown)


Jenkinsfile
```diff
@@ -870,22 +870,30 @@ pass = ${OBS_CREDENTIALS_PASSWORD}
             }
             // Save logs of failed builds to s3 - we want to analyze where we may have jenkins issues.
             script {
-                if ((env.BRANCH_NAME == 'develop' || env.BRANCH_NAME == 'master') && fileExists('bin/clean-jenkins-log')) {
+                if (env.BRANCH_NAME == 'develop' || env.BRANCH_NAME == 'master') {
                     writeFile(file: 'build.log', text: getBuildLog())
-                    sh 'bin/clean-jenkins-log'
+                    sh """
```
f0rmiga: I think these should be 2 sh calls:
```
sh 'bin/clean-jenkins-log'
sh 'container-host-files/opt/scf/bin/klog.sh .......'
```

Jenkinsfile
```diff
@@ -870,22 +870,30 @@ pass = ${OBS_CREDENTIALS_PASSWORD}
...
                     withAWS(region: params.S3_REGION) {
                         withCredentials([usernamePassword(
                             credentialsId: params.S3_CREDENTIALS,
                             usernameVariable: 'AWS_ACCESS_KEY_ID',
                             passwordVariable: 'AWS_SECRET_ACCESS_KEY',
                         )]) {
                             script {
-                                def subdir = "${params.S3_PREFIX}${distSubDir()}"
                                 def prefix = distPrefix()
+                                def subdir = "${params.S3_PREFIX}${distSubDir()}${prefix}${env.BUILD_TAG}/"
```
f0rmiga: I think this is the one generating weird directory names like `PR-2170-jenkins-scf-PR-2170-6`. No need for `PR-2170` showing up twice.

Jenkinsfile
```diff
@@ -894,4 +902,4 @@ pass = ${OBS_CREDENTIALS_PASSWORD}
             }
         }
     }
-}
+}
```
f0rmiga: Why removing the line break?

container-host-files/opt/scf/bin/klog.sh
```diff
@@ -21,7 +21,13 @@ if [ "$1" == "-f" ]; then
   FORCE=1
 fi
 
-NS=${1-scf}
+# read namespace val from args if provided
```
f0rmiga: Not sure I follow this. `NS` could come either from `$1` or `$2`?

Jenkinsfile
andreas-kupries: Why ?

container-host-files/opt/scf/bin/klog.sh
andreas-kupries: Apparently yes. In Jenkins he is calling it with `klog -f NS` ...
It might be easier to handle things with `shift` to squash the processed argument(s) before going to the next.

f0rmiga: Missing capitalization and punctuation.

andreas-kupries: I had a log at the entire section now, and now I am confused too.
The `-f` handling does a shift. I have now no idea what this construction is for either.

f0rmiga: Yes. https://github.com/SUSE/scf/blob/fec8cd4284c9cbc43bb0775519b2dd9599e6aded/container-host-files/opt/scf/bin/klog.sh#L20

Jenkinsfile
f0rmiga: Clearer visualization on Jenkins.

andreas-kupries: Better separation in the log output then ?

f0rmiga: IMO yes.

container-host-files/opt/scf/bin/klog.sh
bikramnehra: Agreed. The code seems redundant. 

`NS=${1-scf}` this bit enough

Jenkinsfile
andreas-kupries: Ok. Then agreed, two calls it should be

bikramnehra: still not convinced how two subsequent calls will provide better logging. To my understanding, since both scripts intend log collection therefore seems logical to group them together.

container-host-files/opt/scf/bin/klog.sh
bikramnehra: now since this comment will be gone, therefore no need.

guess this comment above serves the purpose `INSTANCE_ID defaults to scf`

Jenkinsfile
f0rmiga: https://jenkins-new.howdoi.website/blue/rest/organizations/jenkins/pipelines/scf/branches/PR-2170/runs/6/nodes/81/steps/125/log/?start=0

They are grouped.

Jenkinsfile
bikramnehra: This one even seemed bit odd to me, however I kept the naming convention consistent with what was already there.

One option is to get rid of prefix which is essentially:`PR-${CHANGE_ID}-` and keep `env.BUILD_TAG`. Though this convention will diverge from rest of the script where prefix is being used

container-host-files/opt/scf/bin/klog.sh
f0rmiga: `# read namespace val from args if provided` -> `# Read namespace value from command-line arguments if provided.`

But you are right. If you'll revert this change, no need for the comment.

Jenkinsfile
bikramnehra: After looking at the logs, splitting it into two separate calls.

Jenkinsfile
bikramnehra: this is about newline at the end right?

Jenkinsfile
andreas-kupries: I am ok with this.

bikramnehra: @andreas-kupries just to be clear, are you okay with the existing convention or my proposal?

As I haven't implemented any changes from the proposal yet for this.

container-host-files/opt/scf/bin/klog.sh
```diff
@@ -21,7 +21,8 @@ if [ "$1" == "-f" ]; then
   FORCE=1
 fi
 
-NS=${1-scf}
+NS=${1-scf}
```
f0rmiga: Could you revert this change? (there is an extra space at the end and a line break.

Jenkinsfile
f0rmiga: At least add a slash after `${prefix}` to make it a folder. We don't want to mess with the BUILD_TAG right now.

container-host-files/opt/scf/bin/klog.sh
bikramnehra: Done!

Jenkinsfile
andreas-kupries: I am ok with the existing convention

Jenkinsfile
```diff
@@ -870,22 +870,28 @@ pass = ${OBS_CREDENTIALS_PASSWORD}
...
                     withAWS(region: params.S3_REGION) {
                         withCredentials([usernamePassword(
                             credentialsId: params.S3_CREDENTIALS,
                             usernameVariable: 'AWS_ACCESS_KEY_ID',
                             passwordVariable: 'AWS_SECRET_ACCESS_KEY',
                         )]) {
                             script {
-                                def subdir = "${params.S3_PREFIX}${distSubDir()}"
                                 def prefix = distPrefix()
+                                def subdir = "${params.S3_PREFIX}${distSubDir()}${prefix}${env.BUILD_TAG}/"
```
andreas-kupries: Ok. @f0rmiga has convinced me that the current definition and path are not suitable.

The proposal is now this:

- Change "${params.S3_PREFIX}${distSubDir()}${prefix}${env.BUILD_TAG}/"
  to "${params.S3_PREFIX}${distSubDir()}${prefix}/${env.BUILD_TAG}/"
  This inserts `/` after the prefix.

- This has the problem that the path now looks like `PR-2170-/jenkins-scf-PR-2170-6`. Note the weird `-/` combination.

- To fix that remove the `-` from the output of the `distPrefix()` function. Should be around line 102 of Jenkinsfile.

- Now as `distPrefix()` is already used in other places, we have to compensate for that removal there, by adding the `-` back in where the output of `distPrefix()` is used. This is in the lines 816 (*), 821, 830. 

(*) In the line 816 it is the variable `prefix` holding the result from `distPrefix()` in line 809.

@bikramnehra ^

bikramnehra: @andreas-kupries Agreed, however, since all other cases require `-`. Therefore, instead of making changes in the `distPrefix()`(also so many other places), we can simply omit `-` while appending to the path:

`prefix = prefix.substring(0, prefix.length() - 1)`

just before line 885

andreas-kupries: That is definitely less work to do. Go ahead. I see no future cases at the moment.

bikramnehra: DONE! Triggered a build to verify if everything's okay.

Jenkinsfile
```diff
@@ -870,22 +870,29 @@ pass = ${OBS_CREDENTIALS_PASSWORD}
...
                     withAWS(region: params.S3_REGION) {
                         withCredentials([usernamePassword(
                             credentialsId: params.S3_CREDENTIALS,
                             usernameVariable: 'AWS_ACCESS_KEY_ID',
                             passwordVariable: 'AWS_SECRET_ACCESS_KEY',
                         )]) {
                             script {
-                                def subdir = "${params.S3_PREFIX}${distSubDir()}"
                                 def prefix = distPrefix()
+                                prefix = prefix.substring(0, prefix.length() - 1)
```
f0rmiga: Last nit: put a comment stating why we are cutting the last character here, as it is not clear for anyone that didn't work on this PR (we ourselves may forget soon).

bikramnehra: comments' in

PR #2176, opened 2019-03-12
Add sle15 stack

thardeck:


container-host-files/etc/scf/config/opinions-sle12.yml
```diff
@@ -14,8 +15,11 @@ the-stacks: &the-stacks
     description: "Cloud Foundry Linux-based filesystem"
   - name: "sle12"
     description: "SLE-based filesystem"
+  - name: "sle15"
+    description: "SUSE LINUX Enterprise-based filesystem"
```
jimmykarily: `LINUX` -> `Linux`. Still reads a bit weird (the `-` attached only to the last word) but can't think of a better description so lgtm

PR #2177, opened 2019-03-12
Updates to Vagrantfile customization scripts

f0rmiga:
## Description

- Removed trailing `m` from message.
- Removed unnecessary `Done ...` line.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -205,9 +205,8 @@ Vagrant.configure(2) do |config|
     if [ -d "#{mounted_custom_setup_scripts}/provision.d" ]; then
       scripts=($(find "#{mounted_custom_setup_scripts}/provision.d" -iname "*.sh" -executable -print | sort))
       for script in "${scripts[@]}"; do
-        echo -e "Start customization via \e[1;96m${script}m\e[0m"
+        echo -e "Running \e[1;96m${script}\e[0m"
         "${script}"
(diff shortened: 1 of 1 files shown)


Vagrantfile
```diff
@@ -205,9 +205,8 @@ Vagrant.configure(2) do |config|
     if [ -d "#{mounted_custom_setup_scripts}/provision.d" ]; then
       scripts=($(find "#{mounted_custom_setup_scripts}/provision.d" -iname "*.sh" -executable -print | sort))
       for script in "${scripts[@]}"; do
-        echo -e "Start customization via \e[1;96m${script}m\e[0m"
+        echo -e "Running \e[1;96m${script}\e[0m"
         "${script}"
-        echo -e "Done_ customization via \e[1;96m${script}m\e[0m"
```
andreas-kupries: Instead of removing this line, just moving after the loop, and dropping the script reference ?
I see how it is redundant between scripts, with the `Start customization ...` of the next script serving as the end marker for the previous. This however is not true for the last script.

f0rmiga: I see your point. I'll do it then!

f0rmiga: I also added a message before the loop so we can see when it started exactly as well, if it is not obvious that it is running the customization scripts.

andreas-kupries: LGTM now.

PR #2187, opened 2019-03-13
Fix syslog forwarding brain test

f0rmiga:
## Description

Fixes syslog forwarding brain test and other nits and lint errors.

## Test plan


Jenkinsfile
```diff
@@ -551,8 +551,9 @@ pipeline {
                     export UAA_NAMESPACE="${jobBaseName()}-${BUILD_NUMBER}-uaa"
                     export DOMAIN="${domain()}"
                     export CF_CHART="output/unzipped/helm/cf\${suffix}"
+                    log_uid=\$(hexdump -n 8 -e '2/4 "%08x"' /dev/urandom)
```
bikramnehra: Any particular reason for using `hexdump` instead of `uuidgen`?

f0rmiga: A couple:
1. It reads directly from `/dev/urandom`, and only the number of bytes it needs.
2. It's a single command (no `tr` for stripping `-` out, and no `cut` for grabbing only 16 characters).

f0rmiga:
Used `--generator=run-pod/v1` as it is deprecated not passing a generator. Also, using run-pod as there is no need to use a deployment.

PR #2201, opened 2019-03-19
Enable sle15 as default stack

thardeck:


thardeck:
We need to fix the openSUSE removal before switching the stack I suppose.

PR #2207, opened 2019-03-19
Fix for credhub brain test being skipped on CI

f0rmiga:
## Description

Credhub brain test was being skipped on CI due to output discrepancy from kubectl versions used on Vagrant for development and on CI.

The fix uses kubectl `--output go-template=` to accomplish portability.

## Test plan

Credhub brain test should not be skipped on CI.

### DIFF
--- a/src/scf-release/src/acceptance-tests-brain/test-scripts/testutils.rb
+++ b/src/scf-release/src/acceptance-tests-brain/test-scripts/testutils.rb
@@ -1,5 +1,5 @@
 #!/usr/bin/env ruby
-# This script contains helpers for running tests
+# This script contains helpers for running tests.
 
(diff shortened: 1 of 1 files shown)


src/scf-release/src/acceptance-tests-brain/test-scripts/testutils.rb
```diff
@@ -62,22 +62,22 @@ def run(*args)
     end
 end
 
-# Run the given command line, and return the output (stdout).  If errexit is
-# set, an error is raised on failure.
+# Run the given command line, and return the stadandard output.
```
andreas-kupries: nit: `stadandard` -> `standard`.

PR #2212, opened 2019-03-20
Fix the IP address cf-dev.io A record resolves to

f0rmiga:
## Description

The A record for cf-dev.io changed from 192.168.77.77 to 192.168.47.109. This PR minimizes the impact of subsequent changes to the records of cf-dev.io.

## Test plan

Vagrant should work again with cf-dev.io without any problems, as well as the updates to docs should work as expected.

### DIFF
--- a/README.md
+++ b/README.md
@@ -183,7 +183,7 @@ a working system.
       `wicked ifreload all` and wait for wicked to apply the changes.
     - `VAGRANT_DHCP`: Set this to any value when using virtual networking (as opposed to bridged networking)
       in order to let your VM receive an IP via DHCP in the virtual network. If this environment variable is
-      unset, the VM will instead obtain the IP 192.168.77.77.
+      unset, the VM will instead obtain the IP cf-dev.io points to.
 
 
 **Note:** If every role does not go green in `pod-status --watch` refer to [Troubleshooting](#troubleshooting)
@@ -241,10 +241,10 @@ The way the vagrant box is created is by making a network with a static IP on th
 This means that you cannot connect to it from some other box.
 
 ```bash
-# Attach to the endpoint (self-signed certs in dev mode requires skipping validation)
-# cf-dev.io simply resolves to the static IP 192.168.77.77 that vagrant provisions
-# This DNS resolution may fail on certain DNS providers that block resolution to 192.168.*
-# Unless you've changed the default credentials in the configuration it's admin/changeme
+# Attach to the endpoint (self-signed certs in dev mode requires skipping validation).
+# cf-dev.io resolves to the static IP that vagrant provisions.
+# This DNS resolution may fail on certain DNS providers that block resolution to 192.168.0.0/16.
+# Unless you've changed the default credentials in the configuration, it is admin/changeme.
 cf api --skip-ssl-validation https://api.cf-dev.io
 cf login -u admin -p changeme
 ```
@@ -446,7 +446,7 @@ and execute the following commands:
 make smoke
 make brain
 make scaler-smoke
-make cats 
+make cats
 ```
 
 #### How do I run a subset of SCF acceptance tests?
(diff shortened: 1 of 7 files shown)


src/scf-release/src/acceptance-tests-brain/test-resources/nfs_secgroup.json
```diff
@@ -1,11 +1,11 @@
 [
     {
-        "destination": "192.168.77.77",
+        "destination": "192.168.47.109",
```
andreas-kupries: This is just a placeholder, the actual value does not matter.
The test overwrites it with the actual IP (because it has to work under Jenkins, etc. as well)

`src/scf-release/src/acceptance-tests-brain/test-scripts/011_nfspersi_test.rb`:79:

```
# Fix IP in the various configuration files
SGROUP = "#{tmpdir}/nfs_secgroup.json"
nfs_secgroups = File.open(resource_path('nfs_secgroup.json')) { |f| JSON.load(f) }
nfs_secgroups.map! { |sg| sg.tap { |sg| sg['destination'] = SADDR } }
File.open(SGROUP, 'w') { |f| f.puts nfs_secgroups.to_json }
```

src/scf-release/src/acceptance-tests-brain/test-resources/nfs_mount.json
```diff
@@ -1,3 +1,3 @@
 {
-    "share" : "192.168.77.77/exports/foo"
+    "share" : "192.168.47.109/exports/foo"
```
andreas-kupries: See below, just a placeholder. Code overwriting it is in the tests:

```
SMOUNT = "#{tmpdir}/nfs_mount.json"
File.open(SMOUNT, 'w') { |f| f.puts({share: "#{SADDR}/exports/foo"}.to_json) }
```

README.md
```diff
@@ -241,10 +241,10 @@ The way the vagrant box is created is by making a network with a static IP on th
...
 
 ```bash
-# Attach to the endpoint (self-signed certs in dev mode requires skipping validation)
-# cf-dev.io simply resolves to the static IP 192.168.77.77 that vagrant provisions
-# This DNS resolution may fail on certain DNS providers that block resolution to 192.168.*
-# Unless you've changed the default credentials in the configuration it's admin/changeme
+# Attach to the endpoint (self-signed certs in dev mode requires skipping validation).
+# cf-dev.io resolves to the static IP that vagrant provisions.
+# This DNS resolution may fail on certain DNS providers that block resolution to 192.168.0.0/16.
+# Unless you've changed the default credentials in the configuration, it is admin/changeme.
```
btat: Minor thing: is the `it's` change to reduce use of contractions since this is a somewhat formal document? Curious because `you've` is on the same line

src/scf-release/src/acceptance-tests-brain/test-resources/nfs_secgroup.json
f0rmiga: Ah, thanks for pointing that out.

README.md
```diff
@@ -816,7 +817,8 @@ cf push pora --no-start
 cf enable-service-access persi-nfs
 
 # Create a service and bind it
-cf create-service persi-nfs Existing myVolume -c '{"share":"192.168.77.77/exports/foo"}'
+EXTERNAL_IP=$(getent hosts "cf-dev.io" | awk 'NR=1{print $1}')
+cf create-service persi-nfs Existing myVolume -c "{\"share\":\"${EXTERNAL_IP}/exports/foo\"}"
```
btat: The line following this one uses `'` instead of escaping the `"`. Might be good to keep things consistent.

README.md
f0rmiga: Indeed it is! My poor eyes caught this one, but fooled me with `you've`.

README.md
f0rmiga: I only changed this one because of the interpolation. I would prefer keeping how it is because the next line doesn't need any interpolation. It would just look more weird with the back-slashes:
```
cf bind-service pora myVolume -c "{\"uid\":\"1000\",\"gid\":\"1000\"}"
```

btat: :+1:

PR #2239, opened 2019-03-26
Replaced the runtime installation of the SUSE userspace `fuse`

andreas-kupries:
Was reaching out to unknown machines, installing an unknown version with a combination of build- and runtime installation. Prevents air-gapped deployment.

Ref: https://trello.com/c/dOiHIcaP/847-5-update-nfs-volume-release-to-176

Installation of the `fuse` package moved into the stemcells, see PR #2251.

This PR modifies the patch for the `mapfs` pre-start script to insert a skip over the installation of the pre-packaged .deb archives when `fuse` support is already installed, instead of inserting the old full installation of `fuse` support via zypper and network.

----

Testing: Build the branch and start the cluster. Run the test suites.

### DIFF
--- a/container-host-files/etc/scf/config/scripts/patches/mapfs_pre_start.sh
+++ b/container-host-files/etc/scf/config/scripts/patches/mapfs_pre_start.sh
@@ -2,32 +2,23 @@
 
 set -e
 
-PATCH_DIR=/var/vcap/jobs-src
+PATCH_DIR=/var/vcap/jobs-src/mapfs/templates
 SENTINEL="${PATCH_DIR}/${0##*/}.sentinel"
 
 if [ -f "${SENTINEL}" ]; then
   exit 0
 fi
 
-patch -d "$PATCH_DIR" --force -p1 <<'PATCH'
-From c46738e970913e75d04f0654df6033c6e0ad7d61 Mon Sep 17 00:00:00 2001
-From: Mark Yen <mark.yen@suse.com>
-Date: Wed, 30 Jan 2019 13:51:36 -0800
-Subject: [PATCH] mapfs: install: Support for SUSE
+# This adds support for SUSE-like systems to install fuse:
+# - Detect installed fuse support and skip installation of pre-packaged (ubuntu) debs
+# - Don't use `adduser`, use `useradd`.
+# - SUSE support is installed in a custom job's pre-start script, from a pre-packaged rpm.
+# - Reworks ubuntu code somewhat (modprobe specific to them, better version detection).
 
-This adds support for SUSE-like systems to install fuse:
-- Detect Ubuntu, and keep installing the pre-packaged debs
-- Detect SUSE, and use zypper to install fuse.
-- Don't use `adduser`, use `useradd`.
----
- jobs/mapfs/templates/install.erb | 39 +++++++++++++++++++++++++--------------
- 1 file changed, 25 insertions(+), 14 deletions(-)
-
-diff --git jobs/mapfs/templates/install.erb jobs/mapfs/templates/install.erb
-index 7574622..f3cba25 100644
---- jobs/mapfs/templates/install.erb
-+++ jobs/mapfs/templates/install.erb
-@@ -4,22 +4,33 @@ set -e -x
+patch -d "$PATCH_DIR" <<'PATCH'
+--- install.erb	2019-03-26 12:30:40.614236874 -0700
++++ install.erb	2019-03-26 14:39:32.565499697 -0700
+@@ -4,22 +4,37 @@
  
  echo "Installing fuse"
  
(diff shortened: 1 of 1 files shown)


src/scf-release/jobs/install-suse-fuse/templates/pre-start.erb
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash -e
+##
+set -o xtrace
+set -o errexit -o nounset
+# This installs the FUSE userspace support.
+# The required files are found in a tarball
+# provided by the `suse-fuse` package.
```
andreas-kupries: No. It is the original SUSE rpm archive.

src/scf-release/packages/suse-fuse/packaging
```diff
@@ -0,0 +1,21 @@
...
+SRC=/var/cache/zypp/packages
+
+# Clear the rpm cache. We want to see later only new files.
+zypper clean
+
+# Get fuse rpm and dependencies
+zypper --non-interactive --quiet install --download-only fuse
+
+# Get and copy the content of the rpm cache into the package-to-be.
+cp $(find /var/cache/ -type f -name '*.rpm') ${DST}/
```
andreas-kupries: Oops. The directory should be `${SRC}`

src/scf-release/packages/suse-fuse/packaging
```diff
@@ -0,0 +1,21 @@
...
+
+# Show what we packaged.
+echo ................................. packaged
+find -L $DST -type f
+
+# Force error and log
```
andreas-kupries: Nonsense comment. Remove.

src/scf-release/packages/suse-fuse/packaging
```diff
@@ -0,0 +1,21 @@
...
+set -e -x
+
+DST=${BOSH_INSTALL_TARGET}
+SRC=/var/cache/zypp/packages
+
+# Clear the rpm cache. We want to see later only new files.
+zypper clean
+
+# Get fuse rpm and dependencies
+zypper --non-interactive --quiet install --download-only fuse
```
andreas-kupries: Jenkins fails here with `16:59:57 compilation-suse-fuse > No provider of 'fuse' found.`
Have to check what set of repos are known here.
Difference to Vagrant which was ok.

andreas-kupries: Should remove the `--quiet`

src/scf-release/jobs/install-suse-fuse/spec
```diff
@@ -0,0 +1,9 @@
+---
+name: install-suse-fuse
+description: Install the SUSE userspace FUSE support.
+
+templates:
+  pre-start.erb: bin/pre-start
+
+packages:
+  - suse-fuse
```
jandubois: No need to indent list elements:

```
packages:
- suse-fuse
```

src/scf-release/packages/suse-fuse/packaging
```diff
@@ -0,0 +1,21 @@
...
+set -e -x
+
+DST=${BOSH_INSTALL_TARGET}
+SRC=/var/cache/zypp/packages
+
+# Clear the rpm cache. We want to see later only new files.
+zypper clean
+
+# Get fuse rpm and dependencies
+zypper --non-interactive install --download-only fuse
```
jandubois: I'm a bit of two minds here:

1. I think we should specify the precise package version here. A bosh release should always compile to the same bits and not incorporate whatever is latest at the time it is compiled. The Ubuntu fuse packages are embedded in the bosh release and don't change when Ubuntu updates the package.

2. Hard-coding the fuse version is a maintenance headache. We would have to update the bosh release whenever there is a CVE to this package, and we don't have a mechanism to notify us. However, the situation is only accidentally better by not hard-coding the version: we only rebuild when there is a new release of the full product.

In the end this is an issue produced by upstream. It is simply wrong to have stemcell-specific bits inside a bosh release. ☹️

src/scf-release/packages/suse-fuse/spec
```diff
@@ -0,0 +1,3 @@
+---
+name: suse-fuse
+files:
```
jandubois: I would prefer

```
files: []
```
to indicate that this is an empty list and not just `nil`, but I guess it doesn't make any difference. However, it also makes it obvious that there are not missing lines at the end that got accidentally deleted...

src/scf-release/jobs/install-suse-fuse/templates/pre-start.erb
andreas-kupries: Done with commit 838d12fad5710bd0aaa9d0a107f1fe1b0688a90f

src/scf-release/packages/suse-fuse/packaging
andreas-kupries: Done with commit 838d12fad5710bd0aaa9d0a107f1fe1b0688a90f, then superseded in commit 571a972ea60268849419855b642ed8be94ef47fa

src/scf-release/packages/suse-fuse/packaging
andreas-kupries: Done with commit 838d12fad5710bd0aaa9d0a107f1fe1b0688a90f

src/scf-release/packages/suse-fuse/packaging
andreas-kupries: Added diagnostics with commit 571a972ea60268849419855b642ed8be94ef47fa.
Currently suspect:

- Vagrant = opensuse stemcell, has predefined repos to look at
- Jenkins = sle stemcell, no repos, or other

src/scf-release/jobs/install-suse-fuse/spec
andreas-kupries: Done with commit f5ad84cb4a10919001addd629b9cafd7c85bd1a1

src/scf-release/packages/suse-fuse/spec
andreas-kupries: Done with commit 795b58f9da96e0e6342529f0c244d37648fd4a6d

src/scf-release/packages/suse-fuse/packaging
andreas-kupries: I can see both sides and am also two-minded. No decision yet. Lets see if in the end the fuse version I get differs for opensuse and sle setups.

src/scf-release/packages/suse-fuse/packaging
andreas-kupries: Confirmed. OpenSuse stemcell knows 3 repos, opensuse distribution + update, and stemcell repo.
SLE stemcell only knows the stemcell repo.

andreas-kupries: By adding the relevant repos I I will also get the same fuse for both I guess. Using a fixed version should therefore be no problem.

andreas-kupries: Ref commit c6471b01a4c78e864a42e30e27e0a898eba062b2, trialing

andreas-kupries: Past the build stage now with commits 23cd7f909e4a886a1a3337403a91cabe999f79c6 and 8f344a3cb8b112b3bbaad9ebb76b03dc866a5a6e fixing a few zypper options.

src/scf-release/packages/suse-fuse/packaging
andreas-kupries: outdated with the move of fuse install to stemcells.

container-host-files/etc/scf/config/scripts/patches/mapfs_pre_start.sh
```diff
@@ -2,32 +2,23 @@
...
   exit 0
 fi
 
-patch -d "$PATCH_DIR" --force -p1 <<'PATCH'
-From c46738e970913e75d04f0654df6033c6e0ad7d61 Mon Sep 17 00:00:00 2001
-From: Mark Yen <mark.yen@suse.com>
-Date: Wed, 30 Jan 2019 13:51:36 -0800
-Subject: [PATCH] mapfs: install: Support for SUSE
+# This adds support for SUSE-like systems to install fuse:
+# - Detect installed fuse support and skip installation of pre-packaged (ubuntu) debs
```
f0rmiga: [nit] This is the only point without punctuation.

PR #2240, opened 2019-03-27
acceptance-tests-brain: syslog: Wait for ContainerCreating

mook-as:
It looks like sometimes we end up trying to get logs from the syslog container while it's still in `ContainerCreating`, somehow (even though all conditions are ready); this probably happens because there is no conditions yet.

Explicitly check the container phase to work around this.

Also make it more obivous (in the error message) how to set up `SCF_LOG_HOST` to get this to work.

Trello: https://trello.com/c/Q7XKcyhx/

### DIFF
--- a/src/scf-release/src/acceptance-tests-brain/test-scripts/017_syslog_forwarding_test.rb
+++ b/src/scf-release/src/acceptance-tests-brain/test-scripts/017_syslog_forwarding_test.rb
@@ -45,16 +45,18 @@ Timeout::timeout(ENV.fetch('TESTBRAIN_TIMEOUT', '600').to_i - 60) do
     SCF_LOG_HOST = ENV.fetch('SCF_LOG_HOST', '')
     SCF_LOG_PORT = ENV.fetch('SCF_LOG_PORT', '514')
     SCF_LOG_PROTOCOL = ENV.fetch('SCF_LOG_PROTOCOL', 'tcp')
-    if SCF_LOG_HOST.empty?
-        message = "SCF_LOG_HOST not set"
-        STDERR.puts "\e[0;1;31m#{message}\e[0m"
-        show_env
-        fail message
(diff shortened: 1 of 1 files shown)


src/scf-release/src/acceptance-tests-brain/test-scripts/017_syslog_forwarding_test.rb
```diff
@@ -45,21 +45,23 @@ def show_env
...
+
+    KUBERNETES_DOMAIN_SUFFIX = ".#{$KUBERNETES_NAMESPACE}.svc.#{ENV['KUBERNETES_CLUSTER_DOMAIN']}"
+
     if SCF_LOG_HOST.empty?
-        message = "SCF_LOG_HOST not set"
+        message = "SCF_LOG_HOST not set; expected to end with cluster domain (#{KUBERNETES_DOMAIN_SUFFIX})"
         STDERR.puts "\e[0;1;31m#{message}\e[0m"
         show_env
-        fail message
+        exit 99
```
andreas-kupries: Is the change from `fail` to `exit 99` intended, and if yes, why ? (The new code feel less clear).

mook-as: Yes, because that's our magic code for `skip` (instead of `fail`). I should probably add a comment though?

andreas-kupries: We have a command for skip in the testutils.

mook-as: Oh! Thanks! Force-pushed using that. (And dropped the second one, since if it's set we should probably fail instead of skip)

PR #2245, opened 2019-03-28
brains: minibroker: emit logs on failure

mook-as:
When a run fails, print out logs of the minibroker & -pods containers, in case they contain information that can help us figure out what went wrong.

### DIFF
--- a/src/scf-release/src/acceptance-tests-brain/test-scripts/minibroker_helper.rb
+++ b/src/scf-release/src/acceptance-tests-brain/test-scripts/minibroker_helper.rb
@@ -51,6 +51,14 @@ class MiniBrokerTest
     attr_lazy(:service_plan_id) { |inst| inst.service_plans['resources'].first['entity']['name'] }
 
 
+    def print_logs_in_namespace(ns)
+        capture("kubectl get pods --namespace #{ns} -o name").split.each do |pod|
+            capture("kubectl get --namespace #{ns} #{pod} -o jsonpath='{.spec.containers[*].name}'").split.each do |container|
+                run "kubectl logs --namespace #{ns} #{pod} --container #{container}"
+            end
+        end
+    end
+
     # Run the minibroker test.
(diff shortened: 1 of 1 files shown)


src/scf-release/src/acceptance-tests-brain/test-scripts/minibroker_helper.rb
```diff
@@ -51,6 +51,14 @@ def initialize(service_type, service_port)
     attr_lazy(:service_plan_id) { |inst| inst.service_plans['resources'].first['entity']['name'] }
 
 
+    def print_logs_in_namespace(ns)
+        capture("kubectl get pods --namespace #{ns} -o name").split.each do |pod|
```
f0rmiga: Use `--output` instead of `-o`, please.

src/scf-release/src/acceptance-tests-brain/test-scripts/minibroker_helper.rb
```diff
@@ -51,6 +51,14 @@ def initialize(service_type, service_port)
     attr_lazy(:service_plan_id) { |inst| inst.service_plans['resources'].first['entity']['name'] }
 
 
+    def print_logs_in_namespace(ns)
```
f0rmiga: How about `print_all_container_logs_in_namespace`?

src/scf-release/src/acceptance-tests-brain/test-scripts/minibroker_helper.rb
```diff
@@ -63,6 +71,13 @@ def run_test
 
             at_exit do
                 set errexit: false do
+                    unless @success
+                        # Print out logs
```
f0rmiga: Unnecessary comment?

mook-as: It made more sense before I put everything in its own function :D

PR #2249, opened 2019-03-28
Start streaming logs faster for test containers

f0rmiga:
Sometimes logs are missed between the `kubectl logs` and `kubectl attach`. The changes here make the logs to start to be streamed much quicker.


f0rmiga:
- We have to wait for the container to be running - the changes in this patch will make this check much faster.
- I'm not the original author of the script, but I guess we can do `kubectl logs --follow`?
- `kubectl logs` does not require `--container` because it defaults to the single container in the pod.

PR #2258, opened 2019-04-01
Brain test NFS persi - Get more information out of it

andreas-kupries:
Ref: https://trello.com/c/UL1n6gO0/1007-nfs-persi-brain-failing-on-nightly-caasp3

Extended narrative tracing around the startup of the nfs-test-server pod.

### DIFF
--- a/src/scf-release/src/acceptance-tests-brain/test-scripts/testutils.rb
+++ b/src/scf-release/src/acceptance-tests-brain/test-scripts/testutils.rb
@@ -145,11 +145,20 @@ def wait_for_namespace(namespace, sleep_duration=10)
             name, readiness, status, restarts, age = line.split
             next if status == 'Completed'
             next if status == 'Running' && /^(\d+)\/\1$/ =~ readiness
+            puts "# Waiting for: #{line}"
             ready = false
         end
         break if ready
         sleep sleep_duration
     end
+end
+
+# Show the status of a Kubernetes namespace
+def show_pods_for_namespace(namespace)
+  output = capture("kubectl get pods --namespace #{namespace} --no-headers")
+  output.each_line do |line|
+    puts "# Space: #{line}"
+  end
 end
 
 # Wait for a cf service asynchronous operation to complete.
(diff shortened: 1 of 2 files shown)


src/scf-release/src/acceptance-tests-brain/test-scripts/testutils.rb
```diff
@@ -145,13 +145,22 @@ def wait_for_namespace(namespace, sleep_duration=10)
...
         break if ready
         sleep sleep_duration
     end
 end
 
+# Show the status of a Kubernetes namespace
+def show_pods_for_namespace(namespace)
+  output = capture("kubectl get pods --namespace #{namespace} --no-headers")
+  output.each_line do |line|
+    puts "# Space: #{line}"
```
mook-as: What's with the prefix? (And why isn't letting `kubectl get pods` print its table directly good enough?)

src/scf-release/src/acceptance-tests-brain/test-scripts/testutils.rb
```diff
@@ -145,13 +145,22 @@ def wait_for_namespace(namespace, sleep_duration=10)
             name, readiness, status, restarts, age = line.split
             next if status == 'Completed'
             next if status == 'Running' && /^(\d+)\/\1$/ =~ readiness
+            puts "# Waiting for: #{line}"
```
mook-as: This probably gives too many lines in normal testing? Can we do tricks with `\r` to not spam the output too much?

src/scf-release/src/acceptance-tests-brain/test-scripts/testutils.rb
andreas-kupries: Prefix - Wanted to distinguish output from whatever is before and after.

Print table directly - Right, that should work as well. I guess I was on one side coming from slimming down the def from the original I derived it from, and not seeing far enough.

andreas-kupries: Will change.

src/scf-release/src/acceptance-tests-brain/test-scripts/testutils.rb
andreas-kupries: In normal testing we should not see anything except when we are explicitly verbose.
IMHO this can stay as-is

src/scf-release/src/acceptance-tests-brain/test-scripts/testutils.rb
andreas-kupries: Addressed with commit 3d4f0c814edeb834cfd312ace6db689d119889b1

PR #2283, opened 2019-04-08
Minibroker Postgres brain test assertion cleanup

f0rmiga:
## Description

Removed unnecessary calls on the Minibroker Postgres brain test, as well as added a missing `--fail` to the `/todos` call.

## Test plan

Should pass CI.


f0rmiga:
@mook-as Yes, it's more of a debugging statement that can be added as needed.

PR #2284, opened 2019-04-08
Brain Minibroker install rely on Helm --wait

f0rmiga:
## Description

Rely on Helm's `--wait` for the Minibroker installation on the brain test.

## Test plan


f0rmiga:
Merging as it passed brain tests.

PR #2285, opened 2019-04-08
Brain tests - don't wait for namespace deletion

f0rmiga:
## Description

The Minibroker related namespace deletion can happen on the background. In some cases, waiting for the namespace to be deleted can cause the tests to timeout.

## Test plan


f0rmiga:
Merging as it passed brain tests.

PR #2291, opened 2019-04-09
Fix kube ready state check

svollath:
## Description

This addresses the following issues (esp. when run in the pipeline):

- auth test will fail for nonauth, and log respectively (currently the first test just failed for noauth, saying you need to have 4/4 kube-dns pods)

- remove explicit pod numbers from status messages, as 1. they were wrong, 2. we don't check for a pod number, but only if requested ones are ready - therefore "N/N"

- storage class test passed, when kube couldn't authenticate - now fails here as well

## Test plan

run "./kube-ready-state-check.sh kube" (didn't verified for e.g. AWS/EKS)

### DIFF
--- a/bin/dev/kube-ready-state-check.sh
+++ b/bin/dev/kube-ready-state-check.sh
@@ -88,16 +88,16 @@ if having_category node ; then
     status "docker info should not show aufs"
 fi
 
-# kube-dns shows 4/4 ready
+# kube-dns shows all pods ready
 if having_category kube ; then
     kubectl get pods --namespace=kube-system --selector k8s-app=kube-dns 2> /dev/null | grep -Eq '([0-9])/\1 *Running'
-    status "kube-dns should be running (show 4/4 ready)"
+    status "all kube-dns pods should be running (show N/N ready)"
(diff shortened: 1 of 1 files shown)


bin/dev/kube-ready-state-check.sh
```diff
@@ -88,16 +88,22 @@ if having_category node ; then
     status "docker info should not show aufs"
 fi
 
-# kube-dns shows 4/4 ready
+# kube auth
+if having_category kube ; then
+    kubectl auth can-i get pods --namespace=kube-system 2>&1>/dev/null
```
HartS: Is this intended to suppress stderr? Effectively it's just the same as:

```
kubectl auth can-i get pods --namespace=kube-system > /dev/null
```

If we want to suppress stderr and stdout, we can just use:
```
kubectl auth can-i get pods --namespace=kube-system &>/dev/null
```

svollath: sure

HartS:
:+1: LGTM, I can't merge it though

PR #2295, opened 2019-04-09
Feature flags comments

andreas-kupries:
## Description

Ref: https://trello.com/c/P2Inj8sM/1020-add-feature-flag-details-to-valuesyaml
Associated fissile PR https://github.com/cloudfoundry-incubator/fissile/pull/490 (merged)

Updated to a version of fissile which injects better comments about feature flags into its helm charts.

## Test plan

Check the branch out, run `make vagrant-prep` in a proper vagrant box, then compare the `output/helm` directory against an older saved helm chart from before this change.


andreas-kupries:
Will rebase, this is on the wrong branch

PR #2298, opened 2019-04-09
brains/minibroker-mongo: disable pod liveness probe

mook-as:
## Description

On slow clusters, we could take too long to deploy the mongo pod; if kube happens to kill the pod at a bad time, we could end up with an unrecoverable state.

The proper fix is to change the mongodb helm chart to use an init container, instead of just running the initial db setup in a script.

This also enables extra logging, in case that helps.

## Test plan

The acceptance-tests-brain test `016_minibroker_mongodb_test` should be run a very slow cluster.

### DIFF
--- a/src/scf-release/src/acceptance-tests-brain/test-scripts/016_minibroker_mongodb_test.rb
+++ b/src/scf-release/src/acceptance-tests-brain/test-scripts/016_minibroker_mongodb_test.rb
@@ -4,6 +4,8 @@ require_relative 'minibroker_helper'
 
 tester = MiniBrokerTest.new('mongodb', '27017')
 tester.service_params = {
+    image: { debug: true },
+    livenessProbe: { enabled: false },
     mongodbDatabase: random_suffix('database'),
     mongodbUsername: random_suffix('user'),
     mongodbPassword: random_suffix('pass'),


mook-as:
I seem to be able to pass consistently with a newer version of mongodb: `--set env.KUBERNETES_REPO=https://mook-as.github.io/helm-charts/minibroker-brokers/ ` (chart version 5.16.0, app version 4.0.8).

andreas-kupries:
Does the `pass consistently with ...` include this PR ?

src/scf-release/src/acceptance-tests-brain/test-scripts/016_minibroker_mongodb_test.rb
```diff
@@ -4,6 +4,8 @@
 
 tester = MiniBrokerTest.new('mongodb', '27017')
 tester.service_params = {
+    image: { debug: true },
```
f0rmiga: I'm not favorable of leaving this param for debugging.

f0rmiga:
I'm not sure if I understand. Does the combination of the code changed on this PR with the newer chart version makes the slow cluster happy?

mook-as:
Essentially, the clusters we're testing on are slow. What appears to be happening is that the container appears to be getting murdered halfway through startup (the database gets created, but it never got enough for the user to be created). When the pod restarts, it sees that it already has old data, and skips doing the initial setup (… and no user gets created), even though it reports the expected credentials.

The newer chart version appears to have better logic around detecting if the existing data is correctly initialized.  Also, it enabled auth even in the existing-data case.

Hmm, app version 4.0.8 would be under MongoDB's new funky non-OSS license. I may need to try to use an older version instead.

PR #2304, opened 2019-04-10
Fix autoscaler.cf.api property

f0rmiga:
## Description

Use the internal service URL for App Autoscaler to talk to CloudController.

## Test plan

Autoscaler should be deploy-able at SCF installation time.


andreas-kupries:
LGTM. Starts fine on my vagrant. Jenkins was still on the old scheme, deploy autoscaler on upgrade first. Modified that to deploy autoscaler with initial deployment of cluster. Jenkins is running on that now.

PR #2305, opened 2019-04-10
Improved ingress

f0rmiga:
## Description

- Aligned better with the practices seen on stable/charts.
- Allowed passing `ingress.tls.crt` and `ingress.tls.key`.
- Allowed passing `ingress.annotations`, and handling default annotations.
- Disabled the possibility of setting `UAA_PORT` environment variable by the user.
- Hard-coded GoRouter port as it is not modifiable.
- Updated make scripts to reflect the changes.

Ties to:
- https://github.com/cloudfoundry-incubator/fissile/pull/491
- https://github.com/SUSE/uaa-fissile-release/pull/152

## Test plan

Set SCF to use the ingress controller:
```txt
export INGRESS_CONTROLLER="nginx"
```

Run SCF as per usual:
```txt
make run
```

Check the UAA and CF Ingress resources:
```txt
kubectl get ingress uaa --namespace uaa -o json | jq -C .metadata.annotations
kubectl get ingress cf --namespace cf -o json | jq -C .metadata.annotations
```

Upgrade with helm setting/overriding some annotations:
```txt
make/uaa/upgrade --set "ingress.annotations.foo=bar"
make/upgrade --set "ingress.annotations.foo=bar" --set "ingress.annotations.nginx\.ingress\.kubernetes\.io\/proxy-body-size=128m"
```

Check that the annotations were set/overridden properly:
```txt
kubectl get ingress uaa --namespace uaa -o json | jq -C .metadata.annotations
kubectl get ingress cf --namespace cf -o json | jq -C .metadata.annotations
```


chart-parts/ingress.yaml
```diff
@@ -1,51 +1,43 @@
...
-# instructions to the operator.
-##
-# tls.crt: {{ default "" .Values.secrets.ROUTER_SSL_CERT     | b64enc | quote }}
-# tls.key: {{ default "" .Values.secrets.ROUTER_SSL_CERT_KEY | b64enc | quote }}
-
 apiVersion: "v1"
 kind: "Secret"
 type: kubernetes.io/tls
 metadata:
-  name: "{{ .Values.services.ingress.class }}-ingress-tls"
```
mook-as: Any reason we're making this change? The commit message doesn't tell me.

chart-parts/ingress.yaml
```diff
@@ -1,51 +1,43 @@
...
 ---
-# The cert and the key cannot be filled automatically. If we could it
-# would roughly look like the lines below. As is it is the
-# responsibility of the operator to retrieve the values of these
-# secrets from the deployed SCF and update this secret with the actual
-# values. And the responsibility of the developer to supply the proper
-# instructions to the operator.
-##
-# tls.crt: {{ default "" .Values.secrets.ROUTER_SSL_CERT     | b64enc | quote }}
-# tls.key: {{ default "" .Values.secrets.ROUTER_SSL_CERT_KEY | b64enc | quote }}
```
mook-as: The comment needs to be updated, but still needs to live somewhere.

chart-parts/ingress.yaml
```diff
@@ -1,51 +1,43 @@
...
+  tls.key: {{ .Values.ingress.tls.key | default "" | b64enc | quote }}
 ---
-# This ingress specifies routing and access for the cloud controller
-# public SCF service, i.e. "api.<domain>" and other services in that
-# domain hierarchy.
-#
-# Note that the two other services (tcp routing, app ssh access) are
-# not suitable to routing via http, and not specified. IOW they are
-# TCP, which is not supported by the current ingress specification.
-# Although specific controllers may do so in a custom manner.
```
mook-as: The comment doesn't look that obsolete?

make/include/defaults
```diff
@@ -13,7 +13,7 @@ export CF_NAMESPACE="${NAMESPACE}"
 : "${DOMAIN:=cf-dev.io}"
 : "${UAA_NAMESPACE:=uaa}"
 : "${UAA_HOST:=uaa.${DOMAIN}}"
-: "${UAA_PORT:=2793}"
+export UAA_PORT="2793"
```
mook-as: This is no longer customizable? No chance of somebody running against a non-default UAA?

chart-parts/ingress.yaml
f0rmiga: 1. We don't have `.Values.services.ingress.class` anymore in the code I present in this PR.
2. It is a generic TLS secret for any ingress controller.

chart-parts/ingress.yaml
f0rmiga: I removed because I don't believe it belongs to the template code itself, but rather in the Wiki and official docs, as they already are.

make/include/defaults
f0rmiga: Using our very Vagrant-tied make scripts? No.

chart-parts/ingress.yaml
f0rmiga: I removed it here for the same reason I stated below. I don't think it belongs to the template.

f0rmiga: Brought a relevant comment back.

chart-parts/ingress.yaml
f0rmiga: Added comment back.

PR #2307, opened 2019-04-10
SATS -> BRAIN on Jenkinsfile

f0rmiga:
## Description

Left over? SATS is not used to reference BRAIN tests.

## Test plan

Not applicable.


f0rmiga:
At least it is not an acronym that can be confused with a 23 year old acronym.

PR #2318, opened 2019-04-12
Reworked klog to continue on bad pods, containers.

andreas-kupries:
## Description

Ref: https://trello.com/c/omFBqdlS/1031-klogsh-should-not-error-out-when-containers-are-not-running

Moved all kube operations into functions with nice names.
Prevent abort when retrieving kube logs (`|| true`)
Added narrative output, let users see the script working instead of waiting a few minutes with no indication of operation.
Note: Hoisted determination of pod phase out of container loop, need this only per pod, not per container.

## Test plan

- Start a cluster where one or more pods are CrashLoopBackOff, or not fully started.
- Run klog.sh against the cluster.
- See it complete instead of aborting on the bad pods.

### DIFF
--- a/container-host-files/opt/scf/bin/klog.sh
+++ b/container-host-files/opt/scf/bin/klog.sh
@@ -24,56 +24,107 @@ fi
 NS=${1-scf}
 DONE="${KLOG}/${NS}/done"
 
+# Prevent bail out
 if [ "${FORCE}" == "1" ] ; then
   rm -f "${DONE}" 2> /dev/null
 fi
 
-function get_phase() {
+# Bail out early when already done
+if [ -f "${DONE}" ]; then
+    printf "Already done\n"
+    exit
+fi
+
+function get_pod_phase() {
   kubectl get pod "${POD}" --namespace "${NS}" --output=jsonpath='{.status.phase}'
 }
 
 function check_for_log_dir() {
-  kubectl exec "${POD}" --namespace "${NS}" --container "${CONTAINER}" -- bash -c "[ -d /var/vcap/sys/log ]" 2> /dev/null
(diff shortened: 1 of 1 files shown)


container-host-files/opt/scf/bin/klog.sh
```diff
@@ -24,56 +24,107 @@ fi
 NS=${1-scf}
 DONE="${KLOG}/${NS}/done"
 
+# Prevent bail out
```
f0rmiga: Missing punctuation here and below. There are also a couple comments inside the function. For consistency, how about putting them out of the function?

mook-as:
> For me, it is confusing having a bunch of function declarations in the middle of the procedural logic of the script. Would it be possible to create a directory `//container-host-files/opt/scf/bin/klog` (even though I don't like the base location), and then have separated files for functions and `klog.sh` only containing the procedural logic?

That might be a bad idea, since that makes the script no longer self-contained; shipping it to people who have clusters that need troubleshooting is harder.

andreas-kupries:
The functions are bunched near the top, the only logic before them is command line processing and the bail out. I considered that acceptable, so I did not change this structure, except adding more functions to the block and thus making the main loop more readable.

I am +1 on keeping the script a single entity for easier distribution.

container-host-files/opt/scf/bin/klog.sh
andreas-kupries: Done with commit 51b44ecf987cdde4f6676f0e17e92bc2dda5d189

andreas-kupries:
The brain failure I see is the known mongodb issue, and not relevant to this PR.

f0rmiga:
I see how it can be problematic. I accept as is then.

PR #2319, opened 2019-04-12
acceptance-tests-brain: minibroker: dump more info on failure

mook-as:
This is a debugging PR to dump more information about containers when a minibroker acceptance-tests-brain test fails.


mook-as:
Huh, no permissions for looking at PVs. Didn't end up with any interesting info anyway.

PR #2320, opened 2019-04-12
Add uaa feature to role manifest to allow standalone install

jandubois:
This PR includes a UAA role directly in the SCF chart to simplify deployment. It is disabled by default, so should be completely backwards compatible.

Use `helm install ... --enable.uaa=true` to to deploy with the embedded UAA.

This PR also adds an option to `Jenkinsfile` to run deployment with the embedded UAA instead of the separate chart from the submodule.


container-host-files/etc/scf/config/role-manifest.yml
```diff
@@ -4364,6 +4507,11 @@ variables:
   options:
     description: Concatenation of trusted CA certificates to be made available on
       the cell.
+- name: UAADB_PASSWORD
```
f0rmiga: Could you name it `UAA_DATABASE_PASSWORD` or `UAADB_DATABASE_PASSWORD` to match the other database passwords?

make/run
```diff
@@ -26,16 +26,33 @@ done
...
+    --namespace "${NAMESPACE}"
+    --values "bin/settings.yaml"
+    --set "kube.storage_class.persistent=${STORAGE_CLASS}"
+)
 
-printf "Waiting for uaa cert to be generated by the uaa chart...\n"
-until has_internal_ca ; do
-    sleep 1
-done
+if [ "${EMBEDDED_UAA:-false}" != "true" ]; then
```
f0rmiga: It would make more sense to have `EMBEDDED_UAA` in the https://github.com/SUSE/scf/blob/develop/make/include/defaults.

container-host-files/etc/scf/config/role-manifest.yml
jandubois: I could, but then we should change it in the UAA chart as well: https://github.com/SUSE/uaa-fissile-release/blob/develop/container-host-files/etc/scf/config/role-manifest.yml#L560-L564

But then you have to worry about the upgrade scenario. So not really in favour of changing the name.

make/run
jandubois: I don't know, we have plenty of tests like this in the scripts:

```
if [ -z "${INGRESS_CONTROLLER:-}" ]; then
```

And this is just the same thing: checking if the variable has been set or not. Splitting things up into multiple files does not make the code easier to understand.

PR #2321, opened 2019-04-12
Improve Jenkins log on failure

f0rmiga:
## Description

Execute `kubectl get all --namespace {cf,uaa} --output wide` for a better insight of what was going on on the k8s side of things when tests fail.

## Test plan

Fail a Jenkins test (should happen as of now).


f0rmiga:
As we can see, my code failed and Github was not notified. I guess I agree with you!

PR #2323, opened 2019-04-15
acceptance-tests-brain: wait for ready statefulsets better

mook-as:
## Description

When looking to see if StatefulSets are ready, we should fall back in the case where the readyReplicas count doesn't exist yet.

## Test plan

Run `011_nfspersi_test.rb` on a slow cluster (that should time out) and check for a (lack of) misleading lines:
> `error: error executing template "\"{{ eq .status.replicas .status.readyReplicas }}\"": template: output:1:4: executing "output" at <eq .status.replicas ...>: error calling eq: invalid type for comparison`


mook-as:
I… seems to have gotten confused and fixed this over in #2336

PR #2334, opened 2019-04-17
Dedicated Fissile cache disk

f0rmiga:
## Description

- Refactored `Vagrantfile` to simplify the logic and handle provider-specific shell provisioning.
- Added provider-specific shell provisioning for formatting and creating the filesystem on the cache disk.

## Test plan

Check that a second disk was mounted on `/home/vagrant/.fissile` (use the `df` command). It should be used by Fissile to store the cache.


andreas-kupries:
Just to clarify:
- This change adds a second disk to hold the in-VM `.fissile` compilation cache.
- It does not change the fact that the host-side cache is copied into the VM at creation time, nor does it change the saving of compilation results from VM to host.
- The new disk is as ephemeral as the main disk it was split off from, i.e. it is created and destroyed with the box.

(I had this side thought if we could make this disk persistent and then treat it as the one cache, with no host side cache/directory any longer)

f0rmiga:
@andreas-kupries Interesting notes. I agree with you, but I think we should consider making the disk permanent as a second step.

andreas-kupries:
Oh yes, after I understood that the new disk was not persistent I definitely placed making it so into the category of `future nice to have`. The clarifications was me explaining to myself how this worked, with an implicit question if I was correct.

f0rmiga:
@andreas-kupries You are completely correct. And thanks for making this PR more clear!

PR #2362, opened 2019-04-29
Created and assigned separate service accounts for tests

f0rmiga:
## Description

Fixed when a service account was referenced by more than one CF role, the SA definition was included in the Helm chart and not in the bosh-task Kube config file.

## Test plan

Build and check that `//output/kube/bosh-task/acceptance-tests-brain.yaml` and `//output/kube/bosh-task/sync-integration-tests.yaml` both include its own service account.

Check that the generated helm templates don't include test service accounts, roles, cluster roles, and bindings to these roles.

### DIFF
--- a/container-host-files/etc/scf/config/role-manifest.yml
+++ b/container-host-files/etc/scf/config/role-manifest.yml
@@ -2458,7 +2458,7 @@ configuration:
         roles: [configgin-role, test-role]
         cluster-roles: [nonprivileged, test-cluster-role]
       tests-sits:
-        roles: [configgin-role, test-role]
+        roles: [configgin-role]
         cluster-roles: [nonprivileged, test-cluster-role]
       garden-runc:
         roles: [configgin-role]
--- a/container-host-files/etc/scf/config/role-manifest.yml
+++ b/container-host-files/etc/scf/config/role-manifest.yml
@@ -1797,7 +1797,7 @@ instance_groups:
             max: 1
           flight-stage: manual
           capabilities: []
-          service-account: tests
+          service-account: tests-brain
 - name: acceptance-tests
   type: bosh-task
   tags:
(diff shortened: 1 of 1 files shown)


container-host-files/etc/scf/config/role-manifest.yml
```diff
@@ -2454,7 +2454,10 @@ configuration:
       secret-generator:
         roles: [configgin-role, secrets-role]
         cluster-roles: [nonprivileged]
-      tests:
+      tests-brain:
+        roles: [configgin-role, test-role]
```
mook-as: This still generates `test-role` as part of the chart, rather as part of the task definition.

andreas-kupries: IOW the `test-role` must be split into `test-brain-role` and `test-sits-role` as well, right ?

f0rmiga: Right!

f0rmiga: SITS don't need its own Role. The test ClusterRole is sufficient for it to be able to create a port-forward. I removed it and renamed `test-role` to `test-role-brain` so it is clearer that it shouldn't be reused.

andreas-kupries:
Huh. Looking over things I suspect not.

If removal of this cluster role from the manifest is not possible it will require a change to fissile.
In other words, either when cluster roles got added it was forgotten to do the special handling like done for roles, or when the special handling was added cluster roles were forgotten.

container-host-files/etc/scf/config/role-manifest.yml
```diff
@@ -2454,8 +2454,11 @@ configuration:
       secret-generator:
         roles: [configgin-role, secrets-role]
         cluster-roles: [nonprivileged]
-      tests:
-        roles: [configgin-role, test-role]
+      tests-brain:
+        roles: [configgin-role, test-role-brain]
+        cluster-roles: [nonprivileged, test-cluster-role]
```
mook-as: Got out-of-band confirmation that we want to get rid of `test-cluster-role` too.

andreas-kupries: Ok. @f0rmiga please ping me after this is done, and I will recheck in my vagrant

PR #2363, opened 2019-04-29
Default UAA_CA_CERT to DST Root CA X3

f0rmiga:
## Description

Default `UAA_CA_CERT` to the DST Root CA X3, which IdenTrust used to cross-sign the Let's Encrypt intermediate certificate.

## Test plan

On a deployment, without passing, check that the `UAA_CA_CERT` is set to the value provided in this PR.


f0rmiga:
I am closing as it is not the desired behaviour.

PR #2365, opened 2019-04-30
Merge from master as 2.16.2

satadruroy:
Commit merge from master (2.16.2)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,28 @@
 # Changelog
 
-## [2.16.1] - 2019-02-16
+## [2.16.2] - 2019-04-30
+
+## Changed
+- Bumped binary-buildpack to 1.0.32
+- Bumped CAPI to 1.79.0
+- Bumped dotnet-core-buildpack to 2.2.10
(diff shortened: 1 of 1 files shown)


aegershman:
By the way are these releases no longer going to contain the latest helm charts?

jandubois:
These are really just release candidates. The helm charts will be published once they have been validated by QA testing.

aegershman:
Gotcha. In the interim, is there a way I can generate the helm chart from source myself? E.g. is there a `make` command that'll output a usable (albeit 'non-official') chart that would normally be attached as part of the release or available in the official helm repo?

jandubois:
You will need to build the whole release yourself from source. This will build the corresponding chart as well (under `output/helm`). You can then push the images to your own registry or own docker hub account (I believe `make publish` should help with that, but haven't used it in a while), and then set `kube.registry....` to point to your own registry.

You don't have access to the CI images, which will be copied from an internal registry to the public registry when the charts are published, so even if you could build a chart with the correct image tags, it would still not work without you building and publishing the images as well.

PR #2369, opened 2019-05-01
Vagrantfile: expose additional environments to the VM

mook-as:
## Description

This exposes all environment variables starting with `FISSILE_`, `SCF_`, and `VAGRANT_` to the VM.  This is useful for custom setup scripts.

`VAGRANT_OLD_ENV_` is skipped because vagrant puts all environment variables it starts with there.

## Test plan

```sh
host$ export SCF_BLAHBLAH=blah
host$ vagrant up
host$ vagrant ssh
vagrant$ echo $SCF_BLAHBLAH
```

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -201,7 +201,9 @@ Vagrant.configure(2) do |config|
   SHELL
 
   # Provision the custom config scripts and personal setup.
-  config.vm.provision "shell", privileged: false, inline: <<-SHELL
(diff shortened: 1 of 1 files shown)


mook-as:
Turns out this is a bad idea anyway, because if you `direnv allow` the checkout, then it will pick up the fissile directories in the VM and do things like trying to build in `/home/user/…` (instead of `/home/vagrant/…`)

PR #2401, opened 2019-05-14
Update CF bump instructions

mook-as:
## Description

Our CF bumping instructions still refer to submodules.

Also, add a script to check UAA clients to help it stay in sync.

## Test plan

Test a bump.

For the UAA client thing: Just run the script.


f0rmiga:
I'm taking this as a reviewer while doing the bump to 7.x.

PR #2405, opened 2019-05-14
Add --recreate-pods to make/upgrade

f0rmiga:
## Description

Pass `--recreate-pods` to helm upgrade.

## Test plan

The upgrade should recreate all pods.


make/upgrade
```diff
@@ -29,8 +29,9 @@ stampy "${GIT_ROOT}/scf_metrics.csv" "$0" make-run::upgrade start
 UAA_CA_CERT="$(get_secret "${UAA_NAMESPACE}" "uaa" "INTERNAL_CA_CERT")"
 
 helm_args=(
+    --recreate-pods
```
mook-as: We should document why we want this, so we don't remove it again.

PR #2408, opened 2019-05-15
Fixed Vagrantfile when VirtualBox is not installed

f0rmiga:
## Description

Use the vagrant directory to store the disk files.

## Test plan

Provision a Vagrant machine on a Linux host with libvirt, that doesn't have VirtualBox installed. And provision with VirtualBox as well to make sure it works as expected.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -82,8 +82,8 @@ Vagrant.configure(2) do |config|
 
     vb.customize ['modifyvm', :id, '--paravirtprovider', 'minimal']
 
-    default_machine_folder = `VBoxManage list systemproperties | grep "Default machine folder"`
-    vb_machine_folder = default_machine_folder.split(':')[1].strip()
+    default_machine_folder = `(which VBoxManage 1> /dev/null 2> /dev/null && (VBoxManage list systemproperties | grep "Default machine folder" | sed "s/Default machine folder://")) || echo "/tmp"`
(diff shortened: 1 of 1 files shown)


Vagrantfile
```diff
@@ -82,8 +82,8 @@ Vagrant.configure(2) do |config|
 
     vb.customize ['modifyvm', :id, '--paravirtprovider', 'minimal']
 
-    default_machine_folder = `VBoxManage list systemproperties | grep "Default machine folder"`
-    vb_machine_folder = default_machine_folder.split(':')[1].strip()
+    default_machine_folder = `(which VBoxManage 1> /dev/null 2> /dev/null && (VBoxManage list systemproperties | grep "Default machine folder" | sed "s/Default machine folder://")) || echo "/tmp"`
+    vb_machine_folder = default_machine_folder.strip()
```
andreas-kupries: From chat ... Better to check early in the section if VBoxManage exists, and if not, return.

andreas-kupries: Good error message too, like "Virtualbox provider requires installation of VBoxManage" etc. That way somebody who forgot to install the prequisites of the provider gets a reminder.

f0rmiga: I'm going with a different approach. Let's store the VirtualBox disks under `<project_root>/.vagrant/disks`.

PR #2409, opened 2019-05-16
instance_groups/api_group: patch to use squash migrations

mook-as:
## Description

As MySQL / MariaDB does *not* support transactions around schema modifications, we still need to do something about the high frequency of database migration failures (migrating from an empty database) because something hiccuped in the middle.  So this patch squashes all the existing schema migrations (up to SCF 2.16.4 / capi-release 1.79.0) into one giant migration.  The initial migrations will also drop any existing tables (so if we _do_ get interrupted we still have a good chance of recovery).  This is of course not done if the last table to be created already exists (so we don't lose data).

## Test plan

Start a cluster, and break the MySQL connection to the database halfway through migration (possibly by terminating the switchboard process, or editing the service to have no endpoints).  When the break is reverted the migration should continue without issue.

Also, check that an unpatched database has the same schema as a patched one (i.e. that I didn't incorrectly handle any migration).

Additionally, :cat: should pass (it did locally).

## Additional information

The script to generate the patch is attached as [migration-squasher.patch.gz](https://github.com/SUSE/scf/files/3288327/migration-squasher.patch.gz) (it's a patch to `cloud_controller_ng` due to requiring all of its magic to run the migration scripts).

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,6 +65,7 @@ Table of Contents
       * [How do I use an authenticated registry for my Docker images?](#how-do-i-use-an-authenticated-registry-for-my-docker-images)
       * [Using Persi NFS](#using-persi-nfs)
       * [How do I rotate the CCDB secrets?](#how-do-i-rotate-the-ccdb-secrets)
+      * [CCDB Migration Squashing](#ccdb-migration-squashing)
 
 # Deploying SCF on Vagrant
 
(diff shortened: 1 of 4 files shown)


mook-as:
Script attached and README updated in db37f6be363e6859eed00e6ddde2f17ea7e22bfc. I opted to not put it in `tooling/bin/` as you can't directly run it.

README.md
```diff
@@ -884,3 +889,28 @@ equivalent) can be run again with the same parameters. Some commands need to be
...
+when doing the initial database migration from nothing.  Given that we only
+support upgrades from the previous verison, we have implemented a [patch to
+squash all initial database migrations].  To update this patch:
+
+1. Remove the patch (so that we can correctly generate a new one).
+2. Deploy SCF until it is ready.
+3. Apply [`0001-db-migration-add-script-to-squash-DB-migrations.patch`]
+   to the `api-group` container
+4. Follow the instructions at the top of the patch file (updating timestamps)
+5. Run `rake db:squash` (as noted in the patch file) to generated the new
```
andreas-kupries: Nit: `generated` -> `generate`

tooling/helpers/0001-db-migration-add-script-to-squash-DB-migrations.patch
```diff
@@ -0,0 +1,1237 @@
...
+squashing, so that we can still upgrade from existing installations that
+may have some of the migrations applied.
+
+It appears that there are still some differences with a database
+generated via the squashed migration, compared to a normal one:
+
+- Missing / added charset=utf8, collate=utf8_bin
+  (This shouldn't be an issue, because of table defaults)
+- Some indexes &c. are in a different column order.
+  This is not expected to cause any issues.
```
andreas-kupries: I also see changes to the initial auto-increment settings, i.e `CREATE TABLE ( ... ) AUTO_INCREMENT=N`, where N changed. Thought I should note that, despite not expecting these to cause issues either.

PR #2422, opened 2019-05-23
Remove all references to cflinuxfs2

jimmykarily:
## Description
cflinuxfs2 is no longer supported upstream and is removed. This PR removes it from scf.

## Test plan

Wait for CI to build this PR. Deploy it on some k8s cluster and run `cf stacks`. There should be no cflinuxfs2 listed.


jimmykarily:
1 test in CATs is failing because it explicitly tries to use cflinuxfs2. It should be fixed with the next bump for CATs.

jandubois:
Included in https://github.com/SUSE/scf/pull/2610

PR #2438, opened 2019-05-29
Simplify service accounts and pod security policies

jandubois:
Includes bump to fissile version that implements the simplified model.

Jobs no longer specify a psp; there is just one psp anymore and it is bound to all service accounts.  In addition to capabilities each job now also specifies if the pod needs to be privileged.

Most of the cluster roles have been replaced by regular roles. The exceptions are the `eirini`, `node-reader`, and `test` cluster roles.

Eirini continues to run effectively as cluster-admin and will be addressed at a later time. The node-reader cluster role is required for diego-cells to be able to read the AZ label of the nodes they
are running on. The test-role is used by brain tests and not installed by the helm chart.

See also: https://github.com/cloudfoundry-incubator/fissile/wiki/ServiceAccount-and-PodSecurityPolicy-usage

### DIFF
--- a/bin/common/versions.sh
+++ b/bin/common/versions.sh
@@ -10,7 +10,7 @@ set -o errexit -o nounset
 export BOSH_CLI_VERSION="fcaa9c6caff58ab8da8c56481320681cdea492ee"
 export CFCLI_VERSION="6.42.0"
 export FISSILE_FLAVOR="develop"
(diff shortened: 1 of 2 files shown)


f0rmiga:
Waiting for UAA and Fissile to be updated with the merged ones.

PR #2439, opened 2019-05-30
CAP-106 Don't use eirini service account for diego post-deployment-setup

jandubois:
Requires a bump to fissile to pass through a templated service account name without trying to validate it at build time.


jandubois:
https://github.com/SUSE/scf/pull/2470 splits of Eirini setup into a separate job with it's own service account.

PR #2442, opened 2019-05-31
[#UcByaqlM] [CAP-259] Force abort on errors within AZ and PZ handling.

andreas-kupries:
## Description

See [CAP-259](https://jira.suse.com/browse/CAP-259)

## Notes

  - We only have to remove the `|| true` masking the error. The script runs in a context which aborts on error.
  - Rebased to head of dev. Collapsed the various back and forth in the history.

## Test plan

In the original submission it was `missing privileges of node-reader service account` which should have triggered the error.

Tested in local vagrant with
- Intentionally crippled `node-reader` account. (See the `verbs` of `node-reader-role` to `delete` (Cannot make it empty, that is rejected by `fissile`. Using `delete`, an irrelevant verb here, to get around that))
- Inserted `false`.

### DIFF
--- a/container-host-files/etc/scf/config/scripts/configure-az.sh
+++ b/container-host-files/etc/scf/config/scripts/configure-az.sh
@@ -23,7 +23,7 @@ else
     # the diego-cell instance group.
 
     QUERY="jsonpath={.metadata.labels.${AZ_LABEL_NAME}}"
-    NODE_AZ=$("${kubectl}" get node "${node}" -o "${QUERY}" || true)
+    NODE_AZ=$("${kubectl}" get node "${node}" -o "${QUERY}")
 
     if test -z "${NODE_AZ}"
     then
--- a/container-host-files/etc/scf/config/scripts/configure-pz.sh
+++ b/container-host-files/etc/scf/config/scripts/configure-pz.sh
@@ -23,7 +23,7 @@ else
     # the diego-cell instance group.
 
     QUERY="jsonpath={.metadata.labels.${PZ_LABEL_NAME}}"
-    NODE_PZ=$("${kubectl}" get node "${node}" -o "${QUERY}" || true)
+    NODE_PZ=$("${kubectl}" get node "${node}" -o "${QUERY}")
 
     if test -z "${NODE_PZ}"
     then


container-host-files/etc/scf/config/scripts/configure-az.sh
```diff
@@ -1,5 +1,8 @@
 #!/bin/sh
 
+set -o nounset ;# trap undefined variables
+set -o errexit ;# abort on error
+
 if test -z "${AZ_LABEL_NAME}"
```
mook-as: Wouldn't this be an issue now that we have `nounset`? Or do we always provide an `AZ_LABEL_NAME`, that just happens to be empty sometimes?

container-host-files/etc/scf/config/scripts/configure-az.sh
```diff
@@ -36,3 +39,7 @@ else
         export KUBE_AZ="${NODE_AZ}"
     fi
 fi
+
+# Restore outer settings ...
+set +o nounset ;# allow undefined variables
+set +o errexit ;# stop aborting on error
```
mook-as: That means we can never set them in the outer script? (Otherwise, this will now unset them magically, which is probably worse)

container-host-files/etc/scf/config/scripts/configure-az.sh
andreas-kupries: No, the variable is always set, just empty sometimes.

container-host-files/etc/scf/config/scripts/configure-az.sh
andreas-kupries: Do you request the removal of the `nounset` ?

andreas-kupries:
> Given then || true I'm not sure how this change actually affects anything?
> Was this tested with the node-reader privileges intentionally crippled (so that it failed to read the node info)?

It was tested with an intentionally unset variable.
I completely missed the `|| true`. 
That has to go way, certainly.

container-host-files/etc/scf/config/scripts/configure-az.sh
andreas-kupries: Removed `nounset` with commit 4b739f7c4af0a82bd6ee3bd6348edb058a7a0849

container-host-files/etc/scf/config/scripts/configure-az.sh
andreas-kupries: Removed the `|| true` with commit 4b739f7c4af0a82bd6ee3bd6348edb058a7a0849

andreas-kupries:
The testplan in the description says:

> In my testing in a vagrant-deployed cluster I directly inserted a false, or echo $BOGUS_VARIABLE into the configure-az.sh script to force an error.

And in answer to the first iteration of the question I said 

> It was tested with an intentionally unset variable.

Thus, no, it was not tested with an intentionally crippled node reader role.

How can I cripple the/a role ?

PR #2446, opened 2019-06-03
Bump autoscaler to 1.2.1

bikramnehra:
## Description

Ref: https://github.com/SUSE/scf/pull/2364

**JIRA:** https://jira.suse.com/browse/CAP-251

## Test plan

1. Make sure the build is passing
2. Test upgrades from CAP 1.4
3. Must be tested on CaaSP as per IBM

### DIFF
--- a/container-host-files/etc/scf/config/opinions.yml
+++ b/container-host-files/etc/scf/config/opinions.yml
@@ -48,100 +48,97 @@ properties:
     cf:
       client_id: app_autoscaler
       grant_type: client_credentials
-      password: "" # unused
-      username: "" # unused
+      skip_ssl_validation: true
     api_server:
+      service_offering_enabled: false
+      require_consul: false
+      port: 7100
+      publicPort: 7106
+      health:
+        port: 9101
       eventgenerator:
         port: 7105
       metrics_collector:
         port: 7103
-      port: 7100
-      publicPort: 7106
-      require_consul: false
-      service_broker:
-        port: 7107
       scheduler:
         port: 7102
       scaling_engine:
         port: 7104
+      db_config:
+        max_connections: 5
+        min_connections: 1
+        idle_timeout: 60000
     appmetrics_db:
       databases:
         - name: "autoscaler"
           tag: "default"
-      port: 5432
-      db_scheme: postgres
-    binding_db:
-      databases:
-        - name: "autoscaler"
-          tag: "default"
-      port: 5432
-      db_scheme: postgres
+      db_scheme: postgres          
     eventgenerator:
-      metricscollector:
-        port: 7103
       require_consul: false
-      scaling_engine:
-        port: 7104
       server:
         port: 7105
+      health:
+        port: 9105   
+      metricscollector:
+        host: localhost
+        port: 7103
+      scaling_engine:
+        port: 7104
     instancemetrics_db:
       databases:
         - name: "autoscaler"
           tag: "default"
-      port: 5432
-      db_scheme: postgres
+      db_scheme: postgres          
     lock_db:
       databases:
         - name: "autoscaler"
           tag: "default"
-      port: 5432
-      db_scheme: postgres
+      db_scheme: postgres          
     metricscollector:
+      require_consul: false
       server:
         port: 7103
-      require_consul: false
+      health:
+        port: 9103        
     policy_db:
       databases:
         - name: "autoscaler"
           tag: "default"
-      port: 5432
-      db_scheme: postgres
+      db_scheme: postgres          
     operator:
       require_consul: false
+      health:
+        port: 9108   
       scaling_engine:
+        host: localhost
         port: 7104
       scheduler:
+        host: localhost      
         port: 7102
     scalingengine:
-      health:
-        port: 8083
       require_consul: false
       server:
         port: 7104
+      health:
+        port: 9104
     scalingengine_db:
       databases:
         - name: "autoscaler"
           tag: "default"
-      port: 5432
-      db_scheme: postgres
+      db_scheme: postgres          
     scheduler:
-      port: 7102
       require_consul: false
+      port: 7102
+      health:
+        port: 9102      
       scaling_engine:
+        host: localhost
         port: 7104
     scheduler_db:
       databases:
         - name: "autoscaler"
           tag: "default"
-      port: 5432
-      db_scheme: postgres
-    service_broker:
-      username: username
-      api_server:
-        port: 7100
-      port: 7107
-      publicPort: 7101
-      require_consul: false
+      db_scheme: postgres          
   bits-service:
     active_signing_key:
       key_id: key1
@@ -452,7 +449,7 @@ properties:
       # go into the SCF zone.
       app_autoscaler:
         # Requires read-only admin to look up stats, and write admin to do scaling
-        authorities: cloud_controller.read,cloud_controller.write,cloud_controller.admin
+        authorities: cloud_controller.read,cloud_controller.write,cloud_controller.admin,uaa.resource
         authorized-grant-types: client_credentials
       cc-service-dashboards:
         scope: cloud_controller_service_permissions.read,openid
--- a/container-host-files/etc/scf/config/role-manifest.yml
+++ b/container-host-files/etc/scf/config/role-manifest.yml
@@ -132,9 +132,9 @@ releases:
   version: "4.46"
   sha1: "dc3d8dd8c7bba73b54b67054a478ee93caeb9968"
 - name: app-autoscaler
-  version: "1.0.0"
-  url: "https://bosh.io/d/github.com/cloudfoundry-incubator/app-autoscaler-release?v=1.0.0"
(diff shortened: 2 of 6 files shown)


container-host-files/etc/scf/config/role-manifest.yml
```diff
@@ -2154,45 +2171,62 @@ instance_groups:
...
       bosh_containerization:
+        colocated_containers:
+        - loggregator-agent
         run:
           scaling:
             min: 1
             max: 3
             ha: 2
-          memory: 128
+          memory: 1024
```
bikramnehra: What is the reason behind such a significant change in memory requirements?

container-host-files/etc/scf/config/role-manifest.yml
```diff
@@ -3128,78 +3033,13 @@ variables:
...
-    description: A PEM-encoded TLS key for clients to connect to the Autoscaler API.
-      This includes the Autoscaler ApiServer and the Service Broker.
-    required: true
 - name: AUTOSCALER_ASAPI_PUBLIC_SERVER_CERT
   options:
     secret: true
     ca: INTERNAL_CA_CERT
     alternative_names:
     - autoscaler.{{.DOMAIN}}
+    - localhost
```
bikramnehra: Do we need to have `localhost` in SAN?

src/scf-release/src/acceptance-tests-brain/test-resources/policy.json
```diff
@@ -1,32 +1,13 @@
...
+  "scaling_rules": [{
+      "metric_type": "memoryused",
+      "stat_window_secs": 60,
+      "breach_duration_secs": 60,
+      "threshold": 10,
+      "operator": ">=",
+      "cool_down_secs": 300,
+      "adjustment": "+1"
+  }]
+}
```
bikramnehra: newline?

container-host-files/etc/scf/config/opinions.yml
```diff
@@ -48,100 +48,97 @@ properties:
     cf:
       client_id: app_autoscaler
       grant_type: client_credentials
-      password: "" # unused
-      username: "" # unused
+      skip_ssl_validation: true
```
jandubois: Why is this necessary?

container-host-files/etc/scf/config/opinions.yml
```diff
@@ -452,7 +449,7 @@ properties:
       # go into the SCF zone.
       app_autoscaler:
         # Requires read-only admin to look up stats, and write admin to do scaling
-        authorities: cloud_controller.read,cloud_controller.write,cloud_controller.admin
+        authorities: cloud_controller.read,cloud_controller.write,cloud_controller.admin,uaa.resource
```
jandubois: I don't know what the `uaa.resource` scope does; why does the autoscaler need it?

container-host-files/etc/scf/config/role-manifest.yml
```diff
@@ -2154,45 +2171,62 @@ instance_groups:
...
-          scaling:
-            min: 1
-            max: 3
-            ha: 2
-          memory: 2350
-          virtual-cpus: 2
+      bosh_containerization:    
         ports:
-        - name: scaling-engine  # scalingengineport
+        - name: eventgen  # eventgenereatorport
```
jandubois: Spelling: `eventgeneratorport`

container-host-files/etc/scf/config/role-manifest.yml
```diff
@@ -2154,45 +2171,62 @@ instance_groups:
...
+        - name: eghealthport
+          protocol: TCP
+          external: 9105
+          internal: 9105
+  - name: bpm
+    release: bpm
+    properties:
+      bosh_containerization:
+        run:
+          capabilities: [ALL]
```
jandubois: No need to specify `capabilities` for a `privileged` container.

container-host-files/etc/scf/config/role-manifest.yml
```diff
@@ -2206,110 +2240,71 @@ instance_groups:
...
-            max: 3
-            ha: 2
-          memory: 2350
-          virtual-cpus: 2
-        ports:
-        - name: eventgen  # eventgenereatorport
-          protocol: TCP
-          external: 7105
-          internal: 7105
+          capabilities: [ALL]
```
jandubois: No need to specify `capabilities` for a `privileged` container.

container-host-files/etc/scf/config/role-manifest.yml
```diff
@@ -3258,218 +3099,29 @@ variables:
...
     required: true
-- name: AUTOSCALER_DATABASE_MAX_CONNECTIONS
+- name: AUTOSCALER_DB_ADDRESS
   options:
-    default: 1000
-    description: The maximum number of connections the Autoscaler postgres database
-      server could serve.
+    default: "autoscaler-postgres-postgres.((KUBERNETES_NAMESPACE)).svc.((KUBERNETES_CLUSTER_DOMAIN))"
+    description: The password for the Autoscaler postgres database.
+    required: true
```
jandubois: The `descriptions` sounds wrong.

And `required: true` is meaningless when you have a `default` value.

container-host-files/etc/scf/config/role-manifest.yml
jandubois: I'm a little surprised how many certs need a `localhost` alt name. Not really a problem, but I would like to understand why this was done.

container-host-files/etc/scf/config/opinions.yml
cdlliuy: autoscaler will connect to uaa api for authorization and cf api to get app info and trigger scale.  When running with the self-signed certs ,   we need this flag to ensure the correctly connection .

container-host-files/etc/scf/config/opinions.yml
cdlliuy: this is due to a functional change in app-autoscaler in which we need to allow cf admin's access to autoscaler resourc.  To validate admin role,  the  uaa.resource authority is required to invoke /token endpoint.

container-host-files/etc/scf/config/opinions.yml
jandubois: Maybe I'm confused, but doesn't this just mean that you are missing to authorize the internal CA in some of your roles?

```
  - name: authorize-internal-ca # needs to be second as it is a dependency of other jobs.
    release: scf
```

Assuming you include this job in all your roles, all self-signed certs should verify correctly.

container-host-files/etc/scf/config/role-manifest.yml
cdlliuy: In the recent app-autoscaler releases,  we enabled local cache to reduce the frequency of database access for metric data, so more memory is required here.

container-host-files/etc/scf/config/role-manifest.yml
cdlliuy: To colocated different jobs inside 1 container and avoid the hairpin issue, 
 the "localhost" access is hardcoded, such as , https://github.com/SUSE/scf/blob/507d14f3cb354d35de47fb673e43a8e160268717/container-host-files/etc/scf/config/opinions.yml#L135
Without "localhost" in SAN, the tls connection will be rejected with dns mismatch.

cdlliuy:
I pushed an update in https://github.com/cdlliuy/scf/commit/fed4a57329c8dee558cc186435533c177e117759 since I can't push directly in this repo. 

But given the test on my vagrant box is too slow, I didn't finish the "vagrant-preq" step yet.  That means the changes are not validated ..

container-host-files/etc/scf/config/role-manifest.yml
bikramnehra: already answered in previous comment so resolving this one

bikramnehra:
@cdlliuy Unfortunately again started seeing issues in `autoscaler-brains` after pulling your latest changes. I didn't had a chance to go though the failures yet. However, here is detailed log: [autoscaler-logs.txt](https://github.com/SUSE/scf/files/3259157/autoscaler-logs.txt)

container-host-files/etc/scf/config/opinions.yml
cdlliuy: @jandubois ,  I need to keep 'skip_ssl_validation' here since the access target is the CC public API,  not for internal CA. Just like the "cf cli" and "smoke" test need to use "skip_ssl_valication" flag ..   If not,   autoscaler will report issue 
```
{"timestamp":"2019-06-06T05:01:53.211Z","source":"autoscaler:apiserver","text":"Failed to get Cloud Foundry API information","log_level":"error","data":{"info url":"https://api.cf.svc.cluster.local:9024/v2/info","error":{"code":"UNABLE_TO_VERIFY_LEAF_SIGNATURE","statusCode":500}}}
```

cdlliuy:
@bikramnehra , I pushed a new commit https://github.com/cdlliuy/scf/commit/a83830bcbc59eca167b4f47db7743268336b9be4  to add "skip_ssl_validation" back to pass the test.   Please help to retry.

PR #2456, opened 2019-06-05
CF Bump 7.11

bikramnehra:
## Description

Submodule changes:

```
bpm, 1.0.3 -> 1.04
cf-smoke-tests, 40.0.50 -> 40.0.51
cf-syslog-drain, 9.1 -> 10.0
credhub, 2.1.4 -> 2.2.0
loggregator, 105.0 -> 105.2
routing, 0.184.0 -> 0.187.0
statsd-injector, 1.8.0 -> 1.9.0
loggregator-agent, 3.7 -> 3.9
```
**Notes:**
- Made changes for enabling TLS between trafficcontroller and gorouter. Refer [this](https://github.com/cloudfoundry/loggregator-release/releases/tag/v105.1).
- Switched to using `log-cache` instead of `traffic_controller` for retrieving container metrics. Refer [this](https://github.com/cloudfoundry/loggregator-release/releases/tag/v105.0).

## Test plan

Make sure the build is passing.

### DIFF
--- a/bin/common/versions.sh
+++ b/bin/common/versions.sh
@@ -36,7 +36,7 @@ export GOLANG_VERSION=1.7
 # Used in: make/include/versioning
 
 export PRODUCT_VERSION="1.4"
-export CF_VERSION="7.9.0"
+export CF_VERSION="7.11.0"
 
 # Show versions, if called on its own.
 # # ## ### ##### ######## ############# #####################
--- a/container-host-files/etc/scf/config/role-manifest.yml
+++ b/container-host-files/etc/scf/config/role-manifest.yml
@@ -1469,7 +1469,7 @@ instance_groups:
           healthcheck:
             readiness:
               command:
-              - curl --silent --fail --head http://${HOSTNAME}:8081/set-cookie
+              - curl --silent --fail --head --insecure https://${HOSTNAME}:8081/set-cookie
           affinity:
             podAntiAffinity:
               preferredDuringSchedulingIgnoredDuringExecution:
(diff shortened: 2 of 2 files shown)


bikramnehra:
> Huh, there were no existing users of `log-api.((DOMAIN))` and `loggregator_trafficcontroller.((DOMAIN))`? Then why did we expose them in the first place?
> 
> Given that I find no other references, do we need `doppler.((DOMAIN))`?

I guess we were never having `log-api.((DOMAIN))` anywhere. It was just named `log-api` though was pointing to doppler. Yes we need `doppler.((DOMAIN))` and this is what gets used to route traffic to traffic_controller(though its weird why upstream chose the name `doppler`).

For `loggregator_trafficcontroller` I was not able to find any references.  It would have been useless anyways.

mook-as:
I know that upstream has a `doppler.((DOMAIN))`; I'm asking where we actually _use that value_, given that we didn't have that before and things worked. (Internally, we are using various `*.cluster.local` things, and do not go through the router.)

bikramnehra:
> I know that upstream has a `doppler.((DOMAIN))`; I'm asking where we actually _use that value_, given that we didn't have that before and things worked. (Internally, we are using various `*.cluster.local` things, and do not go through the router.)

As I mentioned, the value is used to connect to `trafficcontroller` server. You'll see this issue if the value isn't specified:

https://github.com/cloudfoundry/cf-deployment/issues/400

Also, here's the api endpoint test making use of it:

https://github.com/cloudfoundry/cli/blob/master/cf/api/endpoints_test.go#L41

We were always having the `doppler.((DOMAIN))` route defined under the name `log-api`.

mook-as:
So to answer my own question, it's hard coded in [cloud controller configs](https://github.com/cloudfoundry/capi-release/blob/1.79.0/jobs/cloud_controller_ng/templates/cloud_controller_ng.yml.erb#L155).  Which means we might need to look into [`log-stream`](https://github.com/cloudfoundry/capi-release/blob/1.79.0/jobs/cloud_controller_ng/templates/cloud_controller_ng.yml.erb#L152) as well.

bikramnehra:
`log-stream` endpoint is for exposing `reverse-log-proxy`(used mainly by `syslog drains`):

https://github.com/cloudfoundry/cf-deployment/blob/v7.11.0/cf-deployment.yml#L1755-L1761

I am not quite certain if its being used anywhere at the moment and since never defined it in our `role-manifest` so I kept it like that.

PR #2457, opened 2019-06-05
CAP-99 Provide a mechanism to set -flakeAttempts when running CATS

jandubois:
CATS often has random failures on Jenkins that are due to resource pressure and not reproducible. -flakeAttempts tells Gingko to retry failed tests multiple times before treating them as failures. Each
individual failed test will still be reported, just not reflected in the exit code.

[CAP-99]

### DIFF
--- a/container-host-files/etc/scf/config/role-manifest.yml
+++ b/container-host-files/etc/scf/config/role-manifest.yml
@@ -1886,6 +1886,7 @@ instance_groups:
           flight-stage: manual
   configuration:
     templates:
+      properties.acceptance_tests.flake_attempts: '"((ACCEPTANCE_TEST_FLAKE_ATTEMPTS))"'
       properties.acceptance_tests.nodes: '"((ACCEPTANCE_TEST_NODES))"'
       scf.cats-suites: '"((CATS_SUITES))"'
 - name: sync-integration-tests
@@ -3066,6 +3067,12 @@ configuration:
     # Really a loggregator_agent job property
     properties.zone: '"((KUBE_AZ))"'
 variables:
+- name: ACCEPTANCE_TEST_FLAKE_ATTEMPTS
+  options:
+    default: 3
+    description: >
+      The number of times Ginkgo will run a test before treating it as a failure.
+      Individual failed runs will still be reported in the test output.
 - name: ACCEPTANCE_TEST_NODES
   options:
     default: 4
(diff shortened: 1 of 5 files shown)


container-host-files/etc/scf/config/role-manifest.yml
```diff
@@ -3066,6 +3067,12 @@ configuration:
     # Really a loggregator_agent job property
     properties.zone: '"((KUBE_AZ))"'
 variables:
+- name: ACCEPTANCE_TEST_FLAKE_ATTEMPTS
+  options:
+    default: 3
+    description: >
+      The number of times Ginkgo will run a test before treating it as a failure.
+      Individual failed runs will still be reported in the test output.
```
mook-as: I know it's implied in the name, but we should mention in the description that this is for CATS (unless you want this to be for other tests too, in which case say that).

src/scf-release/jobs/acceptance-tests/templates/run.erb
```diff
@@ -48,7 +49,7 @@ fi
 
 set +e
 bin/test -r $noColorFlag -slowSpecThreshold=120 -randomizeAllSpecs $verbose \
-  -nodes=$nodes -skipPackage=helpers $skips -keepGoing "${focus[@]}" \
+  -flakeAttempts=$flakeAttempts -nodes=$nodes -skipPackage=helpers $skips -keepGoing "${focus[@]}" \
```
mook-as: I'd expect shellcheck to complain, but it's not like the rest of the line wouldn't trigger the same…

container-host-files/etc/scf/config/role-manifest.yml
jandubois: I changes it to `...run a CATS test...`

I also added support for `-flakeAttempts` to SITS, but it is using a separate config variable.

src/scf-release/jobs/acceptance-tests/templates/run.erb
jandubois: Well, running `shellcheck` against `run.erb` complains about all manner of thing because the file is not a valid shell script.

But once the template is expanded, `shellcheck` will not  complain about these variables because it can now tell that there aren't any spaces in these variable values, so they don't have to be put in quotes.

It also doesn't complain about missing braces because there is no ambiguity in these cases.

I've still fixed the other `shellcheck` errors and added braces everywhere, just for consistency.

PR #2475, opened 2019-06-12
[DUPLICATE] [WIP] Remove cflinuxfs2

bikramnehra:
## Description

TODO

## Test plan

TODO


gaktive:
Until we get to cf-deployment 8.x, let's not merge this in since we may have to release another scf with cf 7.11 and we don't need to remove cflinuxfs2 until then.

jandubois:
This is included in https://github.com/SUSE/scf/pull/2610

PR #2478, opened 2019-06-13
Jenkins: publishing artifacts before deploy stage

bikramnehra:
## Description

This PR addresses: https://jira.suse.com/browse/CAP-397 moving the pubish to docker and s3 before deploy stage.

Also, fixed a minor issue with s3 log dump path.

## Test plan

Make sure if the docker images and artifacts were published before deploy stage.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -485,6 +485,76 @@ pipeline {
             }
         }
 
+        stage('publish_docker') {
+            when {
+                expression { return params.PUBLISH_DOCKER }
+            }
+            steps {
+                withCredentials([usernamePassword(
+                    credentialsId: params.DOCKER_CREDENTIALS,
+                    usernameVariable: 'DOCKER_HUB_USERNAME',
+                    passwordVariable: 'DOCKER_HUB_PASSWORD',
+                )]) {
+                    sh 'docker login -u "${DOCKER_HUB_USERNAME}" -p "${DOCKER_HUB_PASSWORD}" "${FISSILE_DOCKER_REGISTRY}" '
+                }
+                sh '''
+                    set -e +x
+                    source ${PWD}/.envrc
+                    set -x
+                    unset SCF_PACKAGE_COMPILATION_CACHE
+                    make publish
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -872,9 +872,9 @@ pass = ${OBS_CREDENTIALS_PASSWORD}
                                 def prefix = distPrefix()
                                 // If prefix is not empty then trimming trailing "-" from path as distPrefix() returns "PR-${CHANGE_ID}-".
                                 if (prefix){
-                                    prefix = prefix.substring(0, prefix.length() - 1)
+                                    prefix = prefix.substring(0, prefix.length() - 1) + '/'
                                 }
```
satadruroy: Is there a stylistic preference for this cosmetic change? If you want to be OS-agnostic you can use `File.separator` (works in Java, should work in Groovy as well) but given we know what these slave nodes are running it's bit of a moot point...

bikramnehra: This fix is to address https://jira.suse.com/browse/CAP-405.

I guess there are lot of places where `/` is hard-coded, so unless we are planning to remove all references I would keep it like that.

jandubois: I don't think there are even any OSes around anymore that don't support `/` as the path separator. So I don't think it makes any sense to try to generalize things. Next thing you have to worry about is drive letters and VMS volume prefixes...

But in this case the string is to become part of a URL, in which case `/` is the only valid path separator; the local OS doesn't even come into play. So using a literal `/` is arguably *more* correct. 😄

satadruroy: My point was why make this change at all - something I still don't understand. Anyway, it's a minor nit...

bikramnehra: @satadruroy As I mentioned, this change has to do with resolving a bug in s3 log uploads, e.g: 

```
s3://cap-jenkins-logs/nightly//jenkins-scf-develop-1307/
```

Notice `//` in s3 path, we need to append it only if `prefix` is non NULL.

PR #2482, opened 2019-06-14
Dummy PR to test deploy speed

mook-as:
## Description

We seem to be getting slower at deploying SCF.  This is a dummy PR from a commit in April (just after 2.16.2) to check if it's faster with old code, or if it's something to do with the infrastructure.

## Test plan

This will run on Jenkins.

This should not be merged.


mook-as:
No longer relevant.

PR #2493, opened 2019-06-19
Updated product version in versions.sh

satadruroy:
Bump product version to 1.4.1


satadruroy:
@gaktive has alrrerady a PR open.

PR #2503, opened 2019-06-25
Exclude kube-system namespace from pod-status output

jandubois:
On a multi-node cluster with ingress, minibroker, scf, stratos, and uaa deployed the list of pods already fills the screen. Including kube-system pods pushes some entries below the bottom, making them invisible to `pod-status -w`.

### DIFF
--- a/container-host-files/opt/scf/bin/pod-status
+++ b/container-host-files/opt/scf/bin/pod-status
@@ -3,7 +3,9 @@
 set -o errexit -o nounset
 
 NS=""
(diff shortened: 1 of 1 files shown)


jandubois:
> This makes it impossible to ever look at `kube-system` pods

"_Impossible ... ever_" is quite a strong statement; nothing prevents you from using

```
kubectl get pods --namespace kube-system
```

I never wanted to see the `kube-system` pods via `pod-status`, but I guess I can add a `-s` option or something to include them...

PR #2512, opened 2019-07-01
Revert eirini bump

jimmykarily:
## Description

We need to get back to the working eirini bosh release (0.0.4) until we get the latest to work.


jandubois:
I've rebased this PR on top of latest `develop`, which includes the rollback of the ruby-buildpack bump, which has been breaking tests.

PR #2520, opened 2019-07-04
Get rid of cluster-admin role binding for eirini service account

jandubois:
Running pods with `cluster-admin` privileges seems like a very bad idea. Even having service accounts with `cluster-admin` access is dangerous when you have pods that have the right to create or modify pods, deployments, stateful sets etc. because they can then make use of the service account to escalate privileges.

`cluster-admin` was needed for 2 purposes:

* granting access to the eirini namespace to schedule applications there
* allowing the post-deployment setup to create daemonsets for fluentd, and running the cert-copier

Access to the eirini namespace can be provided by a role inside the eirini namespace that is then bound to the service account in the scf namespace.

The daemonset shouldn't be created inside the post-deployment job in the first place; it should be part of the helm chart.

One remaining problem with this PR is that the daemonset uses the `loggregator/agent:latest` image, which doesn't use the Configgin mechanism to configure bosh jobs. Therefore we don't have access to `KUBERNETES_CLUSTER_DOMAIN` and have to hard-code it to the `cluster.local` default value.

Another reason for moving the daemonset to the helm chart was to make it easier to modify it at deployment time to debug problems on different kube installations.

Finally, creating the eirini namespace also makes sure it is being cleaned up when the release is deleted from the cluster.

### DIFF
--- a/chart-parts/eirini-daemonset.yaml
+++ b/chart-parts/eirini-daemonset.yaml
@@ -1,4 +1,5 @@
 ---
+{{- if .Values.enable.eirini }}
 apiVersion: extensions/v1beta1
 kind: DaemonSet
 metadata:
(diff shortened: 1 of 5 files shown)


jandubois:
I've seen one deployment failure inside the vagrant box when running `make run-eirini`:

```
Error: release cf failed: namespaces "eirini" not found
```

I could not determine why this happened, and could not reproduce on re-run, so maybe it was some pre-existing state inside the box. But I wanted to keep a note here, just in case a reviewer hits the same issue.

jandubois:
Replaced by #2649

PR #2530, opened 2019-07-08
Don't hardcode storage class and add some convenience targets

viovanov:


bazel-scf-v2
```diff
@@ -0,0 +1 @@
+/home/toaster/.cache/bazel/_bazel_toaster/0291e00957d43be0bd77a17ccdc76a11/execroot/scf
```
f0rmiga: This symlink slipped out the generated Bazel directories. We need to update the `.gitignore` to include `/bazel-*`.

deploy/helm/scf/assets/operations/instance_groups/database.yaml
```diff
@@ -122,6 +121,9 @@
       processes:
       - name: mariadb_ctrl
         executable: /bin/bash
+        limits:
```
f0rmiga: Any particular reason to move these? I'd prefer having the `args` right after the `executable` as they together make the full command.

deploy/helm/scf/values.yaml
```diff
@@ -1 +1,2 @@
 system_domain: ~
+asd: asd
```
f0rmiga: Left-over?

deploy/helm/scf/assets/operations/instance_groups/database.yaml
viovanov: it's hard for me to read where the keys belong if they're after the long script
can't enforce something like this .. but I'll put them before `executable`

deploy/helm/scf/values.yaml
viovanov: :D

deploy/helm/scf/assets/operations/instance_groups/database.yaml
f0rmiga: I see. This works too.

PR #2535, opened 2019-07-10
cf login instead of cf api

giner:
Get straight to the point using a single command.

This is a cosmetic change to NOTES.txt

### DIFF
--- a/chart-parts/NOTES.txt
+++ b/chart-parts/NOTES.txt
@@ -3,8 +3,8 @@
     The endpoint for use by the `cf` client is
         [1;36mhttps://api.{{ .Values.env.DOMAIN }}[0m
 
-    To target this endpoint run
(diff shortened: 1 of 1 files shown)


chart-parts/NOTES.txt
```diff
@@ -3,8 +3,8 @@
     The endpoint for use by the `cf` client is
         [1;36mhttps://api.{{ .Values.env.DOMAIN }}[0m
 
-    To target this endpoint run
-        [1;36mcf api --skip-ssl-validation https://api.{{ .Values.env.DOMAIN }}[0m
+    To target this endpoint run and login
```
jandubois: Should be:

`To target this endpoint and login, run`

giner: oops

PR #2542, opened 2019-07-10
Add api instance group with cloud_controller_ng

f0rmiga:
## Description

Bumped the default stemcell version and added the api instance group with the cloud_controller_ng job and the binary buildpack.

## Test plan

Deploy the cf-operator and on SCF:

```
bazel build //deploy/helm/scf:chart && helm template --name scf bazel-bin/deploy/helm/scf/scf-3.0.0.tgz --namespace scf --set "system_domain=$(minikube ip).xip.io" | kubectl apply -f - --namespace scf
```

### DIFF
--- a/deploy/helm/scf/assets/operations/instance_groups/api.yaml
+++ b/deploy/helm/scf/assets/operations/instance_groups/api.yaml
@@ -1,0 +1,95 @@
+# Selectively remove jobs temporarily.
+- type: remove
+  path: /instance_groups/name=api/jobs/name=route_registrar
+- type: remove
+  path: /instance_groups/name=api/jobs/name=statsd_injector
+- type: remove
+  path: /instance_groups/name=api/jobs/name=file_server
+- type: remove
+  path: /instance_groups/name=api/jobs/name=routing-api
+- type: remove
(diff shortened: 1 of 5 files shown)


deploy/helm/scf/assets/operations/instance_groups/api.yaml
```diff
@@ -0,0 +1,95 @@
+# Selectively remove jobs temporarily.
```
andreas-kupries: So the instance group is added, but without any jobs actually running ?

andreas-kupries: Never mind, should have read the ticket. This is about adding the CC job in the api group.

f0rmiga: The title says "cloud_controller_ng".

andreas-kupries: Huh ... Oh, PR title, right. I was looking at JIRA ticket title, and then description.

f0rmiga: I'll update JIRA!

PR #2550, opened 2019-07-11
Moved packaging logic to a standalone script

f0rmiga:
## Description

Moved packaging logic to a standalone script to make it "lintable" and easier to maintain.

## Test plan

`bazel build //deploy/helm/scf:chart` should produce the chart package under `bazel-bin/deploy/helm/scf/scf-3.0.0.tgz`.

### DIFF
--- a/rules/helm/def.bzl
+++ b/rules/helm/def.bzl
@@ -10,6 +10,7 @@ def _package_impl(ctx):
         executable = ctx.executable._script,
         env = {
             "PACKAGE_DIR": ctx.attr.package_dir,
+            # TODO(f0rmiga): Figure out a way of working with paths that contain spaces.
             "TARS": " ".join([f.path for f in ctx.files.tars]),
             "HELM": ctx.executable._helm.path,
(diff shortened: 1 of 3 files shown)


rules/helm/def.bzl
```diff
@@ -2,38 +2,21 @@ def _package_impl(ctx):
...
-            package_dir = ctx.attr.package_dir,
-            chart_version = ctx.attr.chart_version,
-            app_version = ctx.attr.app_version,
-            output_filename = output_filename,
-            output_tgz = output_tgz.path,
-        ),
+        executable = ctx.executable._script,
+        env = {
+            "PACKAGE_DIR": ctx.attr.package_dir,
+            "TARS": " ".join([f.path for f in ctx.files.tars]),
```
mook-as: Optionally, we can pass this in as `arguments` (as a parameter to `ctx.actions.run`), in case we ever end up with spaces in the list.

f0rmiga: Good catch. I'll leave a TODO, as it may overlap with the work on generating the patch ops-files.

PR #2553, opened 2019-07-12
Improvements to dev tools and flow with cf-operator

f0rmiga:
## Description

Improved the dev tools and flow with the cf-operator by adding custom Bazel rules and targets for Minikube and Kind. Also added kubectl rules, and extra helm rules to support all the commands to be injected and not depend on any external dependencies. Also moved the repository dependency versions to the top level `def.bzl` for a single source of truth for versions in the repository.

### Minikube targets

- Start the Minikube cluster:
```
bazel run //dev/minikube:start
```

- Stop the Minikube cluster (after verifying the SCF targets below):
```
bazel run //dev/minikube:delete
```

### Kind targets

- Start the Kind cluster:
```
bazel run //dev/kind:start
```

- Stop the Kind cluster (after verifying the SCF targets below):
```
bazel run //dev/kind:delete
```

### SCF targets

- Apply SCF (it will fail to apply the cf-operator custom resources because the cf-operator is not running, it's fine as these changes don't need to validate the SCF chart, but only the dev tools):
```
bazel run //dev/scf:apply
```
- Delete SCF:
```
bazel run //dev/scf:delete
```


dev/kind/BUILD.bazel
```diff
@@ -1,11 +1,40 @@
...
+)
+
+kind_delete(
+    name = "delete",
+    cluster_name = CLUSTER_NAME,
+)
+
+helm_template(
+    name = "rendered_scf",
+    set_values = {
```
viovanov: All values that go into the helm install should come from `value.yaml` files.
Calculated values should be persisted in a **discoverable** and well-known file.

The goal is to allow maximum flexibility to the user, but not constrain the usage of usual commands (currently, if I wanted to use `helm upgrade`, I'd have to remember to `--set system_domain`.

dev/minikube/BUILD.bazel
```diff
@@ -1,6 +1,33 @@
...
+
+minikube_start(
     name = "start",
-    srcs = ["start.sh"],
+)
+
+helm_template(
+    name = "rendered_scf",
+    set_values = {
+        "system_domain": "192.168.99.100.xip.io",
```
viovanov: Same as before - move this to a values file.
Also, this calculation of this "default dev values yaml" can happen in a separate target.

Separation of concern is achieved nicely:
- targets to build the helm chart
- targets to create development environments
- targets to generate configuration for well-known environments
- targets to install

f0rmiga: I'll skip the targets to generate configurations for well-known environments as the requirements for this is not well-defined yet.

dev/kind/README.md
```diff
@@ -9,11 +10,19 @@ bazel run //dev/kind:start
...
-bazel run //dev/kind:clean
+bazel run //dev/kind:delete
+```
+
+## Deploying SCF to Kind
+
+With Kind started and the cf-operator running for the namespace `scf`, run the target:
+
+```txt
+bazel run //dev/kind:apply_scf
```
andreas-kupries: `kind:apply_scf` does not exist (anymore)

dev/minikube/README.md
```diff
@@ -24,9 +18,17 @@ Minikube:
...
 ## Specifying a different Kubernetes version
 
 Set the `K8S_VERSION` environment variable to override the default version.
+
+## Deploying SCF to Minikube
+
+With Minikube started and the cf-operator running for the namespace `scf`, run the target:
+
+```txt
+bazel run //dev/minikube:apply_scf
```
andreas-kupries: `minikube:apply_scf` does not exist (anymore)

dev/scf/README.md
```diff
@@ -1,7 +1,21 @@
 # SCF
 
-For running SCF from source, run the start target:
+The targets under `scf` are used to apply the rendered Helm template to a Kubernetes cluster.
+Any `*values.yaml` files under this directory is ignored by git, but used by Bazel to render the SCF
```
andreas-kupries:
```
-is ignored
+are ignored
```

dev/scf/values.yaml
```diff
@@ -1,2 +1 @@
----
-system_domain: cf-dev.io
+system_domain: scf.suse.dev
```
andreas-kupries: Has Chris confirmed that this is set up ?

f0rmiga: This is a temporary value and is irrelevant at this point.

andreas-kupries: Ok.

PR #2558, opened 2019-07-13
Generate pre_render_scripts ops-files

f0rmiga:
## Description

Added a `pre_render_scripts` ops-file generator. It generates ops-files that are included in the final chart under `assets/operations/pre_render_scripts`.

## Test plan

Blobstore should work with its patch being a generated ops-file now.


bosh/releases/patches/generator/pkg/generator/BUILD.bazel
```diff
@@ -0,0 +1,16 @@
+load("@io_bazel_rules_go//go:def.bzl", "go_library", "go_test")
+
+go_library(
+    name = "go_default_library",
+    srcs = ["generator.go"],
+    importpath = "github.com/SUSE/scf/bosh/releases/patches/generator/pkg/generator",
+    visibility = ["//visibility:public"],
+    deps = ["@in_gopkg_yaml_v2//:go_default_library"],
+)
```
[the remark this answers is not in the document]
f0rmiga: It is not duplicated. The `go_*` targets are generated by Gazelle: https://github.com/bazelbuild/bazel-gazelle. The one under `//bosh/releases/patches/generator/BUILD.bazel` is for the files under that package, which is the `main` package.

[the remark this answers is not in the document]
f0rmiga: We could manage the Go targets manually, but with Gazelle it is waaaaaaaaaaay easier, as it generates things automatically. You can check with `bazel run //:gazelle`.

andreas-kupries:
Thanks to Vlad for (t)his script. I was uneasy about the PR when I reviewed late friday, and your comment put the finger on it, I believe now, namely that a shell script is good enough and complexity of a go solution not warranted.

Currently have only two nits for the script.

  1. The last line `rm -rf "$tmpdir"` should be `trap "rm -rf '$tmpdir'" EXIT` and move up to immediately after `tmpdir=...`. I.e. Ensure removal even in case of errors, and place where created for maximum effect.

  1. The path `/instance_groups/name=$instance_group/jobs/name=$job/properties/bosh_containerization/pre_render_scripts/-` should have an optionality marker (`?`) on `properties`, maybe just `bosh_containerization` if `properties` can be assumed to exist.

Beyond that it looks good to me.

bosh/releases/generators/pre_render_scripts/generator.sh
```diff
@@ -0,0 +1,14 @@
...
+touch "${OUTPUT}"
+
+function append {
+  echo "${1}" >> "${OUTPUT}"
+}
+
+append "- type: replace"
+append "  path: /instance_groups/name=${INSTANCE_GROUP}/jobs/name=${JOB}/properties/bosh_containerization?/pre_render_scripts/-"
+append "  value: |"
+append "$(cat "${PRE_RENDER_SCRIPT}" | sed 's/^/    /')"
```
andreas-kupries: How about  `cat "${PRE_RENDER_SCRIPT}" | sed 's/^/    /' >> "${OUTPUT}"` instead ?
The current construction pulls the entire pre-render-script into memory before writing it out again.
The proposed construction avoids that, streaming it directly from input to output without taking memory.

andreas-kupries: I should say that this is a nit.

f0rmiga: I don't think the size of the script is significant. In any case, it's always good to have it done in the best way. In this case, it doesn't sacrifice readability, so I used your suggestion. I also simplified the script removing the function. Thanks for the suggestion.

andreas-kupries:
Approved again. Last commit fixes issues apparently caused by mismatched lineendings in patch vs patched file.

PR #2562, opened 2019-07-15
Release candidate support

mook-as:
## Description

This branch modifies the Jenkins build process such that when it is built from the `release-candidate` branch, it does some extra things:
- Drops the `-rc?` suffix on the artifacts (so it looks like a release build).
- Tags the repo with the RC release number (so we can go back and look at it).
- Jenkins will only build this on a manual trigger, or updating the [change log](https://github.com/SUSE/scf/blob/develop/CHANGELOG.md).

## Test plan

Please check the new `scf-release-candidate` Jenkins pipeline (not linking because this is a public PR). It's currently set up to build from the PR branch instead of the final RC branch, but otherwise it should be the same.  I manually deleted the extra `0.0.0-rc1` tag it created (I modified the changelog so the top entry was version `0.0.0`).

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -634,6 +634,37 @@ pipeline {
             }
         }
 
+        stage('Push Release Canididate Tag') {
+            when {
+                expression { return getBuildType() == BuildType.ReleaseCandidate }
+            }
+            steps {
+                withCredentials([usernamePassword(
+                        credentialsId: scm.userRemoteConfigs[0].credentialsId,
+                        usernameVariable: 'GIT_USERNAME',
+                        passwordVariable: 'GIT_PASSWORD',
+                )]) {
+                    sh '''
+                        set -o errexit -o nounset
+                        git config --local --replace-all credential.helper \
+                            '/bin/bash -c "echo \"username=${GIT_USERNAME}\"; echo \"password=${GIT_PASSWORD}\""'
+                        # Figure out what the tag was from earlier in this build pipeline
+                        git config --add versionsort.suffix '-alpha'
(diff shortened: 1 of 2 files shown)


jandubois:
Can we use a shorter branch name than `release-candidate`? Maybe `rc` (preferred), or for compromise `relcandi`? (we also have `develop` and not `development`...)

---

My bikeshed is light-blue with pink polka dots!

mook-as:
I'm okay with `rc`; `relcandi` just sounds weird to my ear. We could also try `release`, if we want. (Though if both `develop` and `release` are verbs, does that mean `master` is also being used as a verb?)

----
My bike shed is CGA-style magenta, and too low resolution to have polka dots.

Jenkinsfile
```diff
@@ -564,7 +639,38 @@ pipeline {
             }
         }
 
-        stage('deploy') {
+        stage('Push Release Canididate Tag') {
```
andreas-kupries: `-Canididate`
`+Candidate`

mook-as: Fixed in eb11eec9729d84912c8df6947470fbfba550f32f.

@jandubois Any final word on ~the colour of the bikeshed~ the branch name?

jandubois: > @jandubois Any final word on ~the colour of the bikeshed~ the branch name?

Either `rc` or `release` works for me, but I prefer `rc` because it includes the `candidate` part. `release` sounds like a synonym for `master`.

mook-as: Branch name switched to `rc` in 3861cee61860961cf6356262eda6ac1e116c1cea

PR #2567, opened 2019-07-16
Add router to scf-v2

viovanov:


deploy/helm/scf/values.yaml
```diff
@@ -1 +1,2 @@
-system_domain: ~
+system_domain: cf-dev.io
```
f0rmiga: The default in the main `values.yaml` shouldn't be `cf-dev.io`.

deploy/helm/scf/templates/implicit_vars.yaml
```diff
@@ -11,3 +12,17 @@ metadata:
 type: Opaque
 stringData:
   value: {{ .Values.system_domain | quote }}
+---
```
f0rmiga: It would be better to handle the `deployment_name` in a separate PR, as we need to change other places.

f0rmiga: E.g. https://github.com/SUSE/scf/blob/v2-develop/deploy/helm/scf/templates/bosh_deployment.yaml#L6

andreas-kupries:
Not working for me, under `KinD`.
Both router instances are in post-init CLBO, reporting

```
{"log_level":3,"timestamp":1563314396.029336,"message":"error-fetching-token","source":"vcap.gorouter.uaa-client","data":{"error":"status code: 404, body: {}"}}
{"log_level":6,"timestamp":1563314396.0293424,"message":"routing-api-connection-failed","source":"vcap.gorouter","data":{"error":"unable-to-fetch-token: status code: 404, body: {}"}}
```

deploy/helm/scf/templates/implicit_vars.yaml
f0rmiga: Pushed a commit to fix all the remaining places with `deployment_name`.

viovanov:
Routing is still in a failing state, needs the routing-api in the api-group to work.
This will follow in a separate PR.

PR #2580, opened 2019-07-17
dev/scf: More docs on spinning up

mook-as:
## Description

Add more documentation on starting SCF for development.

## Test plan

Follow the docs, see SCF deployment start (but probably not finish).

### DIFF
--- a/dev/README.md
+++ b/dev/README.md
@@ -7,4 +7,4 @@ This sub-directory contains the resources related to SCF development.
 | minikube | [minikube/README.md](minikube/README.md) |
 | kind     | [kind/README.md](kind/README.md)         |
 
-Also see SCF develop docs in [scf/README.md](scf/README.md).
+Also, see SCF develop docs in [scf/README.md](scf/README.md).
--- a/dev/README.md
+++ b/dev/README.md
@@ -6,3 +6,5 @@ This sub-directory contains the resources related to SCF development.
 | -------- | ---------------------------------------- |
 | minikube | [minikube/README.md](minikube/README.md) |
 | kind     | [kind/README.md](kind/README.md)         |
+
+Also see SCF develop docs in [scf/README.md](scf/README.md).
--- a/dev/scf/README.md
+++ b/dev/scf/README.md
@@ -4,11 +4,65 @@ The targets under `scf` are used to apply the rendered Helm template to a Kubern
 Any `*values.yaml` files under this directory are ignored by git, but used by Bazel to render the
 SCF chart before applying to the cluster with kubectl.
 
+## Prerequesites
+
+You need to have [cf-operator] installed in your Kubernetes cluster (currently
+tip of `master` is required).
+
+[cf-operator]: https://github.com/cloudfoundry-incubator/cf-operator
+
+### Building cf-operator
+
+Build the operator image from a `cf-operator` checkout:
+
+```sh
+SAVE_TARBALL=true make build-image
+```
+
+### Installing cf-operator
+
+Load the cf-operator, depending on if you're using `kind` or `minikube`:
+
+#### When using `kind`
+
+Load the cf-operator image in `kind`:
+
+```sh
+kind load image-archive --name scf binaries/cf-operator-image.tgz
+```
+
+#### When using `minikube`
+
+Load the cf-operator image in `minikube`:
+
+```sh
+eval $(minikube docker-env)
+docker load --input binaries/cf-operator-image.tgz
+```
+
+#### Installing cf-operator after loading
+
+Run the following commands to run `cf-operator` _locally_, taking to your
+kubernetes cluster:
+
+```sh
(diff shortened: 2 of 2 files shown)


dev/README.md
```diff
@@ -6,3 +6,5 @@ This sub-directory contains the resources related to SCF development.
 | -------- | ---------------------------------------- |
 | minikube | [minikube/README.md](minikube/README.md) |
 | kind     | [kind/README.md](kind/README.md)         |
+
+Also see SCF develop docs in [scf/README.md](scf/README.md).
```
f0rmiga: Missing `,` after `Also`.

dev/scf/README.md
```diff
@@ -4,18 +4,72 @@ The targets under `scf` are used to apply the rendered Helm template to a Kubern
...
+SAVE_TARBALL=true make build-image
+```
+
+### Installing cf-operator
+
+Load the cf-operator, depending on if you're using `kind` or `minikube`:
+
+#### When using `kind`
+
+Load the cf-operator image in `kind`:
```
f0rmiga: I think this phrase and the Minikube one are redundant.

dev/scf/README.md
```diff
@@ -4,18 +4,72 @@ The targets under `scf` are used to apply the rendered Helm template to a Kubern
 Any `*values.yaml` files under this directory are ignored by git, but used by Bazel to render the
 SCF chart before applying to the cluster with kubectl.
 
+## Prerequesites
+
+You need to have [cf-operator] installed in your Kubernetes cluster (currently
```
f0rmiga: Missing `,` after `currently`.

dev/scf/README.md
f0rmiga: Or you can keep these and remove the ```Load the cf-operator, depending on if you're using `kind` or `minikube`:```.

dev/scf/README.md
```diff
@@ -4,18 +4,72 @@ The targets under `scf` are used to apply the rendered Helm template to a Kubern
...
+```sh
+kind load image-archive --name scf binaries/cf-operator-image.tgz
+```
+
+#### When using `minikube`
+
+Load the cf-operator image in `minikube`:
+
+```sh
+eval $(minikube docker-env)
```
f0rmiga: I'd prefer giving this instruction in a subshell, so the `eval` doesn't persist:

```
(eval $(minikube docker-env); docker load --input binaries/cf-operator-image.tgz)
```

dev/scf/README.md
```diff
@@ -4,18 +4,72 @@ The targets under `scf` are used to apply the rendered Helm template to a Kubern
...
+Run the following commands to run `cf-operator` _locally_, taking to your
+kubernetes cluster:
+
+```sh
+kubectl create namespace scf
+
+# Get the default network device, being the word after "dev".
+DEFAULT_NET_DEVICE="$(ip route list 0/0 | perl -n -e '/\bdev\s+(\S+)/ && print $1')"
+
+# Get the IP address of that device
```
f0rmiga: Missing punctuation here and below for consistency with the first comment.

dev/scf/README.md
```diff
@@ -4,18 +4,72 @@ The targets under `scf` are used to apply the rendered Helm template to a Kubern
...
+# Deploy cf-operator
+CF_OPERATOR_WEBHOOK_SERVICE_HOST="${DEFAULT_NET_ADDR}" CF_OPERATOR_NAMESPACE=scf SKIP_IMAGE=true make up
+```
+
 ## Apply
 
 Run:
 
-```txt
+```sh
```
f0rmiga: I usually find command-line examples to render weird with `sh`.

mook-as: I prefer using `sh` because it can give you better syntax highlighting; it helps read the examples (since you don't _always_ want to just copy/paste).

(Also, "\`\`\`shell" appears to be always wrong, because it's not a valid language.)

dev/README.md
mook-as: Addressed in 23e13e9231af4d88ea31f53bd0888e3fb2bcb86c

dev/scf/README.md
mook-as: Removed this phrase and the next in 23e13e9231af4d88ea31f53bd0888e3fb2bcb86c.

dev/scf/README.md
mook-as: Changed in 23e13e9231af4d88ea31f53bd0888e3fb2bcb86c.

dev/scf/README.md
mook-as: Added in 23e13e9231af4d88ea31f53bd0888e3fb2bcb86c.

dev/scf/README.md
mook-as: Intentionally not addressed.

PR #2583, opened 2019-07-18
Add shellcheck linter

f0rmiga:
## Description

This PR adds shellcheck and fixes current linting issues on v2-develop.

## Test plan

The scripts should behave as before (considering the linting issues were not critical).


bosh/releases/generators/pre_render_scripts/generator.sh
```diff
@@ -4,7 +4,9 @@ set -o errexit -o nounset
...
-echo "- type: replace" >> "${OUTPUT}"
-echo "  path: /instance_groups/name=${INSTANCE_GROUP}/jobs/name=${JOB}/properties/bosh_containerization?/pre_render_scripts/-" >> "${OUTPUT}"
-echo "  value: |" >> "${OUTPUT}"
-cat "${PRE_RENDER_SCRIPT}" | sed 's/^/    /' >> "${OUTPUT}"
+{
+  echo "- type: replace";
+  echo "  path: /instance_groups/name=${INSTANCE_GROUP}/jobs/name=${JOB}/properties/bosh_containerization?/pre_render_scripts/-";
+  echo "  value: |";
+  sed 's/^/    /' < "${PRE_RENDER_SCRIPT}";
+} >> "${OUTPUT}"
```
mook-as: Using heredoc might be nice (but totally optional); untested:

```sh
{
cat <<EOF
- type: replace
  path: /instance_groups/name=${INSTANCE_GROUP}/jobs/name=${JOB}/properties/bosh_containerization?/pre_render_scripts/-
  value: |
EOF
sed 's/^/    /' < "${PRE_RENDER_SCRIPT}"
} >> "${OUTPUT}
```

(I don't believe you need the explicit semicolons, even in a `{}` brace, if you have new lines.)

dev/kube/klog.sh
```diff
@@ -29,9 +29,17 @@ if [ "${FORCE}" == "1" ] ; then
   rm -f "${DONE}" 2> /dev/null
 fi
 
+function red() {
+  printf '\e[31m%b\e[0m\n' "$1"
+}
+
+function green() {
+  printf '\e[32m%b\e[0m' "$1"
+}
```
mook-as: Why does `red` include a new line, but `green` does not?

Is `red` even used?

dev/cf_deployment/bump/set_release_urls.sh
```diff
@@ -5,12 +5,12 @@ set -o errexit -o nounset
 cf_deployment_yaml="deploy/helm/scf/assets/cf-deployment.yml"
 dest="deploy/helm/scf/assets/operations/set_release_urls.yaml"
 
-printf "# This ops file sets the Quarks images for the releases.\n\n" > "${dest}"
+printf "# This ops file sets the Quarks images for the releases.\\n\\n" > "${dest}"
```
mook-as: FWIW, I feel like using single quotes for the formatting string is clearer. But double quotes and escaping the backslash is fine too.

dev/kube/klog.sh
```diff
@@ -94,57 +102,59 @@ function handle_container() {
...
 
 rm -rf "${KLOG:?}/${NS:?}"
 NAMESPACE_DIR="${KLOG}/${NS}"
 
 # Iterate over pods and their containers.
-PODS=($(get_all_the_pods))
+PODS="$(get_all_the_pods)"
 
-for POD in "${PODS[@]}"; do
+for POD in ${PODS}; do
```
mook-as: Since we don't re-use it, `for POD in $(get_all_the_pods); do` should work too.

dev/kube/klog.sh
```diff
@@ -94,57 +102,59 @@ function handle_container() {
...
   PHASE="$(get_pod_phase)"
 
-  printf "Pod \e[0;32m$POD\e[0m = $PHASE\n"
+  printf 'Pod %s = %s\n' "$(green "${POD}")" "${PHASE}"
 
   # Iterate over containers and dump logs.
-  CONTAINERS=($(get_init_containers_of_pod))
-  for CONTAINER in "${CONTAINERS[@]}"; do
+  CONTAINERS="$(get_init_containers_of_pod)"
+  for CONTAINER in ${CONTAINERS}; do
```
mook-as: Here, I think just `for CONTAINER in $(get_init_containers_of_pod)` would be clearer too (especially since we stop reusing `CONTAINERS`).

dev/kube/kube-ready-state-check.sh
```diff
@@ -41,22 +44,23 @@ function has_command() {
...
+    printf 'Verified: %s\n' "$(green "${1}")"
 }
 
 function trouble() {
-    red "Configuration problem detected: $1"
+    printf 'Configuration problem detected: %s\n' "$(red "${1}")"
 }
 
 function status() {
+    # shellcheck disable=SC2181
```
mook-as: We should have a comment on why we're disabling the shellcheck. (But, as mentioned offline, I'm for disabling that here.)

dev/linters/shellcheck.sh
```diff
@@ -0,0 +1,21 @@
+#!/usr/bin/env bash
+
+set -o errexit
+
+RED='\033[0;31m'
+NC='\033[0m'
```
mook-as: Why do we have a completely different style of colourizing text here?

bosh/releases/generators/pre_render_scripts/generator.sh
f0rmiga: Yeah, looks better. And I just realized I don't need the `touch` above, I can simply `... > "${OUTPUT}"`.

dev/kube/klog.sh
andreas-kupries: Oops. Will fix. And yes, look for `Already done`.

andreas-kupries: Oh. I somehow lost that piece :(

mook-as: If you mean 4 lines lower… that's uncolourized.

andreas-kupries: Fixed with commit 957cb317c81737bce63b4e6313bba145ab10d694

dev/kube/klog.sh
andreas-kupries: Fixed with commit b6e4b4ddb570770345aa1acdb867ed2c093ed925

dev/kube/klog.sh
andreas-kupries: Fixed with commit b6e4b4ddb570770345aa1acdb867ed2c093ed925

dev/linters/shellcheck.sh
andreas-kupries: Historical

dev/kube/kube-ready-state-check.sh
andreas-kupries: Fixed with commit 4a9ca8da6052eb1654d29584c17c897af307b310

PR #2585, opened 2019-07-18
Fixed gorouter connecting to nats

f0rmiga:
## Description

Gorouter was not being able to find nats.

## Test plan

Gorouter should show in the logs:
```
{"log_level":1,"timestamp":1563488075.3089044,"message":"setting-up-nats-connection","source":"vcap.gorouter","data":{}}                                                                                           
{"log_level":1,"timestamp":1563488075.3108475,"message":"Successfully-connected-to-nats","source":"vcap.gorouter.nats","data":{"host":"scf-dev-nats:4222"}}
```

### DIFF
--- a/deploy/helm/scf/assets/operations/instance_groups/router.yaml
+++ b/deploy/helm/scf/assets/operations/instance_groups/router.yaml
@@ -11,7 +11,7 @@
   value: ((deployment-name))-uaa
 - type: replace
   path: /instance_groups/name=router/jobs/name=gorouter/properties/routing_api/uri?
-  value: https://((deployment-name))-api-group
+  value: https://((deployment-name))-api
 - type: replace
   path: /instance_groups/name=router/jobs/name=gorouter/properties/nats?/machines
   value:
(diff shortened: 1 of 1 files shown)


deploy/helm/scf/assets/operations/instance_groups/router.yaml
```diff
@@ -2,14 +2,17 @@
...
   path: /instance_groups/name=router/jobs/name=gorouter/properties/router/dns_health_check_host?
   value: ((deployment-name))-uaa
-
 - type: replace
   path: /instance_groups/name=router/jobs/name=gorouter/properties/uaa/token_endpoint?
   value: ((deployment-name))-uaa
-
 - type: replace
   path: /instance_groups/name=router/jobs/name=gorouter/properties/routing_api/uri?
   value: https://((deployment-name))-api-group
```
mook-as: FWIW, this should be `api` not `api-group`; we just called it `api-group` in SCF v1 because we couldn't handle the conflict with the job name at the time.

f0rmiga: Thanks. This is from the first router commit. I didn't notice this.

PR #2586, opened 2019-07-18
Fix UAA hostnames

f0rmiga:
## Description

Fixed UAA hostnames.

## Test plan

NA. UAA will accept connections using the `((deployment-name))-uaa` hostname.


mook-as:
FWIW, my test step was "Look at the router logs and check that it can get a token from UAA".

PR #2594, opened 2019-07-19
Add uaadb back to list of seeded databases

jandubois:
This was accidentally removed in 42572cd when reverting back from pxc to mysql.

[CAP-560]


jandubois:
> This is not an issue for 1.4.1 (because we didn't support bundled UAA), right?

Well, we claimed that we added this feature for 1.4.1, but unfortunately it is broken.

It may just be removed from the release notes; idk.

PR #2600, opened 2019-07-24
Updated Minikube image and added VM_DRIVER option

f0rmiga:
## Description

It now supports `VM_DRIVER=kvm2`.

## Test plan

Check that `bazel run //dev/minikube:start` works with `virtualbox` and `kvm2`.

### DIFF
--- a/dev/minikube/README.md
+++ b/dev/minikube/README.md
@@ -11,6 +11,9 @@ bazel run //dev/minikube:start
 The following environment variables are used by the `start` target to allocate the resources used by
 Minikube:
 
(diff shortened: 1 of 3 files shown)


f0rmiga:
> This should work, or we can do `${VM_DRIVER:+--vm-driver ${VM_DRIVER}}` so that the default (whatever `minikube config get vm-driver` shows) would be used instead. Up to you (since I can always `.envrc` all the things).

I agree! Thanks for pointing it out.

f0rmiga:
This PR was closed automatically by the v2-develop to v3-develop rename.

PR #2601, opened 2019-07-24
Make the release URLs customizable

mook-as:
## Description

This lets us change the release URLs (actually the docker org for the
images to use); also, refactor how we do implicit variables a bit to
reduce duplication.

No change to default behaviour.

Closes jsc#CAP-553.

## Test plan

A normal `bazel run //dev/scf:apply` should have no changes. However, we should now be able to override the prefix to something different (which will not work because we don't have those images).

### DIFF
--- a/deploy/helm/scf/values.yaml
+++ b/deploy/helm/scf/values.yaml
@@ -1,2 +1,3 @@
 system_domain: ~
 deployment_name: scf
+container_image_prefix: docker.io/cfcontainerization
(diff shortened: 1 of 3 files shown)


deploy/helm/scf/values.yaml
```diff
@@ -1,2 +1,3 @@
 system_domain: ~
 deployment_name: scf
+container_image_prefix: docker.io/cfcontainerization
```
f0rmiga: I wonder if it would be better something like:
```
images:
  repository_prefix: docker.io/cfcontainerization
```

f0rmiga:
This PR was closed automatically by the v2-develop to v3-develop rename.

PR #2603, opened 2019-07-24
Make the release URLs customizable

mook-as:
## Description

This lets us change the release URLs (actually the docker org for the images to use); also, refactor how we do implicit variables a bit to reduce duplication.

No change to default behaviour.

Closes jsc#CAP-553.  Supercedes #2601 (due to the change in the base branch).

## Test plan

A normal `bazel run //dev/scf:apply` should have no changes. However, we should now be able to override the prefix to something different (which will not work because we don't have those images)

### DIFF
--- a/deploy/helm/scf/templates/implicit_vars.yaml
+++ b/deploy/helm/scf/templates/implicit_vars.yaml
@@ -1,8 +1,16 @@
+{{- /*
+  Template "scf.implicit-var" generates the kube secret declaration for a
+  variable.  It takes a list of two arguments: the context, and the variable
+  name.
+*/ -}}
+{{- define "scf.implicit-var" }}
+{{- $variable_name := (last .) }}
+{{- with (first .) }}
 ---
 apiVersion: v1
 kind: Secret
 metadata:
-  name: {{ .Values.deployment_name }}.var-system-domain
+  name: {{ .Values.deployment_name }}.var-{{ $variable_name | replace "_" "-" | replace "." "-" }}
   labels:
     app.kubernetes.io/instance: {{ .Release.Name | quote }}
     app.kubernetes.io/managed-by: {{ .Release.Service | quote }}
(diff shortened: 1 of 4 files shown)


deploy/helm/scf/templates/implicit_vars.yaml
```diff
@@ -11,18 +19,10 @@ metadata:
...
-  labels:
-    app.kubernetes.io/instance: {{ .Release.Name | quote }}
-    app.kubernetes.io/managed-by: {{ .Release.Service | quote }}
-    app.kubernetes.io/name: {{ include "scf.fullname" . }}
-    app.kubernetes.io/version: {{ default .Chart.Version .Chart.AppVersion | quote }}
-    helm.sh/chart: {{ include "scf.chart" . }}
-type: Opaque
-stringData:
-  value: {{ .Values.deployment_name | quote }}
+  value: {{ template "scf.dig" (list .Values ( splitList "." $variable_name )) }}
```
f0rmiga: Should we fail helm template if the value is not a string?

deploy/helm/scf/values.yaml
```diff
@@ -1,2 +1,6 @@
 system_domain: ~
 deployment_name: scf
+releases:
+  # The defaults for all release, where we do not otherwise override them.
```
f0rmiga: release -> releases

deploy/helm/scf/templates/implicit_vars.yaml
mook-as: The `scf.dig` template does the `| quote` part already; unless I'm missing something?

deploy/helm/scf/values.yaml
mook-as: Fixed in d99eded3f0ac6b69429ae70215b5f93052b910aa (which I force-pushed because I forgot, sorry)

PR #2615, opened 2019-07-26
Copy folder on darwin OS

edwardstudy:
## Description

Running on mac, where the `cp -L --recursive` in rules/helm/package.sh fails. This PR worked on replacing `cp` command with `rsync`. Most machines will have rsync available.

## Test plan

Run `bazel run //dev/scf:apply --verbose_failures --sandbox_debug` on different OS.

### DIFF
--- a/rules/helm/package.sh
+++ b/rules/helm/package.sh
@@ -4,7 +4,7 @@ set -o errexit -o nounset
 
 build_dir="tmp/build/${PACKAGE_DIR}"
 mkdir -p "${build_dir}"
-cp -L --recursive "${PACKAGE_DIR}"/* "${build_dir}"
+cp -L -R "${PACKAGE_DIR}"/* "${build_dir}"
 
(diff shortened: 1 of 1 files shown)


rules/helm/package.sh
```diff
@@ -4,7 +4,7 @@ set -o errexit -o nounset
 
 build_dir="tmp/build/${PACKAGE_DIR}"
 mkdir -p "${build_dir}"
-cp -L --recursive "${PACKAGE_DIR}"/* "${build_dir}"
+rsync -a "${PACKAGE_DIR}"/ "${build_dir}"/
```
f0rmiga: rsync messes with the final .tgz output. I confirmed that OS X dislikes `--recursive` as much as it dislikes `--dereference`. So, could you do a `cp -L -R "${PACKAGE_DIR}"/* "${build_dir}"` instead? It will work cross-platform.

PR #2627, opened 2019-07-29
v3 - Added cc_uploader job

f0rmiga:
## Description

Added cc_uploader job to the api instance group.

## Test plan

Check that cc_uploader is running.

### DIFF
--- a/bosh/releases/pre_render_scripts/api/cc_uploader/patch_pre-start.sh
+++ b/bosh/releases/pre_render_scripts/api/cc_uploader/patch_pre-start.sh
@@ -1,0 +1,16 @@
+#!/usr/bin/env bash
+
+set -o errexit -o nounset
+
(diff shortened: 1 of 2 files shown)


bosh/releases/pre_render_scripts/api/cc_uploader/patch_pre-start.sh
```diff
@@ -0,0 +1,16 @@
+#!/usr/bin/env bash
+
+set -o errexit -o nounset
+
+target="/var/vcap/all-releases/jobs-src/capi/cc_uploader/templates/pre-start.erb"
+
+# Remove sysctl calls as we are running in containers.
```
mook-as: It would be better to point to the other script too, as better justification.

f0rmiga: Addressed on https://github.com/SUSE/scf/pull/2628.

PR #2633, opened 2019-07-30
v3 - Specify pod values

edwardstudy:
## Description

- Supported `TRAFFIC_CONTROLLER_IP ` env value 

** Maybe it's **working in the process** that we should replace other values by pod fields.

## Test plan
After deployment, 
IBMMasterMBP:~ bjxzi$ k exec -it -n scf scf-dev-log-api-v1-0 -c loggregator-trafficcontroller-loggregator-trafficcontroller -- bash
/:/# env | grep TRAFFIC_CONTROLLER_IP
TRAFFIC_CONTROLLER_IP=172.30.23.17

### DIFF
--- a/deploy/helm/scf/assets/operations/instance_groups/log-api.yaml
+++ b/deploy/helm/scf/assets/operations/instance_groups/log-api.yaml
@@ -48,3 +48,12 @@
     run:
       memory: 128
       virtual-cpus: 2
+
+- type: replace
+  path: /instance_groups/name=log-api/jobs/name=loggregator_trafficcontroller/properties/bosh_containerization/envs?/-
+  value:
+    name: TRAFFIC_CONTROLLER_IP
+    valueFrom:
+      fieldRef:
+        apiVersion: v1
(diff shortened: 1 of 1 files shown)


deploy/helm/scf/assets/operations/instance_groups/log-api.yaml
```diff
@@ -48,3 +48,12 @@
     run:
       memory: 128
       virtual-cpus: 2
+
+- type: replace
```
f0rmiga: We don't need this `replace`. We already have a `replace` adding `quarks` (you will need to rebase with v3-develop where it is renamed from `bosh_containerization` to `quarks`) on line 42. You will just need to add:
```
    envs:
    - name: TRAFFIC_CONTROLLER_IP
      valueFrom:
        fieldRef:
          apiVersion: v1
          fieldPath: status.podIP
```

f0rmiga:
Ended fixing on https://github.com/SUSE/scf/pull/2668.

PR #2582, opened 2019-07-18
Make Eirini work again and drop fluentd (by adding the eirini-loggregator-bridge)

mudler:
This PR is dropping the fluentd daemonset, and it is replacing the cert-copier stage with a kube [job](https://github.com/cloudfoundry-community/eirini-bosh-release/blob/master/jobs/configure-eirini-scf/templates/eirini-cert-copier.yaml.erb). @jandubois  we tried to make changes as small as possible to not duplicate the work and ease the merge with the work you did so far already.

It adds https://github.com/SUSE/eirini-loggregator-bridge for logging, consumed from the eirini-bosh-release.
Supersedes: https://github.com/SUSE/scf/pull/2589 https://github.com/SUSE/scf/pull/2470 https://github.com/SUSE/scf/pull/2640

PR #2637, opened 2019-07-31
Remove bits service hack

alex-slynko:
## Description

Remove bits service hack. Bits service has properties to use it as a registry..
Additionally, bump the bits service.

## Test plan

Build, deploy with Eirini support, cf push application.


jandubois:
I believe everything in this PR is already covered by #2582, which is expected to land very soon (we just finished the bump to cf-deployment-9.5, that was required to get the the latest CAPI release with Eirini related changes).

alex-slynko:
cool, thanks

PR #2649, opened 2019-08-06
Get rid of cluster-admin binding for the eirini service account

jandubois:
### Description

Instead we create an eirini role with full access inside the eirini namespace and bind it to the eirini service account in the release namespace.

Replaces #2520.
[jsc#CAP-485]

### Test plan

Deploy SCF with `--set enable.eirini=true` and verify that you can push an app.

### DIFF
--- a/chart-parts/eirini-namespace.yaml
+++ b/chart-parts/eirini-namespace.yaml
@@ -1,0 +1,38 @@
+{{- if and .Values.enable.eirini }}
+
+{{- if ne .Values.env.EIRINI_KUBE_NAMESPACE .Release.Namespace }}
+---
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: {{.Values.env.EIRINI_KUBE_NAMESPACE | quote}}
+{{- end }}
+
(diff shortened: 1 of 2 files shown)


chart-parts/eirini-namespace.yaml
```diff
@@ -0,0 +1,38 @@
+{{- if and .Values.enable.eirini }}
```
f0rmiga: You can drop `and` here (unless it's doing something special that I'm not aware of).

jandubois: No, it is not supposed to be there. The line used to have a more complex condition, but was later refactored, but the `and` stayed behind. Will fix.

bikramnehra:
Validated the changes on vagrant using: `make run-eirini`. Also `smoke-tests` are looking good there. Will merge after CATs finish on Jenkins.

bikramnehra:
After talking to @jandubois and @f0rmiga waiting for CATs to pass seems un-necessary. So will go ahead and merge now.

PR #2674, opened 2019-08-19
Jenkinsfile: Fake a version bump for nightlies

mook-as:
## Description

This ensures that nightly builds will get a secret bump when upgrading from the previous release.

## Test plan

- Everything Jenkins should pass
- The resulting build should be version `2.17.1.1` (yes, not semver compliant), which means the secret generator should fill in new secrets relative to `2.17.1`.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -462,6 +462,23 @@ pipeline {
             }
         }
 
+        stage('Version Bump for Develpoment Builds') {
+            when {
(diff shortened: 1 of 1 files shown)


mook-as:
For reference: We use `GIT_TAG` in [`make/include/versioning`](https://github.com/SUSE/scf/blob/develop/make/include/versioning), which is why this has the desired effect.

PR #2677, opened 2019-08-21
v3 - Fixes to make push work e2e

viovanov:
## Description

This should make `cf push ... -o ...` work without any errors.

## Test plan

Push a docker app.

`cf push ... -o ...`

### DIFF
--- a/deploy/helm/scf/assets/operations/instance_groups/api.yaml
+++ b/deploy/helm/scf/assets/operations/instance_groups/api.yaml
@@ -129,7 +129,7 @@
   value: ((deployment-name))-doppler
 - type: replace
   path: /instance_groups/name=api/jobs/name=cloud_controller_ng/properties/cc/loggregator?/internal_url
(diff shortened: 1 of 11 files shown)


deploy/helm/scf/assets/operations/instance_groups/api.yaml
```diff
@@ -125,7 +82,7 @@
   value: ((deployment-name))-doppler
 - type: replace
   path: /instance_groups/name=api/jobs/name=cloud_controller_ng/properties/cc/loggregator?/internal_url
-  value: http://((deployment-name))-log-api:8081
+  value: https://((deployment-name))-log-api:8081
```
f0rmiga: This is not an HTTPS server: https://github.com/cloudfoundry/capi-release/blob/1.83.0/jobs/cloud_controller_ng/spec#L1097.

f0rmiga: The server is HTTP/2.

PR #2680, opened 2019-08-22
New script to attempt to recover stuck lock

mook-as:
## Description

This adds a script that attempts to recover a stale lock for the `autoscaler-metrics` database.  This can occur if the `autoscaler-metrics` pod was terminated while it's holding the lock, as for some unfathomable reason the library implements locking by inserting data into a table, instead of just using actual (possibly row-level) locking.

## Test plan

1. Deploy SCF, and wait for everything to be ready.
2. Delete the `autoscaler-metrics` pod.
3. Watch the logs of the `autoscaler-metrics` container as the pod restarts.
4. Delete the pod again once you see the `Starting Liquibase at`… message.
5. Wait for the respawned pod until you see error traces about `liquibase.exception.LockException: Could not acquire change log lock.`
6. Run the script and see that it fixes things.

### DIFF
--- a/container-host-files/opt/scf/bin/recovery-autoscaler-metrics-liquidbase-stale-lock
+++ b/container-host-files/opt/scf/bin/recovery-autoscaler-metrics-liquidbase-stale-lock
@@ -66,7 +66,7 @@ run 'kubectl', 'exec', postgres_pod,
 
 locked_bys = (1..Float::INFINITY).each do |count|
     STDOUT.printf "\rWaiting for pod to get stuck (try %s)...", count
-    locked_bys = capture('kubectl', 'get', 'pods',
+    candidate_locked_bys = capture('kubectl', 'get', 'pods',
                         '--namespace', namespace,
                         '--selector', 'skiff-role-name==autoscaler-metrics',
                         '--output', 'jsonpath={.items[*].metadata.name}')
@@ -79,7 +79,7 @@ locked_bys = (1..Float::INFINITY).each do |count|
         .select { |message| message.include? EXPECTED_ERROR }
         .map { |message| /Currently locked by (.*?) since /.match(message)[1]}
 
-    break locked_bys unless locked_bys.empty?
+    break candidate_locked_bys unless candidate_locked_bys.empty?
     sleep 1
 end
 puts '' # new line after the wait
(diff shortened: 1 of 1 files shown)


andreas-kupries:
@mook-as

Going through the test plan I managed to both get the pod stuck, and to unstick it with the recovery script.

It is a bit unwieldy however because of three things:
- The recovery script releases only locks for pods which are detected as stuck by the error message in the log
- When the pod is stuck it takes a few minutes for the error message to appear in the log.
- After the message appears the pod keeps around for only about 10 to 20 seconds and is then reaped by kube.

Together this currently means that the user has to first wait on the log, being bored, and then take quick action to invoke the recovery.

While one possible fix for this would be to drop the check for the error message and simply release all locks, that is unsafe, and the reason why the check is there in the first place.

Another possible fix, and assuming that the script will be run only when we know that a pod is stuck:

- Have the script run the main action (determination of lockedby's, and then lock release) in a loop which stops when a stuck pod is found and the associated lock released. Maybe reduce impact on system load by waiting a second between iterations.
- Determination of postgres pod and psql binary is moved to the top, kept out of this proposed loop.

In other words, let the computer do the bored waiting and quick action.

mook-as:
Change in 1e917ef8d98c0cccf8a0e7a112d6ec17bb0dc313 (based off your modifications).

container-host-files/opt/scf/bin/recovery-autoscaler-metrics-liquidbase-stale-lock
```diff
@@ -0,0 +1,95 @@
...
+
+run 'kubectl', 'exec', postgres_pod,
+    '--namespace', namespace,
+    '--',
+    psql_binary,
+    '--username', 'postgres',
+    '--dbname', 'autoscaler',
+    '--command', 'SELECT * FROM databasechangeloglock'
+
+locked_bys = (1..Float::INFINITY).each do |count|
```
andreas-kupries: This confused me at first, given that it has the same name as the variable two lines below. (line 69).
I then realized that the ruby scoping keeps them separate.
Maybe rename the inner variable to `candidate_locked_by` or some such ?
It is however a nit and does not block merging.

PR #2686, opened 2019-08-26
Eirini ssh

jimmykarily:


jandubois:
Can this be closed?

jimmykarily:
With [this now merged](https://github.com/SUSE/scf/pull/2906) we now have our own eirinifs to which we can add `sshd`. We will update this as soon as we pick up the SSH story again.

PR #2705, opened 2019-08-29
jsc#CAP-554: Allow default stemcell to be configured

mook-as:
## Description

This allows us to configure the default stemcell.  Note, however, the image name is still constructed via joining various parts; it's
`((releases-defaults-url))/((release-name)):((stemcell-os))-((stemcell-version))-((release-version))`.

## Test plan

- Run a normal build, everything stays working
- Edit `dev/scf/values.yaml` to add `stemcell: { os: foo, version: bar }` and see the different docker image used when you apply (check the `bpm` pod).

I'm happy to modify the helm chart `values.yaml` layout &c. if there are better names / placements for things.  I tried to put it under `releases/defaults` but it didn't seem to make any sense.

### DIFF
--- a/deploy/helm/scf/assets/operations/set_opensuse_stemcells.yaml
+++ b/deploy/helm/scf/assets/operations/set_opensuse_stemcells.yaml
@@ -1,10 +1,7 @@
 # This ops file sets the openSUSE stemcells.
 
-- type: remove
-  path: /stemcells/alias=default/alias
-
 - type: replace
-  path: /stemcells/-
+  path: /stemcells/alias=default
   value:
     alias: default
     os: ((stemcell-os))
--- a/deploy/helm/scf/assets/operations/set_opensuse_stemcells.yaml
+++ b/deploy/helm/scf/assets/operations/set_opensuse_stemcells.yaml
@@ -4,8 +4,8 @@
   path: /stemcells/alias=default
   value:
     alias: default
-    os: ((stemcell-os))
-    version: ((stemcell-version))
+    os: ((releases-defaults-stemcell-os))
+    version: ((releases-defaults-stemcell-version))
 
 - type: replace
   path: /addons/name=loggregator_agent/include/stemcell/os=ubuntu-xenial/os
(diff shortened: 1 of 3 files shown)


deploy/helm/scf/assets/operations/set_opensuse_stemcells.yaml
```diff
@@ -1,17 +1,19 @@
 # This ops file sets the openSUSE stemcells.
 
-- type: replace
-  path: /stemcells/alias=default/os
-  value: opensuse-42.3
+- type: remove
```
f0rmiga: Instead of removing and then adding again below, you can do a replace like:
```
- type: replace
  path: /stemcells/alias=default
  value:
    alias: default
    os: ((...))
    version: ((...))
```

deploy/helm/scf/assets/operations/set_opensuse_stemcells.yaml
```diff
@@ -1,17 +1,19 @@
...
+  path: /stemcells/-
+  value:
+    alias: default
+    os: ((stemcell-os))
+    version: ((stemcell-version))
 
 - type: replace
   path: /addons/name=loggregator_agent/include/stemcell/os=ubuntu-xenial/os
-  value: opensuse-42.3
+  value: default
```
f0rmiga: Not sure this works as intended. Wouldn't we need to:
```
- type: replace
  path: /addons/name=.../include/stemcell
  value:
    os: ...
    version: ...
```

deploy/helm/scf/assets/operations/set_opensuse_stemcells.yaml
mook-as: I'm removing just the alias (the old stemcell is still listed).  I could totally change things to remove it too, if you prefer.

deploy/helm/scf/assets/operations/set_opensuse_stemcells.yaml
mook-as: No, I'm just replacing which stemcell a instance group refers to; the actual stemcell definition is the previous hunk.

deploy/helm/scf/assets/operations/set_opensuse_stemcells.yaml
mook-as: Addressed in be7b1425d579880cc7140b75d24f04d23e7249a1 since the old stemcell isn't useful anyway

PR #2706, opened 2019-08-29
Jenkinsfile: Always build manually-triggered builds

mook-as:
## Description

If a build was manually triggered, always allow RC buidls.

## Test plan

- Normal Jenkins run of this should build
- I'll tack on a README-only change; that should _not_ build (because it's a PR)
- I'll then force-push to remove the README change, and manually trigger a build before it scans; that _should_ build.


mook-as:
Okay, [build 7](https://jenkins-new.howdoi.website/blue/organizations/jenkins/scf/detail/PR-2706/7/pipeline/35) shows the thing working correctly (that was a manually triggered build that would otherwise have been skipped).

PR #2708, opened 2019-08-31
v3 - Added ingress

f0rmiga:
## Description

Add NGINX Ingress Controller support for SCF v3.
Also, added instructions for development using the Ingress Controller with Minikube.

## Test plan

Follow the instructions in the updated docs in this very same PR.


dev/scf/docs/accessing.md
```diff
@@ -0,0 +1,18 @@
...
+NGINX Ingress Controller with Minikube enables the access from the host to the SCF cluster. Follow the
+commands below for a step-by-step guide.
+
+```sh
+cf api --skip-ssl-validation "https://api.$(minikube ip).xip.io"
+
+# Copy the admin cluster password.
+kubectl get secret -n scf scf.var-cf-admin-password -o jsonpath='{.data.password}' \
+  | base64 --decode \
+  | less
```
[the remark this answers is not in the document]
f0rmiga: It's not about being long, but not printing the secret to the terminal.

deploy/helm/scf/templates/ingress.yaml
```diff
@@ -0,0 +1,82 @@
...
+            backend:
+              serviceName: "{{ .Values.deployment_name }}-router"
+              servicePort: 443
+    - host: "{{ .Values.system_domain }}"
+      http:
+        paths:
+          - path: "/"
+            backend:
+              serviceName: "{{ .Values.deployment_name }}-router"
+              servicePort: 443
```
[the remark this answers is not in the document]
f0rmiga: Yes, UAA is accessible through the router.

dev/scf/docs/installing.md
```diff
@@ -0,0 +1,112 @@
...
+```sh
+echo "system_domain: $(minikube ip).xip.io" \
+  > "$(bazel info workspace)/dev/scf/system_domain_values.yaml"
+```
+
+#### When using Kind
+
+```sh
+echo "system_domain: scf.suse.dev" \
+  > "$(bazel info workspace)/dev/scf/system_domain_values.yaml"
```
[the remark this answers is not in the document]
f0rmiga: Making `scf.suse.dev` the default value is wrong. We can have a discussion on whether we want specialized Bazel targets to deal with this automatically.

dev/scf/docs/accessing.md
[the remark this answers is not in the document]
mook-as: Then we should put that in the comment

dev/scf/docs/installing.md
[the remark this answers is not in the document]
mook-as: As long as we have a follow up card or something, yep

deploy/helm/scf/templates/ingress.yaml
```diff
@@ -0,0 +1,82 @@
...
+# specification. Although specific controllers may do so in a custom manner.
+apiVersion: "extensions/v1beta1"
+kind: "Ingress"
+metadata:
+  name: {{ .Release.Name | quote }}
+  namespace: {{ .Release.Namespace | quote }}
+  annotations:
+    {{- if hasKey .Values.features.ingress.annotations "kubernetes.io/ingress.class" | not -}}
+      {{ $_ := set .Values.features.ingress.annotations "kubernetes.io/ingress.class" "nginx" }}
+    {{- end }}
```
andreas-kupries: This all feels overly complicated. Could this be replaced by
```
kubernetes.io/ingress.class: {{ default "nginx" .Values.features.ingress.annotations | quote }}
```
And similar for the following clauses.

Alternatively, more questions:
  - Is `if haskey ... | not` the same as `if not (haskey ..)` ? If yes, I find the second more readable. Knowing from the outset that there is an inversion/negation. The current way gives me a surprise negation at the end, and I just managed to see that. Others may miss it entirely.
  - What does `$_ := set X Y Z` mean/do ?

dev/scf/docs/accessing.md
```diff
@@ -0,0 +1,18 @@
+# Accessing the SCF cluster
+
+## When using NGINX Ingress Controller
+
+NGINX Ingress Controller with Minikube enables the access from the host to the SCF cluster. Follow the
```
andreas-kupries: Are there similar changed instructions for us using `KinD` ?
I later see `(optional) NGINX Ingress Controller for external access - ONLY FOR MINIKUBE` so I guess I cannot test the nginx ingress changes via KinD ?

Or will nginx ingress work generally and it is only the external access I cannot have ? (Although that would block me from using `cf`, etc. to actually talk to the CF deployment, right ?

f0rmiga: Right. The purpose of the ingress controller is to give external access to whatever backends it exposes. This is why I marked as an option only for Minikube. With KinD we cannot expose anything except using `kubectl port-forward`, which doesn't help us here.

deploy/helm/scf/templates/ingress.yaml
f0rmiga: Using `default` here is hard because we want to use `{{ toYaml .Values.features.ingress.annotations | indent 4 }}` in line 52. For the bullets:
- I think it's the same. But indeed the way you suggested is more readable.
- There is a nice explanation in the sprig documentation here: http://masterminds.github.io/sprig/dicts.html.

f0rmiga: Improved in https://github.com/SUSE/scf/pull/2708/commits/4dedc618264fdbe13e0074f907b2f354120464b9.

andreas-kupries: I still do not understand why the `toYaml` and the dynamic dict would be needed ...

:bulb: Ah. I might see it now.

I tried to rewrite this fully to defaults, and then noticed that the last element in the key I would use would stuff like "kubernetes.io/ingress.class", which might not be writable as dot notation.

This also triggered the realization that the user is maybe allowed to write __any__ annotation they like in that section. And the `toYaml` will then format and insert that properly. 

And thus it also becomes clear why `default` is not usable for the annotations which are required. If the user specifies nothing then default inserts something. No clash with `toYaml`.

But when the user does provide their own setting we will get it inserted twice, once from the `default and then again from the `toYaml`.

So you are checking explicitly, and dynamically extend the underlying dict instead, and everything goes through `toYaml`.

If that is correct, I would really like to have few comment sentences in the template which explain the basics of this for any future dev.

The main :bulb: was the realization that the `annotation:` key in the values is not a fixed set of things, but completely under user control.

f0rmiga: You are correct. Added the comment in https://github.com/SUSE/scf/pull/2708/commits/c75227ae676c2a4cc649f6443e5d65eb93102e5d.

PR #2712, opened 2019-09-03
v3 - Add initial Github Actions pipeline for publishing

f0rmiga:
## Description

This initial pipeline will publish the built tgz chart to https://scf-v3.s3.amazonaws.com/.

## Test plan

Not applicable.

### DIFF
--- a/.github/workflows/publish/build.sh
+++ b/.github/workflows/publish/build.sh
@@ -1,0 +1,8 @@
+#!/usr/bin/env bash
+
+set -o errexit -o nounset
+
+bazel build //deploy/helm/scf:chart
+built_file="$(realpath bazel-bin/deploy/helm/scf/scf-*.tgz)"
+mkdir -p output
+cp "${built_file}" output/
--- a/.github/workflows/publish/publish.sh
+++ b/.github/workflows/publish/publish.sh
@@ -1,0 +1,8 @@
+#!/bin/sh
+
+set -o errexit -o nounset
+
+alias="s3"
+
+mc config host add "${alias}" "https://s3.amazonaws.com" "${AWS_ACCESS_KEY}" "${AWS_SECRET_KEY}"
+mc cp output/scf-*.tgz "${alias}/${BUCKET}"
--- a/.github/workflows/publish/publish.sh
+++ b/.github/workflows/publish/publish.sh
@@ -6,4 +6,4 @@ alias="s3"
 
 mc config host add "${alias}" "https://s3.amazonaws.com" "${AWS_ACCESS_KEY}" "${AWS_SECRET_KEY}"
 mc cp output/scf-*.tgz "${alias}/${BUCKET}"
-echo "URL: https://${BUCKET}.s3.amazonaws.com/$(ls output/scf-*.tgz)"
+echo "URL: https://${BUCKET}.s3.amazonaws.com/$(find output/ -name 'scf-*.tgz' -print0 | xargs -0 basename)"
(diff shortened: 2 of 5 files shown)


.github/workflows/publish/build.sh
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env bash
+
+set -o errexit -o nounset
+
+bazel build //deploy/helm/scf:chart
+built_file="$(realpath bazel-bin/deploy/helm/scf/scf-*.tgz)"
```
mook-as: I think technically it should involve `$(bazel info bazel-genfiles)` instead of `bazel-bin`, but yeah it's kind of annoying that bazel doesn't give you a good way to get the outputs.

Alternatively we can do `bazel build //deploy/helm/scf:chart --build_event_json_file=…` and then try to parse that file.

.github/workflows/publish/publish.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/sh
+
+set -o errexit -o nounset
+
+alias="s3"
+
+mc config host add "${alias}" "https://s3.amazonaws.com" "${AWS_ACCESS_KEY}" "${AWS_SECRET_KEY}"
+mc cp output/scf-*.tgz "${alias}/${BUCKET}"
+echo "URL: https://${BUCKET}.s3.amazonaws.com/$(ls output/scf-*.tgz)"
```
mook-as: Huh, the file is `https://bucket.s3.amazonaws.com/output/scf-foo.tgz`? (I think the `ls` on the last line will show the `output/` directory name too, which seems incorrect.)

.github/workflows/publish/rename.sh
```diff
@@ -0,0 +1,12 @@
...
+set -o errexit -o nounset
+
+built_file="$(ls output/scf-*.tgz)"
+commit_hash="$(git rev-parse --short HEAD)"
+release_filename="$(basename "${built_file}" .tgz)-${commit_hash}"
+if [ -n "$(git status --porcelain)" ]; then
+  release_filename="${release_filename}-dirty"
+fi
+
+mv "output/${built_file}" "output/${release_filename}.tgz"
```
mook-as: Okay, I think I'm misunderstanding something; I thought this would expand to `mv "output/output/scf-….tgz" "output/…"` which definitely looks like it wouldn't have worked. Unfortunately I couldn't see the GitHub Actions logs.

Also this almost looks like it wants to be `git describe` :p

.github/workflows/publish/build.sh
f0rmiga: `bazel-bin` vs `bazel-genfiles` is just a matter of setting it up on the helm rules that we created. I particularly like to use `bazel-genfiles` only for transient generated stuff.

For parsing the output, I'll explore more the build event protocol.

.github/workflows/publish/publish.sh
f0rmiga: Oh, you are right.

.github/workflows/publish/rename.sh
f0rmiga: How did this work on my fork... nobody will ever know.

Yes, this is almost git describe. Right now we are not using the git tag, but eventually, we will.

.github/workflows/publish/build.sh
f0rmiga: The JSON output is bugous. We would need to consume the protobuf output, then this gets a bit out of scope from this PR. I'll create a card for this.

PR #2720, opened 2019-09-04
Changes to RC pipeline

mook-as:
## Description

This makes a couple separate changes to RC releases:

- Always build RC builds, don't filter based on changes.  This needs to be coupled with disabling polling in Jenkins, so that we must manually trigger all builds.
- Upload bundles on RC & nightly after testing, for jsc#CAP-631
- Keep the `-rc1` suffix in the release artifacts, so that we can push them directly to the helm repo.  Unfortunately this means the output will _not_ match the final bits. :warning: please review this more and make sure this is what we want.

## Test plan

- Set `GIT_TAG` to an RC release manually
- `make bundle-dist`

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -113,14 +113,20 @@ void runTest(String testName) {
     """
 }
 
-String distSubDir() {
+String distSubDir(boolean tested) {
     switch (getBuildType()) {
         case BuildType.Master:
             return 'master/'
         case BuildType.ReleaseCandidate:
-            return 'rc/'
+            if (tested) {
+                return 'rc/'
+            }
+            return 'rc-untested/'
         case BuildType.Nightly:
-            return 'nightly/'
+            if (tested) {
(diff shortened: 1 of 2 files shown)


Jenkinsfile
```diff
@@ -113,14 +113,20 @@ void runTest(String testName) {
     """
 }
 
-String distSubDir() {
+String distSubDir(boolean tested) {
```
satadruroy: I'm sure I'm getting confused but does Groovy allow skipping specifying parameters in a method call when the parameter is not marked as optional?

https://github.com/SUSE/scf/blob/3cb8b3600a4e4f82e6f410ac7ed9ad1cb8b0f27c/Jenkinsfile#L579

mook-as: Good catch, fixed in 964821c4c2c8c602300cdfe42247e7225ea25781.

PR #2721, opened 2019-09-05
jsc#CAP-554: Allow default stemcell to be configured

mook-as:
## Description

This is a fixed-up version of #2705, this time hopefully without breaking the addons.

## Test plan

Everything should deploy correctly.


mook-as:
For some reason (updated operator?) the `scf-eirini` service is magically back again.

f0rmiga:
Let's merge this after https://github.com/SUSE/scf/pull/2728 so we can test with Diego as well.

PR #2726, opened 2019-09-05
Provide ingress controller for embedded UAA feature

jandubois:
I couldn't find any documentation about how ingress handles overlapping rules; I hope they will be processed in order, so have sorted the more specific host names first:

```
      - host: "*.uaa.{{ .Values.env.DOMAIN }}"
      - host: "uaa.{{ .Values.env.DOMAIN }}"
      - host: "*.{{ .Values.env.DOMAIN }}"
```

I found some references to "longest prefix first", but it is not clear how this interacts with wildcard notation.

  jsc#CAP-675

### DIFF
--- a/chart-parts/ingress.yaml
+++ b/chart-parts/ingress.yaml
@@ -62,9 +62,28 @@ spec:
   tls:
   - secretName: "ingress-tls"
     hosts:
+    {{- if .Values.enable.uaa }}
+    - "*.uaa.{{ .Values.env.DOMAIN }}"
+    {{- end }}
     - "*.{{ .Values.env.DOMAIN }}"
     - "{{ .Values.env.DOMAIN }}"
   rules:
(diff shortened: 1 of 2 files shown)


chart-parts/ingress.yaml
```diff
@@ -62,21 +62,40 @@ spec:
...
+          serviceName: "uaa-uaa"
+          servicePort: 2793
+  - host: "uaa.{{ .Values.env.DOMAIN }}"
+    http:
+      paths:
+      - path: "/"
+        backend:
+          serviceName: "uaa-uaa"
+          servicePort: 2793
+    {{- end }}
```
f0rmiga: Could you indent this to the same level as the opening `if`?

jandubois:
@troytop says: "it appears to work", so I guess we are ready to merge then 😄

PR #2729, opened 2019-09-06
v3 - Smoke tests

f0rmiga:
## Description

WIP

## Test plan

WIP


testing/README.md
```diff
@@ -0,0 +1,9 @@
+# Testing
+
+## Smoke tests
+
+After the SCF deployment completed, run:
+
+```sh
+bazel run //testing/smoke_tests
+```
```
andreas-kupries: Add section explaining the background and how it has been re-implemented with cf-operator (pre-render, init containers, shared volumes, order of things) ... Because this is quite complex even while the implementation looks simple.

andreas-kupries: Beyond that a test shows that the smoke tests are properly triggered, and that they are able to invoke cf cli just fine. That this then fails with `lookup api.scf.suse.dev on 10.96.0.10:53: no such host` is not issue for this PR, which is about getting it run at all. The overall ticket will need more work tough, to make the test pass.

f0rmiga: The tests pass. It won't pass on KinD because there is no way of getting the smoke test to communicate to the go router. The ingress work was required to get access through the `api.<system_domain>`. It will work nicely on the Minikube approach and any public cluster.

andreas-kupries: So noted. I will re-trial using minikube.

f0rmiga: Added the comment with explanation in https://github.com/SUSE/scf/pull/2729/commits/1185a03a1522bee7cac3ed17af75788291fa38b5.

andreas-kupries: Thank you. Looks good enough.

PR #2734, opened 2019-09-09
v3 - Bumped Kubernetes version to 1.14.6

f0rmiga:
## Test plan

`kubectl` should report version 1.14.6.

```
bazel run @kubectl//:kubectl -- version
```

### DIFF
--- a/def.bzl
+++ b/def.bzl
@@ -21,19 +21,19 @@ project = struct(
         ],
     ),
     kubernetes = struct(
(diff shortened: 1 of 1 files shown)


andreas-kupries:
How do I test / verify that I have the correct version ?

Note: `kubectl version` reports 1.15 (i.e. beyond the 1.14.6 we should be bumped to).

PR #2740, opened 2019-09-10
v3 - Added temporary cf_cli.sh for creating cf-cli pod

f0rmiga:
## Description

This temporary script helps with accessing the CF cluster from within the k8s cluster.

## Test plan

1. Deploy SCF v3.
2. Execute the `cf_cli.sh` script.
3. Check that the `cf-terminal` pod was created and configured the cf-cli.
4. Get a shell session to the `cf-terminal` pod and perform cf-cli operations.

### DIFF
--- a/dev/cf_cli/cf_cli.sh
+++ b/dev/cf_cli/cf_cli.sh
@@ -1,0 +1,50 @@
+#!/bin/bash
+
+set -o errexit -o nounset
+
+namespace="scf"
+deployment_name="scf"
+pod_name="cf-terminal"
+api_ip=""
+echo "Waiting for endpoint..."
+while true; do
+  api_ip=$(kubectl describe endpoints -n "${namespace}" "${deployment_name}-router" | awk 'match($0, /  Addresses:[ ]+(.*)/, ip) { print ip[1] }')
+  if [ -n "${api_ip}" ]; then break; fi
+  sleep 3
+done
+admin_password=$(kubectl get secret --namespace "${namespace}" "${deployment_name}.var-cf-admin-password" -o jsonpath='{.data.password}' | base64 --decode)
+system_domain=$(kubectl get secret --namespace "${namespace}" "${deployment_name}.var-system-domain" -o jsonpath='{.data.value}' | base64 --decode)
+
+kubectl delete pod --namespace "${namespace}" "${pod_name}" || true
+kubectl create --namespace "${namespace}" -f - <<EOF
+apiVersion: v1
+kind: Pod
+metadata:
+  name: "${pod_name}"
+spec:
+  hostAliases:
+  - ip: "${api_ip}"
+    hostnames:
+    - "app1.${system_domain}"
+    - "app2.${system_domain}"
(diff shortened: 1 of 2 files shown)


dev/cf_cli/cf_cli.sh
```diff
@@ -0,0 +1,50 @@
+#!/bin/bash
+
+set -o errexit -o nounset
+
+namespace="scf"
+deployment_name="scf"
+pod_name="cf-terminal"
+api_ip=""
+echo "Waiting for endpoint..."
+while true; do
```
andreas-kupries: WIBNI (extension wish): auto-check for active ingress and use it if present. Alternatively, script option.

dev/cf_cli/cf_cli.sh
```diff
@@ -0,0 +1,50 @@
...
+  api_ip=$(kubectl describe endpoints -n "${namespace}" "${deployment_name}-router" | awk 'match($0, /  Addresses:[ ]+(.*)/, ip) { print ip[1] }')
+  if [ -n "${api_ip}" ]; then break; fi
+  sleep 3
```
andreas-kupries: If it has to sleep and iterate, maybe feedback that it is doing so ?

dev/cf_cli/cf_cli.sh
```diff
@@ -0,0 +1,50 @@
...
+done
```
andreas-kupries: Does it make sense to show the chosen IP ? (Maybe also endpoint or service, to know if router direct, or ingress is used).

dev/cf_cli/cf_cli.sh
```diff
@@ -0,0 +1,50 @@
...
+      set -o errexit
+      cf api --skip-ssl-validation "api.${system_domain}"
+      cf login -u admin -p "${admin_password}"
+      cf create-org demo
+      cf target -o demo
+      cf create-space demo
+      cf target -s demo
+      cf enable-feature-flag diego_docker
+      sleep 3600000
+EOF
```
andreas-kupries: Hm. Wait for pod to be healthy, then automatically ssh in ?
And on exit, auto-destroy the pod again ? (Controlled by option ?)

dev/cf_cli/cf_cli.sh
f0rmiga: My first script had this. It can become very verbose. I'd prefer not having any feedback.

dev/cf_cli/cf_cli.sh
f0rmiga: For a temporary script, this is too much.

dev/cf_cli/cf_cli.sh
f0rmiga: The purpose of this script isn't to talk to ingress, but to allow us to test other aspects of the CF cluster.

dev/cf_cli/cf_cli.sh
f0rmiga: Good suggestion.

PR #2741, opened 2019-09-10
Support external database server for CCDB &c.

mook-as:
## Description

This adds the option to use an external database server for the CF-internal databases (CCDB, CF-USB, etc., and UAA when it is embedded in the SCF deployment).  This lets administrators handle their own database backups if necessary.

A new job is used to initialize the databases, using the same configuration that the mysql-release uses in mariadb_ctrl.  The user must supply the external database host, port, and credentials.

## Test plan

```bash
$ helm install stable/mariadb --name external-db --namespace external-db --set volumePermissions.enabled=true

$ kubectl wait --namespace external-db --for=condition=ready pod/external-db-mariadb-master-0

$ export EMBEDDED_UAA=true

$ make/run --set env.DB_EXTERNAL_HOST=external-db-mariadb.external-db.svc.cluster.local --set env.DB_EXTERNAL_PORT=3306 --set env.DB_EXTERNAL_USER=root \
    --set secrets.DB_EXTERNAL_PASSWORD="$(kubectl get secret -n external-db external-db-mariadb -o jsonpath='{.data.mariadb-root-password}' | base64 --decode)"
```

## Additional information

- This does not yet support non-embedded UAA.
- This only support MySQL / MariaDB.


container-host-files/etc/scf/config/role-manifest.yml
```diff
@@ -2096,9 +2096,23 @@ instance_groups:
           "uaa.((KUBERNETES_NAMESPACE)).svc",
           "uaa.((KUBERNETES_NAMESPACE)).svc.((KUBERNETES_CLUSTER_DOMAIN))"
         ]
+      properties.uaadb.address: >
+        ((#DB_EXTERNAL_HOST))
+          ((DB_EXTERNAL_HOST))
+        ((/DB_EXTERNAL_HOST))
+        ((^DB_EXTERNAL_HOST))
+          mysql-proxy-set.((KUBERNETES_NAMESPACE)).svc.((KUBERNETES_CLUSTER_DOMAIN))
+        ((/DB_EXTERNAL_HOST))
```
andreas-kupries: I see this construction in a number of places (Nine, actually), this here the first.
Would it make sense to add an `&anchor` here and have all the other places just `*anchor` reference it ?

andreas-kupries: Thanks.

jandubois:
> That is something we may be able to do better, i.e. by not running the pods for the internal database(s).

I don't see an easy way to do this automatically, but we could at least add a `default_feature: mysql` property to the database pods, so you can disable them via `--set enable.mysql=false` in addition to specifying the external database settings.

mook-as:
Thinking about this a bit, we should probably put the DB host / port combo behind a service endpoint so that it's easier to migrate around if it gets removed; that doesn't _really_ help given that the auth credentials will probably need to change too, though.

andreas-kupries:
This PR should be moot now, via https://github.com/SUSE/scf/pull/2832

mook-as:
Superseded by #2832

PR #2748, opened 2019-09-11
check_swapaccounting: fix condition

mook-as:
## Description

We were testing for string non-emptiness (which was always true) rather than string contents, which means that we were not correctly disabling diego-cell when swap is available but swap accounting was not on.

Fixes #2684

## Test plan

- Set up vagrant with swap on, but swapaccounting off
- Try to deploy, it should fail
- Turn off the swap, and try again
- It should pass.


mook-as:
Ensuring this works on vagrant:

- Edit `/etc/default/grub` to remove `swapaccount=1`, and build with `grub2-mkconfig -o /boot/grub2/grub.cfg`
- `vagrant reload`
- Build (but don't run)
- `cd ~root && dd if=/dev/zero of=swap bs=1M count=1 && chmod 0600 swap && mkswap swap && swapon -f $PWD/swap` (as root)
- `free -h` to check that we have swap
- Run, and check that things fail.

PR #2754, opened 2019-09-13
Consume eirinifs from the eirinifs-sle15-release

jimmykarily:

### DIFF
--- a/container-host-files/etc/scf/config/role-manifest.yml
+++ b/container-host-files/etc/scf/config/role-manifest.yml
@@ -151,6 +151,11 @@ releases:
   url: "https://suse-final-releases.s3.amazonaws.com/eirini-release-0.0.17.tgz"
   version: "0.0.17"
   sha1: "e969742c39743302836190b3be033ddf22beb411"
+- name: "eirinifs-sle15-release"
+  url: "https://suse-final-releases.s3.amazonaws.com/eirinifs-sle15-release-0.1.1.tgz"
+  version: "0.1.1"
+  sha1: "1ccb423a9e9463d334e55680a21db423d6ac7314"
+
 - name: uaa
   version: '72.0'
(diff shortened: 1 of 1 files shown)


jimmykarily:
I pushed with `cf push -s sle15` and `cf push -s cflinuxfs3`. Both work and result in an app running inside a sle15 container (that's the eirinifs introduced in this PR). The application I used is dizzylizard but I'm not sure if every application would work without changing the staging images to SLE. We have a pipeline that builds those staging images on a SLE base but we still miss the automation to publish them to the production registry upon release. My suggestion is to add the staging images in this PR before merging to avoid breaking deployment on Eirini for some apps.

jkbschmid:
Hi @jimmykarily, I am also trying to get eirini+cf-operator to work right now. 
What's the idea behind the sle15-release? Is there an issue or something similar where I could get some context?

jimmykarily:
Hi @jkbschmid, sorry for the late reply. The email notification somehow slipped between the cracks. The eirinifs is the equivalent of the stack in the diego world. SUSE ships CAP with all images based on SLE (SUSE Linux Enterprise). The [eirini-bosh-release](https://github.com/cloudfoundry-community/eirini-bosh-release/blob/master/config/blobs.yml#L1) currently includes a version of eirinifs which is based on Ubuntu (cflinuxfs3). To let anyone use whatever eirinifs they want based on whatever os they want, the plan is to extract the eirinifs from the eirini-bosh-release into a separate release and consume it from there (for SLE we already have a separate bosh release). Not sure how that will work with the operator but I guess everything possible with the role-manifest is also possible with the operator. Maybe @viovanov can say more about that.

jimmykarily:
Closing in favor of: https://github.com/SUSE/scf/pull/2906/files

viovanov:
@jimmykarily for the operator, we can create an image for any BOSH release - then alter the deployment using ops files

PR #2787, opened 2019-09-17
Initial commit of acceptance tests

mook-as:
## Description

Add initial support for CATS; it runs as an ejob, same as smoke tests.

Requires BOSH release [mook-as/cf-acceptance-tests-release].

[mook-as/cf-acceptance-tests-release]: https://github.com/mook-as/cf-acceptance-tests-release/

Does not pass yet; this is just for the initial support so we can start looking at the failures.

## Test plan

- Deploy SCF
- Wait for everything to go green
- Edit the `acceptance-tests` ejob to have a `strategy` of `now` to trigger the job

## Notes
- We need a pipeline for building the BOSH release later (and a better home for the release)
- See also mook-as/cf-acceptance-tests-release#1

### DIFF
--- a/deploy/helm/scf/assets/operations/instance_groups/acceptance-tests.yaml
+++ b/deploy/helm/scf/assets/operations/instance_groups/acceptance-tests.yaml
@@ -1,0 +1,34 @@
+- path: /releases/-
+  type: replace
+  value:
+    name: cf-acceptance-tests
+    url: {{ .Values.releases.defaults.url | quote }}
+    version: 0.0.1
+- path: /instance_groups/-
+  type: replace
+  value:
+    azs:
+    - z1
+    instances: 1
+    jobs:
+    - name: acceptance-tests
+      properties:
+        acceptance_tests:
+          admin_password: ((cf_admin_password))
+          admin_user: admin
+          api: https://api.((system_domain))
+          apps_domain: ((system_domain))
+          include: >-
+            {{ .Values.releases.cf_acceptance_tests.properties.acceptance_tests.include }}
+          skip_ssl_validation: true
+        bpm:
+          enabled: true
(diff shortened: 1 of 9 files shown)


andreas-kupries:
@mook-as See `testing/smoke_tests/BUILD.bazel` for the code to trigger smoke tests. The cats trigger should be analogous.

andreas-kupries:
@mook-as 
> Edit the `acceptance-tests` ejob to have a strategy of now to trigger the job

The actual and full name of the ejob is `scf-acceptance-tests`, and that has to be used for triggering. IOW namespace from the operator is part of the name.

Here is my `testing/acceptance_tests/BUILD.bazel` file:

```
package(default_visibility = ["//visibility:public"])

load("//rules/kubectl:def.bzl", kubectl_patch = "patch")
load("//:def.bzl", "project")

kubectl_patch(
    name = "acceptance_tests",
    namespace = project.namespace,
    resource_type = "ejob",
    resource_name = "scf-acceptance-tests",
    patch_type = "merge",
    patch_file = ":trigger.yaml",
)
```

For the referenced `trigger.yaml` I use a copy of the file for the smoke tests. This should possibly be moved out of the directory and consolidated into a single file.

Running the tests is then done with `bazel run //testing/acceptance_tests`.

f0rmiga:
@andreas-kupries 

> For the referenced trigger.yaml I use a copy of the file for the smoke tests. This should possibly be moved out of the directory and consolidated into a single file.

I don't agree that we should use a single file to trigger. This YAML file can be used to patch other stuff in the future if we need to.

andreas-kupries:
Ok.

mook-as:
Rebased with new image, now everything should be review-ready-ish.

PR #2790, opened 2019-09-18
Use kube csr feature for opi-registry in eirini

jandubois:
The old mechanism of using a docker cert copying daemon set does not work with non-docker based container runtimes. It also has problems when the host filesystem is not writable by the daemon set.

The new mechanism uses the kube certificate service to sign a cert for the registry. This still requires the kube CA to be installed such that the container runtimes on each worker node will trust it.

This commit uses a nodePort service on `127.0.0.1` for the registry:

* Can't use `registry.$DOMAIN` because it uses the wildcard cert from the gorouter. Also we don't want to expose the registry publicly.

* Can't use `bits-bits-service.$NS.svc.cluster.local` because the name doesn't resolve outside the cluster (container runtime cannot pull from there).

* Can't use `localhost` because it doesn't work (for unknown reasons).

All cluster-roles are moved out of the rolemanifest into `chart-parts/` so that only the ones used in a specific configuration are defined and bound to the service accounts.

[jsc#CAP-587]

### DIFF
--- a/chart-parts/cluster-roles.yaml
+++ b/chart-parts/cluster-roles.yaml
@@ -1,0 +1,136 @@
+{{- if and (eq (printf "%s" .Values.kube.auth) "rbac") (.Capabilities.APIVersions.Has "rbac.authorization.k8s.io/v1") }}
+
+{{- if .Values.enable.eirini }}
+
+# eirini-webhook cluster role for eirini service account.
+# Used to implement eirinix extensions.
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: {{ template "fissile.SanitizeName" (printf "%s-eirini-webhook" .Release.Namespace) }}
+  labels:
+    {{- template "scf.role-labels" "eirini-webhook" }}
+    {{- template "scf.labels" . }}
+rules:
+- apiGroups: ['admissionregistration.k8s.io']
+  resources: ['mutatingwebhookconfigurations']
+  verbs:     ['*']
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: {{ template "fissile.SanitizeName" (printf "%s-eirini-webhook" .Release.Namespace) }}
+  labels:
+    {{- template "scf.role-labels" "eirini-webhook" }}
+    {{- template "scf.labels" . }}
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: {{ template "fissile.SanitizeName" (printf "%s-eirini-webhook" .Release.Namespace) }}
+subjects:
+- kind: ServiceAccount
+  name: eirini
+  namespace: {{.Release.Namespace}}
+
+# eirini-csr cluster role for secret-generator service account.
+# Used to create certificate signing requests for the opi-registry certificate.
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: {{ template "fissile.SanitizeName" (printf "%s-eirini-csr" .Release.Namespace) }}
+  labels:
+    {{- template "scf.role-labels" "eirini-csr" }}
+    {{- template "scf.labels" . }}
+rules:
+- apiGroups: [certificates.k8s.io]
+  resources: [certificatesigningrequests]
+  verbs:     [create, get, delete]
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: {{ template "fissile.SanitizeName" (printf "%s-eirini-csr" .Release.Namespace) }}
+  labels:
+    {{- template "scf.role-labels" "eirini-csr" }}
+    {{- template "scf.labels" . }}
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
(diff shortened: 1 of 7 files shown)


chart-parts/cluster-roles.yaml
```diff
@@ -0,0 +1,136 @@
...
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: {{ template "fissile.SanitizeName" (printf "%s-eirini-csr-approver" .Release.Namespace) }}
+subjects:
+- kind: ServiceAccount
+  name: secret-generator
+  namespace: {{ .Release.Namespace }}
+{{- end }}
+
+{{- else }}{{/* if .Values.enable.eirini */}}
```
mook-as: I wonder if it would make things easier to read if we had separate Eirini and Diego cluster role files. (This does not block merging, of course)

container-host-files/etc/scf/config/role-manifest.yml
```diff
@@ -279,7 +281,8 @@ instance_groups:
           internal: 443
   configuration:
     templates:
-      properties.route_registrar.routes: '[{"name":"bits-service","tls_port":443,"uris":["bits.((DOMAIN))", "registry.((DOMAIN))"],"registration_interval":"20s","server_cert_domain_san":"bits","tags":{"component":"bits-service"}}]'
+      properties.bits-service.registry_endpoint: "https://127.0.0.1"
+      properties.route_registrar.routes: '[{"name":"bits-service","tls_port":443,"uris":["bits.((DOMAIN))"],"registration_interval":"20s","server_cert_domain_san":"bits","tags":{"component":"bits-service"}}]'
```
mook-as: We're removing `registry.((DOMAIN))` here right? Do we no longer need it?

container-host-files/etc/scf/config/role-manifest.yml
```diff
@@ -3057,7 +3049,7 @@ configuration:
     # TODO: this needs to be fixed upstream - it should be an array
     properties.opi.nats_ip: "nats-0.nats-set.((KUBERNETES_NAMESPACE)).svc.((KUBERNETES_CLUSTER_DOMAIN))"
     properties.opi.nats_password: "((NATS_PASSWORD))"
-    properties.opi.registry_address: "registry.((DOMAIN))"
+    properties.opi.registry_address: "127.0.0.1:((OPI_REGISTRY_NODEPORT))"
```
mook-as: Ah because we don't use bits here anymore?

container-host-files/etc/scf/config/scripts/set-opi-registry-port.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/sh
+
+# Locate the kubectl binary.
+kubectl="/var/vcap/packages/kubectl/bin/kubectl"
```
mook-as: Huh, wouldn't munging `$PATH` be easier?

chart-parts/cluster-roles.yaml
```diff
@@ -0,0 +1,136 @@
...
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: {{ template "fissile.SanitizeName" (printf "%s-eirini-csr" .Release.Namespace) }}
+subjects:
+- kind: ServiceAccount
+  name: secret-generator
+  namespace: {{ .Release.Namespace }}
+
+{{- if eq (toString .Values.env.KUBE_CSR_AUTO_APPROVAL) "true" }}
```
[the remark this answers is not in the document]
jandubois: > Oh, is it because it's coming in as a string (so `"false"` would be truthy)? Bleh.

Yes, it is. 😞 It is rather errorprone, but I think I'm going to blame `helm` for it.

chart-parts/cluster-roles.yaml
jandubois: We could; I'm kind of indifferent, as `diego` is kind of `not-eirini`, but I can make them separate files, if you prefer.

container-host-files/etc/scf/config/role-manifest.yml
jandubois: Yes, we don't want to expose the registry externally. At least now it has TLS support; when I started this commit it only supported HTTP.

The only reason to support the external endpoint would be as some mechanism to "export" apps as docker images from the cluster. I think that use case is misguided, as you cannot properly show what went into building this image.

container-host-files/etc/scf/config/role-manifest.yml
jandubois: Yes

chart-parts/cluster-roles.yaml
mook-as: Meh, I'm not too worried about it. Hopefully we'll just drop diego at some point :p

container-host-files/etc/scf/config/scripts/set-opi-registry-port.sh
jandubois: I just copied the code from https://github.com/SUSE/scf/blob/develop/container-host-files/etc/scf/config/scripts/configure-az.sh and https://github.com/SUSE/scf/blob/develop/container-host-files/etc/scf/config/scripts/configure-pz.sh to make it look familiar. We could also just hardcode the full path when we call `kubectl` because we only use this once.

What do you prefer?

mook-as: 🤷‍♂

PR #2793, opened 2019-09-18
Change DEFAULT_STACK to cflinuxfs3 when Eirini is enabled

jandubois:
because Eirini currently doesn't work with any of the other bundled stacks, so better make the one that works the default

[jsc#CAP-721]

## Test plan

Deploy with Diego and check that `cf push` uses the `sle15` stack.
Deploy with Eirini and check that `cf push` uses the `cflinuxfs3` stack.


andreas-kupries:
@jandubois What was your test app ?

jandubois:
> @jandubois What was your test app ?

https://github.com/scf-samples/12factor

andreas-kupries:
Thanks

PR #2807, opened 2019-09-20
v3 - Shellcheck managed by Bazel

f0rmiga:
## Description

Now that shellcheck publishes OS X binaries (not just Linux and Windows ones), it's easy to manage all the shellcheck binaries using Bazel.

## Test plan

Run `//dev/linters/shellcheck.sh` from a Linux, a Mac and a Windows machines.


andreas-kupries:
Actually run `dev/linters/shellcheck.sh` (without leading `//`). It is a script path, not a bazel target.
That said, looks ok.

PR #2809, opened 2019-09-20
Add liveness & readiness probes

mook-as:
## Description

This series of commits adds liveness and readiness probes for most jobs (I didn't touch the addons yet).

In general, the liveness probes are mostly checking that the correct process is running.  The readiness probes check for a health endpoint where available; some of them check `ppro` or `expvar` output, which isn't exactly correct but at least checks that something responds.  Unfortunately most processes do not have truly useful health check endpoints.

Notably, UAA & CCNG do not have liveness probes set, as they both have a long bootstrap step (cert import and db migration / seeding, respectively) that is run under a different executable.  We may be able to get away with moving those into pre-start steps, but that would have to be separate PRs. (And IIRC, we did do so for CCNG, but it wasn't an often-used configuration by upstream.)

## Test plan

- Deploy SCF
- Manually check that each pod has ~4 processes without liveness / readiness probes (because those are from the loggregator agent addon).

### DIFF
--- a/deploy/helm/scf/assets/operations/instance_groups/adapter.yaml
+++ b/deploy/helm/scf/assets/operations/instance_groups/adapter.yaml
@@ -18,5 +18,9 @@
       healthcheck:
         adapter:
           readiness:
+            # The healthcheck port is bound to localhost only
             exec:
-              command: ['curl', '--fail', '--head', 'http://127.0.0.1:8080/health']
+              command: [curl, --fail, --head, --silent, http://localhost:8080/health]
+          liveness:
+            exec:
+              command: [pidof, adapter]
--- a/deploy/helm/scf/assets/operations/instance_groups/adapter.yaml
+++ b/deploy/helm/scf/assets/operations/instance_groups/adapter.yaml
@@ -21,6 +21,3 @@
             # The healthcheck port is bound to localhost only
             exec:
               command: [curl, --fail, --head, --silent, http://localhost:8080/health]
-          liveness:
-            exec:
-              command: [pidof, adapter]
--- a/deploy/helm/scf/assets/operations/instance_groups/api.yaml
+++ b/deploy/helm/scf/assets/operations/instance_groups/api.yaml
@@ -193,20 +193,16 @@
                 printf "GET /healthz HTTP/1.0\r\n\r\n" |
                 nc -q 2 -U /var/vcap/data/cloud_controller_ng/cloud_controller.sock |
                 grep --silent "200 OK"
-            httpGet:
-              path: /healthz
-              port: 9022
           # We don't want a liveness probe here as we do migration here, and we
-          # do not want to interrupt that.
+          # do not want to interrupt that.  We may want to consider using a
+          # startupProbe in the future (once that feature stabilizes).
           liveness: ~
         local_worker_1:
           readiness: &cc_local_worker_readiness
             exec:
               command: [/usr/bin/pgrep, --full, cc_api_worker]
-          liveness: *cc_local_worker_readiness
         local_worker_2:
           readiness: *cc_local_worker_readiness
-          liveness: *cc_local_worker_readiness
         nginx:
           readiness:
             httpGet:
@@ -239,9 +235,6 @@
             # routing-api does not expose a health check endpoint
             tcpSocket:
               port: 3000
-          liveness:
-            exec:
-              command: [pidof, routing-api]
 
 # Add quarks properties for cc_uploader.
 - type: replace
@@ -262,9 +255,6 @@
             # socket.
             tcpSocket:
               port: 9091
-          liveness:
-            exec:
-              command: [pidof, cc-uploader]
 
 # Add quarks properties for file_server.
 - type: replace
@@ -281,23 +271,14 @@
             httpGet:
               path: /v1/static/file_server/bin/file-server
               port: *file-server-port
-          liveness:
-            exec:
-              command: [pidof, file-server]
 
 # Add quarks properties for statsd_injector.
 - type: replace
-  path: /instance_groups/name=api/jobs/name=statsd_injector/properties/quarks?/run/healthcheck/statsd_injector
+  path: /instance_groups/name=api/jobs/name=statsd_injector/properties/quarks?/run/healthcheck/statsd_injector/readiness/exec/command
   value:
-    readiness:
-      exec:
-        command:
-        - /bin/sh
-        - -c
-        - ss -nlu src localhost:8125 | grep :8125
-    liveness:
-      exec:
-        command: [pidof, statsd-injector]
+  - /bin/sh
+  - -c
+  - ss -nlu src localhost:8125 | grep :8125
 
 # Add quarks properties for policy-server.
 - type: replace
@@ -313,9 +294,6 @@
           readiness:
             httpGet:
               port: 4002
-          liveness:
-            exec:
-              command: [pidof, policy-server]
     post_start:
       condition:
         exec:
@@ -332,9 +310,6 @@
           readiness: &policy_server_internal_readiness
             httpGet:
               port: 31946
-          liveness:
-            exec:
-              command: [pidof, policy-server-internal]
     post_start:
       condition:
         exec:
@@ -344,22 +319,13 @@
 - type: replace
   path: /instance_groups/name=api/jobs/name=route_registrar/properties/quarks?/run/healthcheck/route_registrar
   value:
-    readiness: &route_registrar_readiness
+    readiness: ~
       # The route registrar doesn't expose anything to indicate if the
-      # routes are healthy; we can only detect that it's still running.
-      exec:
-        command: [pidof, route-registrar]
-    liveness: *route_registrar_readiness
+      # routes are healthy
 
 - type: replace
-  path: /instance_groups/name=api/jobs/name=loggr-udp-forwarder/properties/quarks?/run/healthcheck/loggr-udp-forwarder
-  value:
-    readiness:
-      exec:
-        command: [sh, -c, 'ss -nlu sport = 3457 | grep :3457']
-    liveness:
-      exec:
-        command: [pidof, udp-forwarder]
+  path: /instance_groups/name=api/jobs/name=loggr-udp-forwarder/properties/quarks?/run/healthcheck/loggr-udp-forwarder/readiness/exec/command
+  value: [sh, -c, ss -nlu sport = 3457 | grep :3457]
 
 {{- $root := . -}}
 {{- range $path, $bytes := .Files.Glob "assets/operations/pre_render_scripts/api_*" }}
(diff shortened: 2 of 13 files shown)


deploy/helm/scf/assets/operations/instance_groups/api.yaml
```diff
@@ -170,12 +170,56 @@
       internal: 9024
     run:
       healthcheck:
+        ccng_monit_http_healthcheck:
+          readiness:
+            # This job exists just to healthcheck cloud_controller_ng; we're
+            # already doing that separately.
```
mook-as: It might be useful to see if we can just drop this job completely.

deploy/helm/scf/assets/operations/instance_groups/api.yaml
```diff
@@ -170,12 +170,56 @@
...
+          #  exec:
+          #    command:
+          #    - /usr/bin/pgrep
+          #    - --full
+          #    - /var/vcap/packages/cloud_controller_ng/cloud_controller_ng/bin/cloud_controller
+        local_worker_1:
+          readiness: &cc_local_worker_readiness
             exec:
-              command: [curl, --fail, --head, --silent, http://127.0.0.1:9022/healthz]
+              command: [/usr/bin/pgrep, --full, cc_api_worker]
```
mook-as: This isn't awesome, but there isn't a good health endpoint for the worker (nor the separate cc-worker job).

deploy/helm/scf/assets/operations/instance_groups/api.yaml
```diff
@@ -210,18 +275,26 @@
...
+            exec:
+              command: [pidof, file-server]
 
 # Add quarks properties for statsd_injector.
 - type: replace
-  path: /instance_groups/name=api/jobs/name=statsd_injector/properties/quarks?
+  path: /instance_groups/name=api/jobs/name=statsd_injector/properties/quarks?/run/healthcheck/statsd_injector
   value:
-    ports:
-    # TODO: Can we remove this port?
```
mook-as: Yes, we could. It never worked anyway, since statsd-injector was listening on **UDP** port 8125.

deploy/helm/scf/assets/operations/instance_groups/api.yaml
```diff
@@ -170,12 +170,56 @@
...
+              command:
+              - /usr/bin/pgrep
+              - --full
+              - --exact
+              - bash /var/vcap/jobs/cloud_controller_ng/bin/ccng_monit_http_healthcheck
         cloud_controller_ng:
           readiness: &cloud_controller_ng_readiness
+            httpGet:
+              path: /healthz
+              port: 9022
```
mook-as: This is _technically_ not correct; it's actually checking at the endpoint behind nginx.  However, the version of curl in our stemcell is too old to support talking to unix sockets.  We should be able to move to `curl --unix-socket` once we bump our stemcell to Leap/SLE 15.

deploy/helm/scf/assets/operations/instance_groups/diego-api.yaml
```diff
@@ -18,6 +18,9 @@
 - type: replace
   path: /instance_groups/name=diego-api/jobs/name=bbs/properties/diego/bbs/advertisement_base_hostname?
   value: {{ .Values.deployment_name }}-diego-api
+- type: replace
+  path: /instance_groups/name=diego-api/jobs/name=bbs/properties/diego/bbs/health_addr?
+  value: 0.0.0.0:8890
```
mook-as: We _could_ drop this, and do a local curl instead; but I didn't see the harm in exposing the health check endpoint internally.

deploy/helm/scf/assets/operations/instance_groups/diego-cell.yaml
```diff
@@ -81,9 +89,6 @@
   path: /instance_groups/name=diego-cell/jobs/name=rep/properties/quarks?
   value:
     ports:
-    - name: rep
-      protocol: TCP
-      internal: 1800
```
mook-as: This wasn't necessary (we were only listening to port 1800 on loopback).

deploy/helm/scf/assets/operations/instance_groups/log-api.yaml
```diff
@@ -39,12 +39,56 @@
...
+          readiness:
+            exec:
+              command: [curl, --fail, --silent, --head, http://localhost:14826/health]
+          liveness:
+            exec:
+              command: [pidof, rlp]
+
+- type: replace
+  path: /instance_groups/name=log-api/jobs/name=reverse_log_proxy_gateway/properties/quarks?/bpm/processes/name=reverse_log_proxy_gateway/env/PPROF_PORT?
+  value: "33045"
```
mook-as: Needed to set _something_ here, otherwise it picked a random port.

deploy/helm/scf/assets/operations/instance_groups/doppler.yaml
```diff
@@ -35,16 +35,61 @@
...
+            exec:
+              command: [curl, --fail, --head, --silent, http://localhost:14825/health]
+          liveness:
+            exec:
+              command: [pidof, doppler]
 
 # Add quarks properties for log-cache.
+- type: replace
+  path: /instance_groups/name=doppler/jobs/name=log-cache/properties/health_addr?
+  value: "::6060"
```
f0rmiga: Why use IPv6 notation here?

deploy/helm/scf/assets/operations/instance_groups/adapter.yaml
```diff
@@ -18,5 +18,9 @@
       healthcheck:
         adapter:
           readiness:
+            # The healthcheck port is bound to localhost only
```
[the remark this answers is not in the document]
mook-as: Well, I added it because otherwise (had it be bound to `0.0.0.0` or `::`) I'd have used a `httpGet` verb instead :)

deploy/helm/scf/assets/operations/instance_groups/api.yaml
mook-as: Created https://github.com/SUSE/scf/issues/2825

deploy/helm/scf/assets/operations/instance_groups/api.yaml
mook-as: I actually tried that first; it didn't work for some reason (I kept getting a `bad request` response, which would have failed to healthcheck).  I can try it again (I don't have a v3 deployment ready at the moment).

deploy/helm/scf/assets/operations/instance_groups/doppler.yaml
mook-as: Mainly because I'm lazy and wanted to type less :p I guess now that I have to explain it I've actually had to type more…

deploy/helm/scf/assets/operations/instance_groups/api.yaml
f0rmiga: If you "Allow edits from maintainers", I can take this over and push commits to your fork, so I don't interrupt your current work.

deploy/helm/scf/assets/operations/instance_groups/adapter.yaml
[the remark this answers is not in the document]
f0rmiga: I see.

deploy/helm/scf/assets/operations/instance_groups/doppler.yaml
f0rmiga: I was just curious if there was another reason that I was not seeing. :P
Sorry for making you type more. haha

deploy/helm/scf/assets/operations/instance_groups/api.yaml
```diff
@@ -185,6 +229,16 @@
...
+    run:
+      healthcheck:
+        routing-api:
+          readiness:
+            # routing-api does not expose a health check endpoint
+            tcpSocket:
+              port: 3000
+          liveness:
+            exec:
+              command: [pidof, routing-api]
```
viovanov: What's the reasoning behind this?

deploy/helm/scf/assets/operations/instance_groups/api.yaml
```diff
@@ -234,31 +307,56 @@
...
+      condition:
+        exec:
+          # policy-server-internal doesn't support HTTP HEAD requests
+          command: [curl, --fail, --silent, http://localhost:31946/]
+
+- type: replace
+  path: /instance_groups/name=api/jobs/name=route_registrar/properties/quarks?/run/healthcheck/route_registrar
+  value:
+    readiness: &route_registrar_readiness
+      # The route registrar doesn't expose anything to indicate if the
```
viovanov: You don't have to do that. The container daemon does it for you.
If the process goes away, the container should go away.

viovanov:
I don't think any of the `pidof` probes are useful.
Also, as a rule I think liveness probes should only be used when needed, and there's a clear problem they're solving. They have a very serious side effect - the container is killed if the probe doesn't pass.

deploy/helm/scf/assets/operations/instance_groups/api.yaml
mook-as: Fixed in 215d8e8101fec5f68087915249cd475381a6289e — I needed to add `-q 2` to the `nc` command line to keep the connection open for a bit, in order for `nc` to read the response back.

(_Allow edits from maintainers_ was on when the PR was created)

deploy/helm/scf/assets/operations/instance_groups/api.yaml
```diff
@@ -170,12 +170,56 @@
...
+              - --exact
+              - bash /var/vcap/jobs/cloud_controller_ng/bin/ccng_monit_http_healthcheck
         cloud_controller_ng:
           readiness: &cloud_controller_ng_readiness
+            httpGet:
+              path: /healthz
+              port: 9022
+          # We don't want a liveness probe here as we do migration here, and we
+          # do not want to interrupt that.
+          #liveness:
```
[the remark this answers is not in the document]
mook-as: Changed in 215d8e8101fec5f68087915249cd475381a6289e

deploy/helm/scf/assets/operations/instance_groups/api.yaml
mook-as: Our process tree is
```
/usr/bin/dumb-init
\_ /var/vcap/all-releases/container-run/container-run
  \_ /var/vcap/packages/route_registrar/bin/route-registrar
```
(Or similar for the other things)

So we could potentially get stuck and never run the route-registrar process / not notice it exiting.  If we had `route-registrar` as pid 1, then the `pidof` checks would definitely be unnecessary.

I'm okay with just getting rid of the `pidof` checks completely, though; I feel like that's less crazy then just having the same liveness & readiness probes.

f0rmiga: If we do the `pidof` approach, we are saying that we don't trust the software we wrote (ok, I did) and has 100% test coverage. Unless we also don't trust dumb-init.

mook-as: Already got rid of `pidof` :) I blame the task for being specified as add liveness & readiness probes, and not _as appropriate_ :stuck_out_tongue: So I added a bunch inappropriately :grin:

viovanov:
@mook-as if you resolve conflicts and are able to run locally, I'm happy to merge this now

mook-as:
Rebased for merge conflicts. Still seems to run.

PR #2820, opened 2019-09-24
Gitlab CI

f0rmiga:
## Description

Moving from GitHub Actions to Gitlab CI. This initial Gitlab work should have the same behaviour as the GitHub Actions workflow + shellcheck linting step.


.gitlab-ci.yml
```diff
@@ -0,0 +1,47 @@
...
+lint:
+  stage: lint
+  image: *bazel_docker_image
+  script:
+  - ./dev/linters/shellcheck.sh
+
+build:chart:
+  stage: build
+  image: *bazel_docker_image
+  script:
```
andreas-kupries: I prefer to have scripts like these as a proper script file.
See `publish:chart:` for another place where I would prefer that.
See `lint` for a place where this is done properly (IMHO).

.gitlab-ci.yml
```diff
@@ -0,0 +1,47 @@
...
+    expire_in: 1 week
+
+publish:chart:
+  stage: publish
+  only:
+    refs:
+    - *default_branch
+  image: minio/mc
+  script:
+  - mc config host add "s3" "https://s3.amazonaws.com" "${AWS_ACCESS_KEY}" "${AWS_SECRET_KEY}"
```
andreas-kupries: See above, about placing these things into script files.

.gitlab/deploy/runner/kubernetes/minikube/README.md
```diff
@@ -0,0 +1,4 @@
+# Deploy Gitlab runner on Minikube
+
+Run the `deploy` target to install the Gitlab runner using Helm. This target assumes Helm Tiller is
+already installed.
```
andreas-kupries: Where is the runner installed to ?
Based on the directory name I suspect that it is the local minikube, the same one used to normally deploy the operator to, right ?

andreas-kupries: Oh, forgot something, came to me looking over the main PR description ... Should possibly mention in some global README that this depends on having a local gitlab runner deployed. It feels to be that way, at least (what with bazel target and scripts added, to deploy such).

.gitlab-ci.yml
f0rmiga: You are right! I should've done this from the beginning. Thanks.

.gitlab/deploy/runner/kubernetes/minikube/README.md
f0rmiga: So, right now we are using shared runners on gitlab.com. I was initially experimenting with a runner running on a local Minikube and thought that it would be useful to have some scripts to help us deploying and using it. I realized now after the review that we don't need specific instructions to Minikube. I'll keep the, almost same, script but as a generic installer to k8s. Once we decide/need to run custom runners as privileged, then we already have the scripts. We can do it on whatever long-running cluster.

andreas-kupries: ok

.gitlab-ci.yml
andreas-kupries: Confirming it done, ty

PR #2834, opened 2019-09-27
Get rid of bits-service

jandubois:
We don't need it; we can just talk directly to the blobstore.

For Eirini we need the bits registry, but even Eirini's recipe up- and downloaders will talk to blobstore if bit-service is not enabled.

All this allows us to use just `127.0.0.1` as the common name for the bits ssl cert. This is important on EKS because the kube cert signer does not support subject alternate names, so the cert can
only have a single name, and it MUST be the registry address for it to be trusted by the container runtime.

[jsc#CAP-587]

### DIFF
--- a/container-host-files/etc/scf/config/opinions.yml
+++ b/container-host-files/etc/scf/config/opinions.yml
@@ -163,14 +163,12 @@ properties:
       directory_key: cc-packages
       webdav_config:
         username: blobstore_user
+    signing_users: []
   capi:
     tps:
       watcher:
         skip_consul_lock: true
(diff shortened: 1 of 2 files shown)


mudler:
For reference, Eirini tests are green with this PR.

Deployment: https://concourse.suse.de/teams/main/pipelines/eirini-ci/jobs/deploy-eirini-pr/builds/38 Smoke tests: https://concourse.suse.de/teams/main/pipelines/eirini-ci/jobs/smokes-eirini-pr/builds/25

PR #2838, opened 2019-09-27
v3 - Added mechanism for allowing user provided job properties

### LINKED ISSUE #2766: Mechanism to expose any BOSH property
As a user of SCF v3, I want to be able to configure any BOSH Job property.
A possible values.yaml:

```
properties:
  instance-group-name:
    job-name:
      some-property: some-value
```

f0rmiga:
## Description

Adding a mechanism that allows a user to provide job properties via Helm values.

Resolves #2766 

## Test plan

Set some properties and assert they took effect. E.g. add the following to `values.yaml`:

```yaml
properties:
  adapter:
    adapter:
      quarks:
        ports:
        - name: adapter
          protocol: TCP
          internal: 4443
        - name: another
          protocol: TCP
          internal: 7171
      scalablesyslog:
        adapter:
          logs:
            addr: scf-log-api:8082
```

### DIFF
--- a/deploy/helm/scf/values.yaml
+++ b/deploy/helm/scf/values.yaml
@@ -3,10 +3,13 @@ deployment_name: scf
 
 # Set or override job properties. The first level of the map is the instance group name. The second
 # level of the map is the job name. E.g.:
-#   properties:
-#     adapter:
-#       adapter:
-#         scalablesyslog.adapter.logs.addr: scf-log-api:8082
+#  properties:
+#    adapter:
+#      adapter:
+#        scalablesyslog:
+#          adapter:
+#            logs:
+#              addr: scf-log-api:8082
 properties: {}
 
 releases:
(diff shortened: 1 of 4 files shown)


deploy/helm/scf/values.yaml
```diff
@@ -1,6 +1,14 @@
 system_domain: ~
 deployment_name: scf
 
+# Set or override job properties. The first level of the map is the instance group name. The second
+# level of the map is the job name. E.g.:
+#   properties:
+#     adapter:
+#       adapter:
+#         scalablesyslog.adapter.logs.addr: scf-log-api:8082
```
viovanov: Can we make this a deep object instead of using the `.` notation?

f0rmiga: I'll see if it's feasible using the Sprig library that Helm exposes to us.

PR #2865, opened 2019-10-02
(Option to) use helm/stable NATS chart

### LINKED ISSUE #2723: Allow usage of the official NATS helm chart
There should be a feature flag that allows users to deploy the official NATS helm chart instead of the converted BOSH release.

mook-as:
## Description
This adds support to use the upstream helm chart for NATS, rather than using NATS based on the BOSH release.

## Motivation and Context
Fixes #2723

## How Has This Been Tested?
Brought up (on minikube/libvirt) both with `nats.external = true` and `= false` configurations, and checked that everything looks ready and smoke tests start.

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)

## Checklist:

## Additional Information:

:warning: The NATS subchart currently still gets deployed even when not used; not sure yet how I'd go about disabling that.

### DIFF
--- a/deploy/helm/scf/values.yaml
+++ b/deploy/helm/scf/values.yaml
@@ -31,6 +31,9 @@ features:
   # published to docker.io.
   suse_buildpacks: false
 
+  # Set this to true to disable the NATS release, and use the official helm chart
+  external-nats: false
+
 operations:
   # A list of configmap names that should be applied to the BOSH manifest.
   custom: []
--- a/deploy/helm/scf/values.yaml
+++ b/deploy/helm/scf/values.yaml
@@ -31,9 +31,6 @@ features:
   # published to docker.io.
   suse_buildpacks: false
 
-  # Set this to true to disable the NATS release, and use the official helm chart
-  external-nats: false
-
 operations:
   # A list of configmap names that should be applied to the BOSH manifest.
   custom: []
@@ -42,3 +39,11 @@ k8s-host-url: ""
 k8s-service-token: ""
 k8s-service-username: ""
 k8s-node-ca: ""
+
+nats:
+  # Set this to true to disable the NATS release, and use the official helm chart
+  # In that case the nats password is required.
+  external: false
+  auth:
+    user:     nats # Erini hard codes this to "nats"
+    password: pikachu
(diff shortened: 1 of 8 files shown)


deploy/helm/scf/values.yaml
```diff
@@ -39,3 +39,11 @@ k8s-host-url: ""
 k8s-service-token: ""
 k8s-service-username: ""
 k8s-node-ca: ""
+
+nats:
```
andreas-kupries: Shouldn't this section go under `features` ?

mook-as: Unfortunately, no; because of how helm subcharts work, I can _only_ put it under `nats` (so that the subchart picks up the settings from here).

Basically, the _only_ way to propagate the configs down is to have a top-level key with the subchart name…

Supposedly helm v3 is going to support lua scripting so that can be done, but that isn't implemented yet (even in `3.0.0-beta.3`).  Also, for some insane reason they plan to run subchart scripts before parent chart scripts…

andreas-kupries: Ah! Ok, then I would request a (small) comment about this.

andreas-kupries:
Ok, will use the same operator.

andreas-kupries:
Definitely have trouble standing up the system with external nats.
Nats does not seem to start, and then all the roles requiring it for communications fall over.

viovanov:
@mook-as cool that this works, but I think we need to find a solution for secret sharing/generation among charts
It could be that we create a secret like `scf.var-nats-password` and give up generation.. I'd rather we keep looking though.

What would you require of the subchart, to make integration simpler? (if you had to write a story to make it compatible with us)

mook-as:
Assuming I'm rewriting the NATS chart from scratch, I'd need to:

- Have an (optional) helm value that provides the name of a secret to use to look up the auth password (actually due to the granularity, one secret _each_ for user name, password, cluster password, …)
- In the SCF `values.yaml`, provide those with the expected secret names (and tell the administrator to _not_ change them)
- Have the secret generator generate those secrets (and hope helm will let the secret generator run before the subchart is deployed).

I think I'll need to explore hacking on the NATS subchart locally for a bit to know what actually would work.  In general helm doesn't seem very prepared for non-trivial subcharts.

PR #2876, opened 2019-10-04
Add install instructions for the helm release

manno:


INSTALL.md
```diff
@@ -0,0 +1,111 @@
...
+
+```
+helm install --namespace cfo --name cf-operator --set "operator.watchNamespace=scf" https://s3.amazonaws.com/cf-operators/helm-charts/cf-operator-v0.4.1%2B92.g77e53fda.tgz
+```
+
+This allows us to restart the operator, because it's not affected by webhooks. We can also delete the SCF deployment namespace to start from scratch, without redeploying the operator.
+
+## Install SCFv3
+
+Enable Eirini explicitly when installing. The `system_domain` DNS record needs to point to the IP of the external load balancer.
```
manno: There seems to be another helm chart value we can use the set the defaults Ingress IP? 
`--set "controller.service.externalIPs={$(minikube ip)}"`

Issue #2867, opened 2019-10-03
cf push fails - spec.template.spec.hostAliases.ip: Invalid value: \"\": must be valid IP address


drnic:
Deploying against scf from s3 bucket:

```
$ git clone https://github.com/starkandwayne/bootstrap-gke
$ cd bootstrap-gke
$ export PATH=$PWD/bin:$PWD/vendor/helm-tiller-manager/bin:$PATH

$ bootstrap-gke up --cf
... installs cf-operator from src + scf from s3 bucket

When I deploy an app I get an internal error about hostAliases.ip not having a value:

```
$ cf push mygoapp
Pushing app mygoapp to org system / space dev as admin...
Getting app info...
Updating app with these attributes...
  name:                mygoapp
  path:                /Users/drnic/Projects/cloudfoundry/apps/sample-go-app
  disk quota:          1G
  health check type:   port
  instances:           1
  memory:              1G
  stack:               cflinuxfs3
  routes:
    mygoapp.scf.suse.dev

Updating app mygoapp...
Mapping routes...
Comparing local files to remote cache...
Packaging files to upload...
Uploading files...
 988.53 KiB / 988.53 KiB [========================================================================================================================] 100.00% 3s

Waiting for API to complete processing files...

Stopping app...

Staging app and tracing logs...
Unexpected Response
Response code: 500
CC code:       0
CC error code:
Request ID:    9c25ea52-c689-4ccd-55a6-521b1bc54cc3::be431664-345d-416a-a913-35cefbcbb1e9
Description:   {
  "description": "Runner error: job already exists: Job.batch \"ad283b73-9063-43a7-a372-8b56ce6f0742\" is invalid: spec.template.spec.hostAliases.ip: Invalid value: \"\": must be valid IP address",
  "error_code": "CF-RunnerError",
  "code": 170016
}
```

**Comment by drnic** [comment]
I've repeated this error with latest from master:

```
bazel build //deploy/helm/scf:chart
```

phil9909:
This issue is probably related to eirini and was already fix there (https://github.com/cloudfoundry-incubator/eirini/commit/c23632d3b2a56e899db14cd5f6f0bed97bd00213), but v3 currently uses an outdated eirini version (0.0.14).
I opened a PR to bump the eirini version (https://github.com/SUSE/scf/pull/2883), but it required some more changes and might take a while to get merged.

@drnic maybe you could checkout the PR and see if it really fixes your issue.

PR #2883, opened 2019-10-07
Make cf push work from outside

phil9909:
This PR makes `cf push` work from outside the cluster. All traffic coming from outside is routed over the ingress which requires `features.ingress.enabled=true`.

## Description

* This PR primarily bumps the version of eirini from 0.0.14 to 0.0.17.
* In turn some additional changes were required as eirini now is https only.
* The blobstore was only accessible for a small set of IP-Addresses.
* Some patches were fixed for ease of debugging. See #2844. These changes are not required. We could move them out of this PR.
* The certificate for `registry.(( system domain))` needs to be signed by the `internalCA` not `kubernetes`. Otherwise the router will not be able to forward requests.

## Motivation and Context
#2845 
#2867

## How Has This Been Tested?
`cf push`

## Types of changes

## Checklist:

Maybe requiring ingress implies "My change requires a change to the documentation"?


deploy/helm/scf/assets/operations/instance_groups/bits.yaml
```diff
@@ -31,7 +31,7 @@
     namespace: {{ .Release.Namespace }}
 - type: replace
   path: /variables/name=bits_service_ssl/options/signer_type?
-  value: cluster
+  value: local
```
viovanov: Can you please explain this?
From our tests, using a Kube CSR was the only way to get the nodes to trust our registry.

phil9909: * In order to make push work from the outside the bit-service must be reachable, which is achieved via the `gorouter`. But the `gorouter` does not trust the `cluster` CA.
* To keep it possible for the `kubelet` to pull the image, we changed it to pull the image using the external url (i.e. it also goes thru the `gorouter`).

viovanov: I don't think your first statement is correct. The bits service should _only_ be used as a registry for Eirini docker images.
Pulling the image through the gorouter means the Kube nodes need to trust its CA - how does that happen?

viovanov:
@phil9909 going through the PR again - I don't understand what the title has to do with your changes.
Also, please explain why you're making registry traffic pass through the router. We're specifically trying to avoid this. We want the registry to be exposed using a cert signed by Kube so it's trusted by nodes.

phil9909:
> @phil9909 going through the PR again - I don't understand what the title has to do with your changes.
> Also, please explain why you're making registry traffic pass through the router. We're specifically trying to avoid this. We want the registry to be exposed using a cert signed by Kube so it's trusted by nodes.

I see your point (now).
How is the kubelet able to access the bits-service / image-registry inside the cluster without an external IP / DNS? We did not understand this, but if you figured out how to do this we are happy to rework this PR.

phil9909:
I went further down the rabbit hole and finally figured out the root cause of the confusion here.
Clusters provided by the Google Cloud Platform (i.e. GKE clusters) are different from the clusters we use ([Gardener](https://github.com/gardener/gardener/) provided clusters) in the following way: the docker-daemon which runs on the nodes of GKE clusters trusts the cluster CA, which is not the case on Gardener Clusters.
This is why we weren't able to get things running and thought this change would be required. I have opened an issue at the Gardener project https://github.com/gardener/gardener/issues/1515 and they will look into this soon. I have no idea if the current implementation also breaks on other clusters providers.

> Pulling the image through the gorouter means the Kube nodes need to trust its CA - how does that happen?

As the gorouter in our setup has a valid (let's encrypt) certificate this is a no-brainer. I understand the for some setups (mostly dev and CI setups) this won't be true.

f0rmiga:
@phil9909 I pushed a commit to fix the Eirini addresses that I removed when switching to use the BOSH DNS aliases and a commit to add `-o errexit -o nounset` to the configure-eirini errand.

f0rmiga:
@phil9909 Could you reopen this PR on github.com/SUSE/kubecf, please?

kramerul:
@f0rmiga  I opened two new PRs on [SUSE/kubecf](https://github.com/SUSE/kubecf) derived from this one
* [Bump eirini version in scf to 0.0.21](https://github.com/SUSE/kubecf/pull/59)
* [Support additional cidrs for k8s clusters](https://github.com/SUSE/kubecf/pull/58)

PR #2902, opened 2019-10-08
Fix loss of half of PR 2877.

andreas-kupries:
## Description

These are the changes I thought to have been merged with PR https://github.com/SUSE/scf/pull/2877
Except only half (the first commit) looks to be merged, with commit 639c4f83563d675ba169b9cda4414dce68e6da90

This new PR has all the things now, and is looking for proper merge.

## Motivation and Context

Some bad merging seems to have happened, losing part of the changes.
See also https://jira.suse.com/browse/CAP-787

## How Has This Been Tested?

See #2877 

## Types of changes

See #2877 

## Checklist:

See #2877

### DIFF
--- a/container-host-files/etc/scf/config/scripts/check_swapaccounting.sh
+++ b/container-host-files/etc/scf/config/scripts/check_swapaccounting.sh
@@ -11,9 +11,12 @@ set -x
 # First a set of commands to show us what the following checks are seeing.
 
 grep Swap /proc/meminfo ;# Do we have swap, how much ?
-if [ -e /etc/default/grub ] ; then
+if [ -r /etc/default/grub ] ; then
     grep CMDLINE /etc/default/grub ;# what does the boot loader say about swap/cgroup ?
 fi
+
+cat /proc/cmdline ;# and what does the running kernel believes about how it is configured?
+
 if [ ! -e /sys/fs/cgroup/memory/memory.memsw.limit_in_bytes ] ; then
     echo No swap accounting
 fi


container-host-files/etc/scf/config/scripts/check_swapaccounting.sh
```diff
@@ -1,19 +1,33 @@
...
 # enabled on the kernel.  It is an error to have swap without swapaccounting (as
 # that would lead to mis-tracking of memory usage).  If there is no swap,
 # however, not having swap accounting is safe; in that case we turn it off for
 # garden as well.
 
+# First a set of commands to show us what the following checks are seeing.
+
+grep Swap /proc/meminfo ;# Do we have swap, how much ?
+if [ -e /etc/default/grub ] ; then
+    grep CMDLINE /etc/default/grub ;# what does the boot loader say about swap/cgroup ?
```
mook-as: We should be looking at `/proc/cmdline` instead no? (Since that's the _currently running_ arguments)

Also: nit: use `-r` instead of `-e`, since it might exist but is not readable :D

andreas-kupries: I see. Checking /proc/cmdline means I would not have to look for grub or other config file, because I query the kernel directly, and not the boot stuff ... That said, I could query both, in case it manages to not pick up what we think the config is.

Will change.

andreas-kupries: Addressed with commit a280e36f66e6c86a4299200409588e618fbe7a6b

PR #2906, opened 2019-10-09
Add Eirini staging images in the imagelist file

jimmykarily:
This PR includes all the changes needed in order to have all Eirini images (staging + eirinifs) based on SLE.

Includes scripts used in CI plus the latest image builds.

These changes will also let them be copied over to the final location upon releasing. Will also help users copy the needed images in their private registries for air-gapped installation.

## Checklist:

### DIFF
--- a/bin/dev/create_helm_charts_pr.sh
+++ b/bin/dev/create_helm_charts_pr.sh
@@ -59,6 +59,8 @@ sed -i 's@^\(\s\+\)hostname:\s\+".*"$@\1hostname: "registry.suse.com"@' stable/u
 sed -i 's@^\(\s\+\)organization:\s\+".*"$@\1organization: "cap"@' stable/cf/values.yaml
 sed -i 's@^\(\s\+\)organization:\s\+".*"$@\1organization: "cap"@' stable/uaa/values.yaml
 
+sed -i 's@\(\s\+EIRINI_.*_IMAGE:\s\+"registry.suse.com\/\)cap-staging@\1cap@' stable/cf/values.yaml
+
 $HUB config user.email "cf-ci-bot@suse.de"
 $HUB config user.name "${GITHUB_USER}"
 $HUB checkout -b $bundle
--- a/bin/dev/create_helm_charts_pr.sh
+++ b/bin/dev/create_helm_charts_pr.sh
@@ -59,7 +59,7 @@ sed -i 's@^\(\s\+\)hostname:\s\+".*"$@\1hostname: "registry.suse.com"@' stable/u
 sed -i 's@^\(\s\+\)organization:\s\+".*"$@\1organization: "cap"@' stable/cf/values.yaml
 sed -i 's@^\(\s\+\)organization:\s\+".*"$@\1organization: "cap"@' stable/uaa/values.yaml
 
-sed -i 's@\(\s\+EIRINI_.*_IMAGE:\s\+"registry.suse.com\/\)cap-staging@\1cap@' stable/cf/values.yaml
+sed -i 's@registry.suse.com/cap-staging/@@' stable/cf/values.yaml
 
 $HUB config user.email "cf-ci-bot@suse.de"
 $HUB config user.name "${GITHUB_USER}"
--- a/bin/image-list.rb
+++ b/bin/image-list.rb
@@ -17,11 +17,10 @@ Dir.glob(File.join(helm, "templates", "*.yaml")).each do |file|
   end
 end
 
-File.open(File.join(helm, "values.yaml")).each do |line|
-  ["DOWNLOADER","EXECUTOR","UPLOADER"].each do |image|
-    /^\s+EIRINI_#{image}_IMAGE: ["'].+\/([^\/]*)["']$/.match(line) do |match|
-      images[match[1]] = true
-    end
+# Eirini staging images
+values['env'].each_value do |value|
+  /^registry\.suse\.com\/cap-staging\/(.*)$/.match(value) do |match|
+    images[match[1]] = true
   end
 end
 
--- a/bin/image-list.rb
+++ b/bin/image-list.rb
@@ -17,6 +17,14 @@ Dir.glob(File.join(helm, "templates", "*.yaml")).each do |file|
   end
 end
 
+File.open(File.join(helm, "values.yaml")).each do |line|
+  ["DOWNLOADER","EXECUTOR","UPLOADER"].each do |image|
+    /^\s+EIRINI_#{image}_IMAGE: ["'].+\/([^\/]*)["']$/.match(line) do |match|
+      images[match[1]] = true
+    end
+  end
+end
+
 File.open(File.join(helm, "imagelist.txt"), "w") do |file|
   file.puts(images.keys.sort)
 end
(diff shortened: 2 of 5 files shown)


jimmykarily:
Works together with: https://github.com/SUSE/cloudfoundry/pull/137

bin/dev/create_helm_charts_pr.sh
```diff
@@ -59,6 +59,8 @@ sed -i 's@^\(\s\+\)hostname:\s\+".*"$@\1hostname: "registry.suse.com"@' stable/u
 sed -i 's@^\(\s\+\)organization:\s\+".*"$@\1organization: "cap"@' stable/cf/values.yaml
 sed -i 's@^\(\s\+\)organization:\s\+".*"$@\1organization: "cap"@' stable/uaa/values.yaml
 
+sed -i 's@\(\s\+EIRINI_.*_IMAGE:\s\+"registry.suse.com\/\)cap-staging@\1cap@' stable/cf/values.yaml
```
jandubois: This is not correct. The registry name should be stripped completely, e.g. `registry.suse.com/cap-staging/recipe-downloader:0.4.0` should become `recipe-downloader:0.4.0`.

I would also drop the match on `EIRINI...` so this works generally. We should not have any other occurrences of `registry.suse.com/cap-staging/` in our `values.yaml` file.

```suggestion
sed -i 's@registry.suse.com/cap-staging/@@' stable/cf/values.yaml
```

`recipe-downloader:0.4.0` will work correctly because `kube.registry` is set to the correct values both for the staging and for the release chart.

bin/image-list.rb
```diff
@@ -17,6 +17,14 @@
...
 end
 
+File.open(File.join(helm, "values.yaml")).each do |line|
+  ["DOWNLOADER","EXECUTOR","UPLOADER"].each do |image|
+    /^\s+EIRINI_#{image}_IMAGE: ["'].+\/([^\/]*)["']$/.match(line) do |match|
+      images[match[1]] = true
+    end
+  end
+end
+
```
jandubois: This whole block should not be necessary because the case is already handled by the "config variables" section above because the image name now no longer includes a slash.

bin/dev/create_helm_charts_pr.sh
jimmykarily: +1 to remove the registry part, I didn't realize [this](https://github.com/cloudfoundry-incubator/fissile/blob/develop/kube/pod.go#L516) also applied to env variables. I just looked in the helm chart and it should work: 

```        - name: "EIRINI_DOWNLOADER_IMAGE"
          value: {{if ne (typeOf .Values.env.EIRINI_DOWNLOADER_IMAGE) "<nil>"}}{{if contains "/" .Values.env.EIRINI_DOWNLOADER_IMAGE}}{{.Values.env.EIRINI_DOWNLOADER_IMAGE | quote}}{{else}}{{print .Values.kube.registry.hostname "/" .Values.kube.organization "/" .Values.env.EIRINI_DOWNLOADER_IMAGE | quote}}{{end}}{{else}}""{{end}}
```

Regarding the EIRINI part, I wanted to be explicit because this script assumes that some other script has already copied the images to their final location (in this case the release tool). Since the release tool [explicitly copies only 3 images](https://github.com/SUSE/cloudfoundry/pull/137/files#diff-94defef02205ea151bd44a340383d2aeR161) I wanted to limit the assumption here only to those 3. What I'm thinking is that this script might replace a url for an image which the release tool never actually copied. In any case, there are too many moving parts so none of the solutions is bullet proof. I will go with your suggestion.

One final note, the staging chart doesn't have the "correct" values for the EIRINI images in the staging chart (as described also here: https://github.com/SUSE/cloudfoundry/pull/137/files#r333841514) but does in the released one (after the copying has happened). Since this script prepares the release chart, we can remove the registry part.

bin/image-list.rb
jimmykarily: I could be wrong but I think this script is run on very first un-altered staging values.yaml. That `values.yaml` file should have whatever we define [in the role-manifest](https://github.com/SUSE/scf/pull/2906/files#diff-a65976ef35c13467fe140d46c6ca0d3aR3786) which is the cap-staging url with the slashes in it.

jandubois: Ah, yes, in that case just loop over the `values` hash that has already been setup. Something like this (untested braindump):

```
values['env'].each_value do |value|
    /^registry\.suse\.com\/cap-staging\/(.*)$/.match(value) { |match| images[match[1]] = true }
end
```

I don't know if we can guarantee that this will be run with Ruby 2.5+, then we could make it more readable like

```
values['env'].each_value do |value|
    image = value.delete_prefix("registry.suse.com/cap-staging/")
    images[image] = true unless image.nil?
end
```

jimmykarily: I'll go with the first suggestion to be safe. Regular expressions are very readable (said no one ever).

container-host-files/etc/scf/config/role-manifest.yml
```diff
@@ -148,9 +148,13 @@ releases:
   url: "https://bosh.io/d/github.com/cloudfoundry-incubator/bits-service-release?v=2.28.0"
   sha1: "bd173fc9952277ea962a330c04555d9bd1e8a10e"
 - name: "eirini"
-  url: "https://suse-final-releases.s3.amazonaws.com/eirini-release-0.0.17.tgz"
+  url: "https://s3.eu-central-1.amazonaws.com/tmp-aljfkdjadk-eirinifs-bucket/eirini-bosh-release-dev-b21b0c6.tgz"
```
jandubois: This is just for testing and will be replaced with a final-release before this PR is merged, right?

container-host-files/etc/scf/config/role-manifest.yml
```diff
@@ -2149,9 +2153,6 @@ instance_groups:
           memory: 256
           virtual-cpus: 1
           service-account: eirini
-  configuration:
-    templates:
-      properties.eirini.run_cert_copier: ""
```
jandubois: Why are you removing this? This setting prevents eirini from deploying the cert_copier. If you remove this setting, then you have to update eirini-bosh-release to unconditionally stop deploying the cert_copier (for SCF).

jandubois: Oh, I see you have already done this in https://github.com/cloudfoundry-community/eirini-bosh-release/commit/f2856a00bdc234e3437bff61888b5d5df64e4dc4

container-host-files/etc/scf/config/role-manifest.yml
jimmykarily: yes, letting Jenkins build a chart to try out on a local cluster. Will reolace with a final release if it works

jimmykarily: It didn't work with the cri-o kind version, worked on the containerd kind version and also worked on CaaSP 4 (crio). The opi registry certificate was not working for some reason on crio-kind but this might be a problem with kind since it works on CaaSP. We will investigate more but this PR should be ok. It changes nothing relevant to the opi registry anyway.

PR #2913, opened 2019-10-09
Create ~/.kube/config with https endpoint to use RBAC

jandubois:
The default http endpoint bypasses RBAC validation, so all requests are accepted. Which also means that `kubectl --as` doesn't work.

## How Has This Been Tested?

Verify that you can run commands as a restricted service account, and make sure commands that are not supposed to be allowed actually fail, e.g.

```
vagrant@vagrant-kube:~> kubectl --as system:serviceaccount:default:default cluster-info

To further debug and diagnose cluster problems, use 'kubectl cluster-info dump'.
Error from server (Forbidden): services is forbidden: User "system:serviceaccount:default:default" cannot list resource "services" in API group "" in the namespace "kube-system"
```

### DIFF
--- a/vagrant/create_kubeconfig.sh
+++ b/vagrant/create_kubeconfig.sh
@@ -6,9 +6,9 @@
 
 set -o errexit -o nounset -o xtrace
 
-SECRET=$(kubectl get sa default -n kube-system -o jsonpath="{.secrets[0].name}")
-CA_CRT=$(kubectl get secrets ${SECRET} -n kube-system -o jsonpath="{.data['ca\.crt']}")
-TOKEN=$(kubectl get secrets ${SECRET} -n kube-system -o jsonpath="{.data['token']}" | base64 -d)
+SECRET=$(kubectl get serviceaccount default --namespace kube-system --output jsonpath="{.secrets[0].name}")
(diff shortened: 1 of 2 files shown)


vagrant/create_kubeconfig.sh
```diff
@@ -0,0 +1,34 @@
...
+
+# Create ~/.kube/config file that uses the https endpoint to make
+# sure all requests go through rbac validation (requests via the
+# http endpoint bypass all validation).
+
+set -o errexit -o nounset -o xtrace
+
+SECRET=$(kubectl get sa default -n kube-system -o jsonpath="{.secrets[0].name}")
+CA_CRT=$(kubectl get secrets ${SECRET} -n kube-system -o jsonpath="{.data['ca\.crt']}")
+TOKEN=$(kubectl get secrets ${SECRET} -n kube-system -o jsonpath="{.data['token']}" | base64 -d)
```
mook-as: Nit: use full names for scripts
```suggestion
SECRET=$(kubectl get serviceaccount default --namespace kube-system --output jsonpath="{.secrets[0].name}")
CA_CRT=$(kubectl get secrets ${SECRET} --namespace kube-system --output jsonpath="{.data['ca\.crt']}")
TOKEN=$(kubectl get secrets ${SECRET} --namespace kube-system --output jsonpath="{.data['token']}" | base64 -d)
```

PR #2916, opened 2019-10-10
Switched from k8s headless svc DNS resolution to BOSH aliases

### LINKED ISSUE #2780: Spike: BOSH-compatible hostname aliases
A lot of the ops files in SCF are concerned with changing hostnames (in properties as well as certificates).
Everything would be simpler if we could re-use the hostnames used by the BOSH releases.

f0rmiga:
Resolves #2780.


f0rmiga:
Depends on https://github.com/cloudfoundry-incubator/cf-operator/pull/598.

f0rmiga:
I'll get https://github.com/cloudfoundry-incubator/cf-operator/pull/642 in before merging this PR.

PR #2992, opened 2019-10-17
Jenkinsfile: Get the correct secret when testing rotation

mook-as:
## Description
We changed configgin to put exported BOSH properties in secrets; the Jenkins test for rotated secrets is now failing to find the correct secret.  We need to update it to locate the correct secret.

## Motivation and Context
Jenkins builds are failing with:
`Secret \${secret_name} not correctly rotated`

## How Has This Been Tested?
This will trigger CI

## Types of changes

## Checklist:

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -778,12 +778,16 @@ pipeline {
                     . make/include/secrets
 
                     # Get the last updated secret
-                    secret_resource="\$(kubectl get secrets --namespace="${cfNamespace}" --output=jsonpath='{.items[-1:].metadata.name}' --sort-by=.metadata.resourceVersion)"
+                    secret_resource="\$(kubectl get secrets --namespace="${cfNamespace}" --output=name --sort-by=.metadata.resourceVersion | grep secrets- | tail -n1)"
 
                     # Get a random secret that should be rotated (TODO: choose this better)
                     secret_name=internal-ca-cert
                     # And its value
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -778,12 +778,16 @@ pipeline {
...
 
                     # Get the last updated secret
-                    secret_resource="\$(kubectl get secrets --namespace="${cfNamespace}" --output=jsonpath='{.items[-1:].metadata.name}' --sort-by=.metadata.resourceVersion)"
+                    secret_resource="\$(kubectl get secrets --namespace="${cfNamespace}" --output=name --sort-by=.metadata.resourceVersion | grep secrets- | tail -n1)"
 
                     # Get a random secret that should be rotated (TODO: choose this better)
                     secret_name=internal-ca-cert
                     # And its value
-                    old_secret_value="\$(kubectl get secret --namespace="${cfNamespace}" "\${secret_resource}" -o jsonpath="{.data.\${secret_name}}" | base64 -d)"
+                    old_secret_value="\$(kubectl get --namespace="${cfNamespace}" "\${secret_resource}" -o jsonpath="{.data.\${secret_name}}" | base64 -d)"
```
andreas-kupries: I believe that we should note here and below, that the changed method of getting the `secret_resource` now has the type of resource, i.e. `secret` in the string, which is why it was removed here. At least that is my deduction from staring at the changes a bit.

Beyond that LGTM.

mook-as: Addressed in 0abf552e3e4cacd9be8adb41714159312fd13337

PR #3010, opened 2019-10-22
Bump stemcell version to include configgin 0.19.1

jandubois:
That configgin version will  no longer update the statefulset pod templated for bosh links imported from other jobs included in the same instance group.

This should avoid having the mysql role triggering a restart upon deployment.

[scf#CAP-150]

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -328,7 +328,7 @@ pipeline {
         )
         string(
             name: 'FISSILE_STEMCELL_VERSION',
-            defaultValue: '12SP4-4.ga824718-0.230',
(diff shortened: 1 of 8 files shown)


jandubois:
> I do still see pod restarts;

This change would only avoid restarts due to bosh link imports from jobs co-located within the same instance group. In particular, this should avoid restarts of the mysql pods (but not the mysql-proxy pods).

However, @satadruroy observed even longer startup times than before, so this may not be suitable for merging into 1.5.1 without additional verification.

I won't be able to look at this until next week.

jandubois:
Functionality included in #3089

PR #2990, opened 2019-10-17
Switch to obs fissile stemcell

thardeck:
Test pr for Jenkins run

PR #3067, opened 2019-11-07
[CI Trigger] Try the OBS built stemcell

jimmykarily:


jimmykarily:
QA done, closing this. Btw, this looks like a better way to do it: https://github.com/SUSE/scf/pull/2990/files

PR #3089, opened 2019-11-16
Avoid pod restarts

jandubois:
Break all cyclic dependencies on bosh links because they are incompatible with the new configgin version that sequences pod startup so that all imported properties (via bosh links) are generated before the pod starts.

Also includes various changes to speed up deployment in general.

## How Should This Be Tested?

Initial testing involves just watching the deployment and verifying that pods don't restart after entering `Running` state. The whole cluster should become ready much faster than before this PR, but otherwise operate unchanged.

Extended testing should include HA deployments, scaling up instance counts, secrets rotation, and updates, both from the previous release, as well as to a hypothetical next release, that also uses the new configgin mechanism.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -328,7 +328,7 @@ pipeline {
         )
         string(
             name: 'FISSILE_STEMCELL_VERSION',
-            defaultValue: '12SP4-4.ga824718-0.230',
+            defaultValue: '12SP4-11.g2837aef-0.230',
             description: 'Fissile stemcell version used as docker image tag',
         )
         booleanParam(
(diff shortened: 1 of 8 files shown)


jandubois:
This PR is not ready; it is missing and update to fissile from https://github.com/cloudfoundry-incubator/fissile/pull/525 (PR should be ready for merging, but CI is currently broken).

It is also missing the update to configgin that implements the new logic. For testing https://github.com/cloudfoundry-incubator/configgin/commit/ed8c96f2ca276a73dd1b005c1964d99f2c7d9aae can be used to update a local stemcell, but that branch only works for initial deployment of SA setups and is only at the beginning of a bigger refactoring to finish various update scenarios.

PR #3189, opened 2019-12-11
Set topologyKey for podAntiAffinity rule to kubernetes.io/hostname

jandubois:
Also make it configurable via kube.podAntiAffinityTopologyKey.

The original value of beta.kubernetes.io/os doesn't make any sense: it means to try to schedule each pod on a different OS. But our pods only work on Linux, and we don't expect to see any non-Linux nodes in the cluster anyways.

[jsc#CAP-1067]

### DIFF
--- a/container-host-files/etc/scf/config/role-manifest.yml
+++ b/container-host-files/etc/scf/config/role-manifest.yml
@@ -746,7 +746,7 @@ instance_groups:
                     - key: "app.kubernetes.io/component"
                       operator: In
                       values:
-                      - mysql
+                      - mysql-proxy
                   topologyKey: '{{if eq (typeOf .Values.kube.podAntiAffinityTopologyKey) "<nil>"}}kubernetes.io/hostname{{else}}{{.Values.kube.podAntiAffinityTopologyKey}}{{end}}'
   tags:
   - active-passive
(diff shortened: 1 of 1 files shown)


jandubois:
> it looks like we may want to set it to _empty_ instead (to mean some combination of hostname, zone, and region)?

You would think that (and so did I), but it doesn't actually work:

```
Required value: can not be empty, spec.affinity.podAntiAffinity.preferredDuringSchedulingIgnoredDuringExecution[0].podAffinityTerm.topologyKey: Invalid value: "": name part must be non-empty
```

It was part of the original implementation in kube 1.7, but didn't work even then. kube 1.8 added validation that explicitly forbids this: https://github.com/kubernetes/kubernetes/pull/49976

Pod scheduling in kube is still mostly an unsolved problem. Look at https://kubernetes.io/docs/concepts/scheduling/kube-scheduler/ for a list of all the things you can tweak, and maybe read https://kubernetes.io/docs/concepts/workloads/pods/pod-topology-spread-constraints/ for yet another feature to eventually reach resilient and balanced pod placement. Enter at your own risk...

Read https://github.com/kubernetes/kubernetes/issues/41598 if you ever wonder why the docs say:

>If the name of the StatefulSet contains dashes (“-”), volume zone spreading may not provide a uniform distribution of storage across zones.

Just some background on why I decided to set topology to hostname and called it a day... 😃

PR #3194, opened 2019-12-12
Allow empty UAA_CA_CERT to use system cert store

### LINKED ISSUE #3187: CF API returns HTTP 503/502 errors when UAA has an upstream load balancer with SSL
**Describe the bug**

When UAA is using an SSL certificate signed by anything other than the SCF Internal CA, some CF API calls return errors (though they still *do* things, despite the errors). I've observed this with `cf create-org` and `cf create-space`, but it likely applies to others.

**To Reproduce**

Deploy to EKS, attach an ACM cert to the UAA ELB, attempt commands, watch them fail. Remove the cert, watch them succeed.

**Expected behavior**

If the certificate is signed by a CA in the container's default CA list, it should just work.

**Logs**

With SSL on the load balancer...

```
$ cf create-org test-org-2
Creating org test-org-2 as admin...
OK

Assigning role OrgManager to user admin in org test-org-2...
Unexpected Response
Response code: 503
CC code:       0
CC error code:
Request ID:    3ba41173-a8f7-4baa-416c-3c56cc2fb2e8::211288ee-e9c5-4575-ac88-fd4ad56667e8
Description:   {
  "description": "The UAA service is currently unavailable",
  "error_code": "CF-UaaUnavailable",
  "code": 20004
}

FAILED


$ cf create-org test-org-2

Creating org test-org-2 as admin...
OK

Org test-org-2 already exists.
```

Without SSL on the load balancer...

```
$ cf create-org test-org-2
Creating org test-org-2 as admin...
OK

Assigning role OrgManager to user admin in org test-org-2...
OK

TIP: Use 'cf target -o "test-org-2"' to target new org
```

**Environment**

SCF 1.5.1
Kubernetes 1.14.8
Amazon EKS (v5)

**Additional context**

This happens because `cloud_controller_ng` communicates with `uaa` externally. Changing `uaa.internal_url` yields 502 errors, rather than 503s. This seems to be because the value of `uaa.internal_url` includes a zone prefix, and `cloud_controller_ng` is relying on this to indicate the appropriate zone.

Manually reconfiguring `uaa.ca_file` to `/var/lib/ca-certificates/ca-bundle.pem` resolves this issue, since the SCF Internal CA is part of the default list of trusted CAs for the container.

However, this value is not configurable at deploy-time.

jandubois:
Despite the branch name this is CAP-1144.

Fixes #3187

One functional change: now unconditionally requires TLS for the connection between credhub and the database. Commit d1059152 claims this is not supported by our config, but it seems to work just fine with the credhub brain test. This change is required to make credhub work with UAA_CA_CERT being empty and the embedded UAA using the internal cert.


jandubois:
Please wait for confirmation from @colstrom before merging!

andreas-kupries:
Tested under vagrant using split UAA and embedded UAA cases. All work, and embedded UAA exercises the new code paths in the patched ERBs.

We are not worse off than before. Merging was go.

PR #3223, opened 2019-12-16
Use CONFIGGIN_SA_TOKEN to fetch NATS statefulset

jandubois:
The default service account lacks permission to do so on GKE.

[jsc#CAP-1149]


jandubois:
> Seems reasonable; as Jenkins doesn't test this, I'll assume you did?

I tested that SA deployment still works, so there is no syntax error in the change.

Testing that the change actually makes a difference on GKE will be done by QA CI systems, which require a nightly build, so I'm going ahead and merging the PR to get it into tonight's build.

PR #3227, opened 2019-12-19
(DEBUGGING PR) Fix ha pod restarts original stemcells

jandubois:
PR exists solely to verify some scenarios; it should never be merged!


jandubois:
Turns out to be not needed.

PR #3256, opened 2020-01-10
Remove vestigial cf-mysql-release

jandubois:


jandubois:
We have `cf_mysql_enabled: false` in our opinions, so monit it not actually running any code at runtime. So removing the code so close to the RC seems too risky.

PR #3274, opened 2020-01-15
Keep exposing the file_server HTTP port for upgrades

jandubois:
Already deployed applications have the file_server URL stored in the LRP in the Diego database. They will break if these URLs stop working. With HTTPS enabled the HTTP should return a redirect to the new URL, so applications don't have to be restarted after an upgrade.

Ref: https://www.pivotaltracker.com/n/projects/1003146/stories/165872883


mook-as (approved):
LGTM; confirmed that we're still listening on that port (so no further changes are needed).

PR #3288, opened 2020-01-27
Use experimental loggregator dev release to test DNS fix

jandubois:
**This is just a test and should never be merged**

Ref https://github.com/cloudfoundry/loggregator-release/commit/6d66cf7a793e4fcec18ff036caacc2ebf7bade6e


mook-as (changes requested):
Making sure we don't accidentally merge this. (Not that it's likely, given it's a draft PR.)

PR #3291, opened 2020-01-27
Omnibus PR to get from RC2 to RC3

jandubois:
### Bump external UAA to cf-deployment-12.17 level

This update was overlooked when the SCF changes were merged, and the embedded and external UAA were out of sync.

The external UAA was also mistakenly using the database seeder credentials when using an external database; it now always uses `uaaadmin`.

### Bump back loggregator to v105.6.3

This is a hotfix releases for the v105 version and has been tagged in Dec 2019. There is no bosh.io release for it because v106 was already out, so we had to build a local dev release and put it into our own S3 bucket. The purpose of this change is to avoid the DNS spamming behaviour of v106 (see https://github.com/cloudfoundry/loggregator-release/issues/401).

### Disentangle credhub TLS from supporting an empty `UAA_CA_CERT`

The changes in RC2 required credhub to use TLS to get the `INTERNAL_CA_CERT` into the Java trust store. Now a release script, that runs before configgin, will install it into the system cert store, so it will always be copied by the credhub pre-start script.

### Add `DB_EXTERNAL_USER_HOST_SUFFIX` variable

On Azure MariaDB the username must always include the hostname, like `'user@host'@'host.azure.com`. This variable is appended to all hostnames in the rolemanifest.

Note that the cf-usb username is hardcoded in the release, so it will be incompatible with external database support on Azure.

### DIFF
--- a/container-host-files/etc/scf/config/opinions.yml
+++ b/container-host-files/etc/scf/config/opinions.yml
@@ -338,7 +338,6 @@ properties:
       sql:
         db_port: 3306
         db_schema: diego
-        db_username: diego
     executor:
       garden:
         address: 127.0.0.1:7777
@@ -349,7 +348,6 @@ properties:
       sql:
         db_port: 3306
         db_schema: diego_locket
-        db_username: diego_locket
     rep:
       evacuation_timeout_in_seconds: 60
       preloaded_rootfses:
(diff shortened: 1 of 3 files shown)


gaktive (approved):
Jan walked Satadru & I through the changes. LGTM

PR #3309, opened 2020-02-04
Extended PR creation script with command removing RC references from charts to be released.

andreas-kupries:
## Description

Added commands to the script creating the release PR which squash `-rcN` strings from the
chart. Affected: `version`, `scfVersion`, `CLUSTER_BUILD`. The last is the information goes into the `v2/info` endpoint of a deployed CF.

## Motivation and Context

See https://jira.suse.com/browse/CAP-1120

## How Has This Been Tested?

Manual testing with github operations disabled and copying a fixed bundle into place instead of downloading anything.

## Types of changes
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)

## Checklist:
- [x] My change requires a change to the documentation.

The documentation to change is the https://confluence.suse.com/display/CAP/CAP+Release+Process+Template, section `Shipping`.

### DIFF
--- a/bin/dev/create_helm_charts_pr.sh
+++ b/bin/dev/create_helm_charts_pr.sh
@@ -42,6 +42,10 @@ unzip -d bundle/ bundle.zip
 bundle=$(basename "$CAP_BUNDLE" .zip)
 bundle_readable=$(printf '%b' "${bundle//%/\\x}")
 product=`echo $bundle_readable | cut -d+ -f1`
+
+# Strip -rcN from the product version shown in PR description.
+product="$(echo "${product}" | sed -e 's|-rc.*||')"
+
 pr_title="Release $product"
 pr_description="Publish Helm charts for release $product created from $bundle_readable.zip built in Jenkins run $SOURCE_BUILD."
 
@@ -52,6 +56,25 @@ rm -rf stable/uaa
 # Place the new ones
 cp -r ../bundle/helm/cf stable/
 cp -r ../bundle/helm/uaa stable/
+
+# Show the places in the charts having RC references
+grep -rn '.-rc' stable/{cf,uaa}
+
+# Fixing RC references.
+
+# CLUSTER_BUILD: "2.20.2-rc1"
+sed -i 's|-rc.*"$|"|' stable/cf/values.yaml
+sed -i 's|-rc.*"$|"|' stable/uaa/values.yaml
+
+# scfVersion: 2.20.2-rc1+cf12.17.0.0.g5c8dc458
+sed -i 's|-rc.*cf|+cf|' stable/cf/Chart.yaml
+sed -i 's|-rc.*cf|+cf|' stable/uaa/Chart.yaml
+
+# version: 2.20.2-rc1
+sed -i 's|-rc.*$||'     stable/cf/Chart.yaml
+sed -i 's|-rc.*$||'     stable/uaa/Chart.yaml
+
+# Fixing docker registry references.
 
 sed -i 's@^\(\s\+\)hostname:\s\+".*"$@\1hostname: "registry.suse.com"@' stable/cf/values.yaml
 sed -i 's@^\(\s\+\)hostname:\s\+".*"$@\1hostname: "registry.suse.com"@' stable/uaa/values.yaml
--- a/bin/dev/create_helm_charts_pr.sh
+++ b/bin/dev/create_helm_charts_pr.sh
@@ -67,9 +67,6 @@ sed -i 's|-rc[0-9][0-9]*||' stable/cf/values.yaml
 sed -i 's|-rc[0-9][0-9]*||' stable/uaa/values.yaml
 
 # scfVersion: 2.20.2-rc1+cf12.17.0.0.g5c8dc458
-sed -i 's|-rc[0-9][0-9]*||' stable/cf/Chart.yaml
-sed -i 's|-rc[0-9][0-9]*||' stable/uaa/Chart.yaml
-
 # version: 2.20.2-rc1
 sed -i 's|-rc[0-9][0-9]*||' stable/cf/Chart.yaml
 sed -i 's|-rc[0-9][0-9]*||' stable/uaa/Chart.yaml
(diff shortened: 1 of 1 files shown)


bin/dev/create_helm_charts_pr.sh
```diff
@@ -42,6 +42,10 @@ unzip -d bundle/ bundle.zip
 bundle=$(basename "$CAP_BUNDLE" .zip)
 bundle_readable=$(printf '%b' "${bundle//%/\\x}")
 product=`echo $bundle_readable | cut -d+ -f1`
+
+# Strip -rcN from the product version shown in PR description.
+product="$(echo "${product}" | sed -e 's|-rc.*||')"
```
mook-as:
```suggestion
product="$(echo "${product}" | sed -e 's|-rc.*$||')"
```
Not sure that's actually needed here.

bin/dev/create_helm_charts_pr.sh
```diff
@@ -53,6 +57,25 @@ rm -rf stable/uaa
...
 cp -r ../bundle/helm/uaa stable/
 
+# Show the places in the charts having RC references
+grep -rn '.-rc' stable/{cf,uaa}
+
+# Fixing RC references.
+
+# CLUSTER_BUILD: "2.20.2-rc1"
+sed -i 's|-rc.*"$|"|' stable/cf/values.yaml
+sed -i 's|-rc.*"$|"|' stable/uaa/values.yaml
```
mook-as: I didn't see this one, but that's probably dine.

bin/dev/create_helm_charts_pr.sh
```diff
@@ -53,6 +57,25 @@ rm -rf stable/uaa
...
+
+# scfVersion: 2.20.2-rc1+cf12.17.0.0.g5c8dc458
+sed -i 's|-rc.*cf|+cf|' stable/cf/Chart.yaml
+sed -i 's|-rc.*cf|+cf|' stable/uaa/Chart.yaml
```
mook-as: Might be good to comment that this is _not_ a typo, we do indeed say `cf` in the UAA chart.
Though this might still be better as:
```suggestion
sed -i 's|-rc[0-9]\+||' stable/uaa/Chart.yaml
```

mook-as (approved):

bin/dev/create_helm_charts_pr.sh
andreas-kupries: Hm. The `-rc...` is always at the end, AFAIK/CT. Maybe go with digits as well, instead of the general `.*`.

bin/dev/create_helm_charts_pr.sh
andreas-kupries: Note how that is not in the `Chart.yaml`, but the `values.yaml`. And not in the first page either. Somewhere deep under `env:`.

bin/dev/create_helm_charts_pr.sh
andreas-kupries: Will trial this without `cf`. Note, will also go to multi-digit. As much as we wish to keep the number of RCs down ...

bin/dev/create_helm_charts_pr.sh
andreas-kupries: Actually, thinking about it ... Using `-rc[0-9][0-9]*` generally should simplify replacements (always empty), and will find everything, regardless of location. Trialing that now.

andreas-kupries: Addressed in commit a38786f3eb1a4343fe7bb3a205775774729fa753.

bin/dev/create_helm_charts_pr.sh
andreas-kupries: Addressed in commit a38786f3eb1a4343fe7bb3a205775774729fa753.

bin/dev/create_helm_charts_pr.sh
andreas-kupries: Addressed in commit a38786f3eb1a4343fe7bb3a205775774729fa753.

bin/dev/create_helm_charts_pr.sh
```diff
@@ -53,6 +57,25 @@ rm -rf stable/uaa
...
+sed -i 's|-rc[0-9][0-9]*||' stable/cf/values.yaml
+sed -i 's|-rc[0-9][0-9]*||' stable/uaa/values.yaml
+
+# scfVersion: 2.20.2-rc1+cf12.17.0.0.g5c8dc458
+sed -i 's|-rc[0-9][0-9]*||' stable/cf/Chart.yaml
+sed -i 's|-rc[0-9][0-9]*||' stable/uaa/Chart.yaml
+
+# version: 2.20.2-rc1
+sed -i 's|-rc[0-9][0-9]*||' stable/cf/Chart.yaml
+sed -i 's|-rc[0-9][0-9]*||' stable/uaa/Chart.yaml
```
mook-as, line 72: There's no need to run the replace twice in the same file (even though there are two instances of it); `sed` will just do both, right?

Also, I still don't see it in _UAA_ `values.yaml`, just the CF copy.

```
$ grep -riP -- '-rc[^\s]' .
./helm/uaa/Chart.yaml:version: 2.20.3-rc1
./helm/uaa/Chart.yaml:scfVersion: 2.20.3-rc1+cf12.17.0.0.g7175b9de
./helm/cf/values.yaml:  CLUSTER_BUILD: "2.20.3-rc1"
./helm/cf/Chart.yaml:version: 2.20.3-rc1
./helm/cf/Chart.yaml:scfVersion: 2.20.3-rc1+cf12.17.0.0.g7175b9de
```

mook-as (changes requested):

bin/dev/create_helm_charts_pr.sh
andreas-kupries: Actually, right. With the modified pattern the first per file will catch them all. Oops.

andreas-kupries: Addressed in commit 3266af72113342eac79ac2a55ba89260e0758ffc

mook-as (approved):
Good enough :D

